Issue #38, opened 2014-02-24
FAIL: test_set_verify_callback_exception on FreeBSD (pyopenssl 0.14)


koobs:
OS: FreeBSD 9.2-STABLE #0 r260870
Python: 2.7.6

```
======================================================================
FAIL: test_set_verify_callback_exception (OpenSSL.test.test_ssl.ContextTests)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/mnt/home/user/repos/freebsd/ports/security/py-openssl/work/pyca-pyopenssl-0146d44/OpenSSL/test/test_ssl.py", line 1002, in test_set_verify_callback_exception
    Exception, self._handshake_test, serverContext, clientContext)
  File "/mnt/home/user/repos/freebsd/ports/security/py-openssl/work/pyca-pyopenssl-0146d44/OpenSSL/test/util.py", line 270, in failUnlessRaises
    % (exception.__name__, result))
AssertionError: Exception not raised (None returned)
----------------------------------------------------------------------
```

alex:
This code has also certainly been rewritten in the last 3 years, so I'm going to close. Please comment if this is still an issue.

PR #77, opened 2014-03-24
make test_set_verify_callback_exception pass on the *BSD

kouk:
There is a timing problem in the code that tests the verification callback that apparently occurs in BSD systems (at least FreeBSD and NetBSD, see https://github.com/pyca/pyopenssl/issues/38). This small fix seems to do the trick for me, although I can't remember exactly why (it was a month ago that I tested it). Hope @koobs and @0-wiz-0 can confirm that it works for them too.


exarkun:
Pending fixes and some kind of BSD continuous integration.  Feel free to re-open when ready to take the next step.

PR #93, opened 2014-04-11
Make X509Name instances usable as dictionary keys.

Bluehorn:
I tried to use X509Name instances as dictionary keys and was surprised that looking up the certificates by subject failed. Checking the X509_NAME_hash function I was wondering why this can go wrong only to notice on the third read that X509Name implements a method hash but not the special method **hash**.

I think that it is much more intuitive to have X509Name work like special strings. Also, if a == b and a.hash() == b.hash() I would also expect hash(a) == hash(b) to hold.

I hope you agree :-)

### DIFF
--- a/OpenSSL/crypto.py
+++ b/OpenSSL/crypto.py
@@ -394,6 +394,10 @@ class X509Name(object):
         return _lib.X509_NAME_hash(self._name)
 
 
+    def __hash__(self):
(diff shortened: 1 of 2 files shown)


alex:
This hasn't been updated in a while. At this point we aren't expanding the X509 APIs, please use cryptography instead.

PR #96, opened 2014-04-16
Add EC support to PKey

mrjefftang:
Depends on pyca/cryptography#933


exarkun:
Why is this closed?

mrjefftang:
I had to clear out my repository so the branch it referred to no longer existed.

Also I'm thinking there's probably a better way of going about this such as creating a subclass of PKey like ECPKey which just overrides the necessary methods.

PR #110, opened 2014-05-10
Next Protocol Negotiation redux

Lukasa:
This is the same as #86, but with all the changes that @exarkun requested in place.

Tests pass for me locally, let's hope Travis is happy.

### DIFF
--- a/OpenSSL/SSL.py
+++ b/OpenSSL/SSL.py
@@ -293,6 +293,10 @@ class Context(object):
         self._info_callback = None
         self._tlsext_servername_callback = None
         self._app_data = None
+        self._npn_advertise_callback = None
+        self._npn_advertise_callback_args = None
+        self._npn_select_callback = None
+        self._npn_select_callback_args = None
 
         # SSL_CTX_set_app_data(self->ctx, self);
         # SSL_CTX_set_mode(self->ctx, SSL_MODE_ENABLE_PARTIAL_WRITE |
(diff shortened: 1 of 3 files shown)


OpenSSL/SSL.py
```diff
@@ -809,6 +812,92 @@ def wrapper(ssl, alert, arg):
...
+
+        :param callback: The callback function.  It will be invoked with one
+            argument, the Connection instance.  It should return a list of
+            bytestrings representing the advertised protocols, like
+            ``[b'http/1.1', b'spdy/2']``.
+        """
+        @wraps(callback)
+        def wrapper(ssl, out, outlen, arg):
+            conn = Connection._reverse_mapping[ssl]
+            protos = callback(conn)
```
exarkun: This call probably merits some exception handling so that Python-level exceptions are turned in to OpenSSL-level error codes?  Also test coverage for that case.

OpenSSL/SSL.py
```diff
@@ -809,6 +812,92 @@ def wrapper(ssl, alert, arg):
...
+            instr = _ffi.buffer(in_, inlen)[:]
+            protolist = []
+            while instr:
+                l = indexbytes(instr, 0)
+                proto = instr[1:l+1]
+                protolist.append(proto)
+                instr = instr[l+1:]
+
+            # Call the callback
+            outstr = callback(conn, protolist)
```
exarkun: Similar comment about exception handling here.

OpenSSL/SSL.py
```diff
@@ -1550,6 +1646,16 @@ def get_cipher_version(self):
...
+    def get_next_proto_negotiated(self):
+        """
+        Get the protocol that was negotiated by NPN.
+        """
+        data = _ffi.new("unsigned char **")
+        data_len = _ffi.new("unsigned int *")
+
+        _lib.SSL_get0_next_proto_negotiated(self._ssl, data, data_len)
+
+        return _ffi.buffer(data[0], data_len[0])[:]
```
exarkun: Who is responsible for the memory (in particular, releasing it) used by the protocol name in `data`?

exarkun:
Sorry about the long delay in getting to this.  Thanks for your work so far Lukasa.  Are you still interested in getting this into pyOpenSSL?  If so I'll be sure to get back to you about further revisions more quickly.

Lukasa:
@exarkun I'm happy to keep moving this forward. =) I'd rather this got merged than abandon the patch, I'm probably best placed to keep it going.

OpenSSL/SSL.py
Lukasa: What worries me is debuggability. Swallowing exceptions here will make it really tricky for developers to spot their mistakes. Ideally I'd call `log.exception` before swallowing it, but I'm a bit nervous about adding logging where there previously was none.

Do you have any ideas?

OpenSSL/SSL.py
Lukasa: We are (we allocated it, we have to free it). However, it should be freed up automatically because `data` will go out of scope at the bottom of the function (CPython will free it immediately, PyPy will get around to it). That's fine, because `_ffi.buffer()` will have copied what we care about out already.

OpenSSL/SSL.py
exarkun: > What worries me is debuggability. Swallowing exceptions here will make it really tricky for developers to spot their mistakes.

Yep.  I'd definitely like to see something better than just swallowing the exception.

In other places, pyOpenSSL saves the exception and re-raises it at a later point where OpenSSL isn't going to get in the way.  This can make sense in cases where "a later point" is obvious and soon - for example, certificate verification happens in response to a `SSL_read` call (typically).  So it's somewhat straightforward to save verification exceptions and raise them from `Connection.recv` (which is the wrapper for `SSL_read`).

Does NPN fit into a model like that?  Maybe NPN is advertised in response to an `SSL_read` call as well?  After the server receives the `ClientHello`, perhaps?  Take a look at `_VerifyHelper` to see how this is implemented for the certificate verification callback.  I'm not sure any of the code is re-usable (at least, not as currently written) but maybe it will give you some ideas.

Lukasa: NPN is handled in the handshake, which means the appropriate place to raise them would be in `Connection.connect`, `Connection.connect_ex` (I think? I'm not entirely sure about what this method is actually for), and `Connection.accept`. I think that the approach in the `_VerifyHelper` will work here for these methods. I'll prototype it up and see.

exarkun: Cool.  Keep in mind that it's also possible to drive the handshake manually via `do_handshake`.

Lukasa:
Ok @exarkun, I believe I've made the changes you requested in 0ea76e7d977b19f2bb4ce4dee9bee8aa179eaff0. =)

Lukasa:
\o/ Thanks @exarkun!

PR #115, opened 2014-05-31
Set the string-mask to utf8only

Spindel:
If subject had utf-8 characters in them, the encoding chosen by OpenSSL for
defaults T61.

Requires changes in cryptography, commit id f8561cdf06f163806a57b0dd36290192414e3496

From the OpenSSL source code:
	 * utf8only : only use UTF8Strings (RFC2459 recommendation for 2004).

That was 10 years ago, and the last remnant that had problems with it
was Netscape, which is no longer a problem.

A request changes from:
   13:d=5  hl=2 l=   3 prim: OBJECT            :commonName
   18:d=5  hl=2 l=   9 prim: T61STRING         :Gurka ���

To:
   13:d=5  hl=2 l=   3 prim: OBJECT            :commonName
   18:d=5  hl=2 l=  12 prim: UTF8STRING        :Gurka åäö

OpenSSL/test/test_crypto.py
	Update test DER data to have utf8string.
	( \x0c instead of \0x13, PrintableString )

### DIFF
--- a/OpenSSL/crypto.py
+++ b/OpenSSL/crypto.py
@@ -456,9 +456,6 @@ class X509Name(object):
         if isinstance(value, _text_type):
             value = value.encode('utf-8')
 
-        # Make it so OpenSSL generates utf-8 strings.
-        _lib.ASN1_STRING_set_default_mask_asc(b'utf8only')
-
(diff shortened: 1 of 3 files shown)


reaperhulk:
This has been rebased and is in #234.

PR #127, opened 2014-06-19
Add a meta testenv for metadata: style & packaging

lvh:
This adds some features for tox to test some meta stuff about PyOpenSSL. Specifically:

- pyflakes
- check-manifest
- pip-review, which notifies us when there are newer packages than what we install in virtualenvs
- pyroma, which should keep us generally informed about how good a packaging citizen we're being.

This isn't ready to be merged. It fails, and not as a false positive: MANIFEST.in is incomplete, and there's a few pyflakes warnings. I'm happy to do these on this PR, or on separate PRs.

PR #133, opened 2014-06-23
Tox-powered Travis configuration

lvh:
This makes Travis run a bunch of tox environments. This is good, because it enables moving more behavior into tox. That's good, because it enables running something a lot closer to what CI runs on your own machine.

Refs #127.


reaperhulk:
obsoleted by #230

Issue #139, opened 2014-07-10
X509Extension has a memory leak in its private member function _subjectAltNameString() which is publicly exposed via the member function __str__()


crc32a:
just calling the member method __str__() on an X509Extention instance object will trigger a memory leak when the extension is a SubjectAltName. IE when __str__() is called and the condition  if _lib.NID_subject_alt_name == self._nid triggers return self._subjectAltNameString()

proposed solution. Not sure I suspect GENERAL_NAMES_free may need to be registered as a garbage collection function at some point. I also noticed the GENERAL_NAMES_free is has no bindings in cryptography hazmat. :( Investigating.

hynek:
Fixed via #252

PR #140, opened 2014-07-10
Fix for Alt Subject Name memory leak.

crc32a:
fixes https://github.com/pyca/pyopenssl/issues/139


reaperhulk:
This leak has been fixed in #252

Issue #136, opened 2014-06-27
Add ability to dump public key


zbyte64:
One should be able to:

```
dump_publickey(crypto.FILETYPE_PEM, pkey)
```

Resurrected from:
https://code.launchpad.net/~shanemhansen/pyopenssl/dump_publickey

PR #141, opened 2014-07-14
add crypto method for dumping the public key

zbyte64:
Adds dump_publickey for #136


exarkun:
Thanks again.  Please re-open when tests have been added.

PR #143, opened 2014-07-22
First attempt at creating a more GeneralNames based  X509SubjectAltnativeName class

crc32a:
The tests will fail untill the X509_NAME_free and X509V3_EXT_d2i are exposed from the cryptography project.  This new get_subject_alt_name method will split up the alt Names into seperate components.


alex:
This hasn't been updated for a while; at this point we aren't going to be changing the X509 APIs much, instead you should use cryptography's X509 API.

PR #147, opened 2014-08-12
Handle when an OpenSSL error doesn't contain a reason

alex:
(Or any other field)

You can reproduce the error by running:

```
treq.get('https://nile.ghdonline.org')
```

from within a twisted program (and doing the approrpiate deferred stuff).

I'm unsure how to craft a unit test for this

### DIFF
--- a/OpenSSL/_util.py
+++ b/OpenSSL/_util.py
@@ -7,6 +7,8 @@ lib = binding.lib
 
 def exception_from_error_queue(exceptionType):
     def text(charp):
+        if not charp:
+            return ""
         return native(ffi.string(charp))
 
     errors = []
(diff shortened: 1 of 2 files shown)


hynek:
This looks useful,  correct and the builders agree.  Anything more needed?

alex:
It feels RFC to me :-)

glyph:
I'm inclined to agree with @hynek .  Anything I can do to move this along?

posita:
> It feels RFC to me :-)
...
> I'm inclined to agree with @hynek . Anything I can do to move this along?

Does this need a `merge master` first? It looks like trivial conflicts were introduced by de0754690e01202bb4bd7e47c529db87765fc805 et seq.

glyph:
Yay!

Issue #154, opened 2014-08-27
No way to validate/verify chain.


sholsapp:
It seems I cannot do the equivalent of:

``` bash
$ openssl verify -CAfile my-ca-bundle.crt foo.cert
```

All of the cryptography ffi bindings are available to make this possible, so I'm going to start working on a PR that adds support for this in pyOpenSSL. I'm thinking that it makes sense to add something like `OpenSSL.crypto.verify_chain(cert, chain, trust)` makes most sense as it matches the  args to OpenSSL's `X509_STORE_CTX_init` function.

If there are any concerns/opinions, let me know!

PR #155, opened 2014-08-28
Add OpenSSL.crypto.verify_cert method.

sholsapp:
This change adds support for verifying a certificate or a certificate
chain. This implementation uses OpenSSL's underlying X509_STORE_CTX_*
class of functions to accomplish this.

This change also adds an intermediate signing certificate/key and a
service certificate/key signed with the intermediate signing
certificate, to make testing the OpenSSL.crypto.verify_chain method
easier to test. I figured I would add it to the top level module so
other people can use an intermediate signing certificate in their own
tests.

Issue: https://github.com/pyca/pyopenssl/issues/154

### DIFF
--- a/OpenSSL/crypto.py
+++ b/OpenSSL/crypto.py
@@ -25,13 +25,43 @@ TYPE_RSA = _lib.EVP_PKEY_RSA
 TYPE_DSA = _lib.EVP_PKEY_DSA
 
 
+
 class Error(Exception):
     """
     An error occurred in an `OpenSSL.crypto` API.
     """
 
 
+
+def _exception_from_context_error(exception_type, store_ctx):
+    """
+    Convert a :py:func:`OpenSSL.crypto.verify_cert` failure into a Python
+    exception.
+
+    When a call to native OpenSSL X509_verify_cert fails, additonal information
+    about the failure can be contained from the store context.
+    """
+
+    errors = [
+        _lib.X509_STORE_CTX_get_error(store_ctx._store_ctx),
+        _lib.X509_STORE_CTX_get_error_depth(store_ctx._store_ctx),
+        _native(_ffi.string(_lib.X509_verify_cert_error_string(_lib.X509_STORE_CTX_get_error(store_ctx._store_ctx)))),
+    ]
+    _x509 = _lib.X509_STORE_CTX_get_current_cert(store_ctx._store_ctx)
+    if _x509 != _ffi.NULL:
+      _cert = _lib.X509_dup(_x509)
+      pycert = X509.__new__(X509)
+      pycert._x509 = _ffi.gc(_cert, _lib.X509_free)
+    e = exception_type(errors)
+    e.certificate = pycert
+    raise e
+
+
+
 _raise_current_error = partial(_exception_from_error_queue, Error)
+_raise_context_error = partial(_exception_from_context_error, Error)
+
+
 
 def _untested_error(where):
     """
@@ -1357,6 +1387,44 @@ X509StoreType = X509Store
 
 
 
+class X509StoreContext(object):
+    """
+    An X.509 store context.
+
+    A :py:class:`X509StoreContext` is used to verify a certificate in some
+    context in conjunction with :py:func:`verify_cert`. The information
+    encapsulated in this object includes, but is not limited to, a set of
+    trusted certificates, verification parameters and revoked certificates.
+
+    :param store: A :py:class:`X509Store` of trusted certificates.
+    :param cert: An :py:class:`X509` certificate to be validated during a
+      subsequent call to :py:func:`verify_cert`.
+    """
+
+    def __init__(self, store, cert):
+        store_ctx = _lib.X509_STORE_CTX_new()
+        self._store_ctx = _ffi.gc(store_ctx, _lib.X509_STORE_CTX_free)
+        self._store = store
+        self._cert = cert
+
+    def _init(self):
+        """
+        Set up the store context for a subsequent verification operation.
+        """
+        ret = _lib.X509_STORE_CTX_init(self._store_ctx, self._store._store, self._cert._x509, _ffi.NULL)
+        if ret <= 0:
+            _raise_current_error()
+
+    def _cleanup(self):
+        """
+        Internally cleans up the store context.
+
+        The store context can then be reused with a new call to
+        :py:meth:`init`.
+        """
+        _lib.X509_STORE_CTX_cleanup(self._store_ctx)
+
+
 def load_certificate(type, buffer):
     """
     Load a certificate from a buffer
@@ -2303,6 +2371,19 @@ def verify(cert, signature, data, digest):
     if verify_result != 1:
         _raise_current_error()
 
+
+def verify_cert(store_ctx):
+    """
+    Verify a certificate in a context.
+
+    :param store_ctx: The :py:class:`X509StoreContext` to verify.
+    :raises: Error
+    """
+    store_ctx._init()
+    ret = _lib.X509_verify_cert(store_ctx._store_ctx)
+    store_ctx._cleanup()
+    if ret <= 0:
+        _raise_context_error(store_ctx)
 
 
 def load_crl(type, buffer):
--- a/OpenSSL/crypto.py
+++ b/OpenSSL/crypto.py
@@ -1404,6 +1404,10 @@ class X509StoreContext(object):
         self._store_ctx = _ffi.gc(store_ctx, _lib.X509_STORE_CTX_free)
         self._store = store
         self._cert = certificate
+        # Make the store context available for use after instantiating this
+        # class by initializing it now. Per testing, subsequent calls to
+        # :py:meth:`_init` have no adverse affect.
+        self._init()
 
 
     def _init(self):
@@ -1420,7 +1424,7 @@ class X509StoreContext(object):
         Internally cleans up the store context.
 
         The store context can then be reused with a new call to
-        :py:meth:`init`.
+        :py:meth:`_init`.
         """
         _lib.X509_STORE_CTX_cleanup(self._store_ctx)
 
@@ -1448,6 +1452,16 @@ class X509StoreContext(object):
         return X509StoreContextError(errors, pycert)
 
 
+    def set_store(self, store):
+        """
+        Set the context's trust store.
+
+        :param X509Store store: The certificates which will be trusted for the
+            purposes of any *future* verifications.
+        """
+        self._store = store
+
+
     def verify_certificate(self):
         """
         Verify a certificate in a context.
@@ -1455,6 +1469,8 @@ class X509StoreContext(object):
         :param store_ctx: The :py:class:`X509StoreContext` to verify.
         :raises: Error
         """
+        # Always re-initialize the store context in case
+        # :py:meth:`verify_certificate` is called multiple times.
         self._init()
         ret = _lib.X509_verify_cert(self._store_ctx)
         self._cleanup()
--- a/OpenSSL/crypto.py
+++ b/OpenSSL/crypto.py
@@ -1439,11 +1439,12 @@ class X509StoreContext(object):
             _native(_ffi.string(_lib.X509_verify_cert_error_string(
                         _lib.X509_STORE_CTX_get_error(self._store_ctx)))),
         ]
+        # A context error should always be associated with a certificate, so we
+        # expect this call to never return :class:`None`.
         _x509 = _lib.X509_STORE_CTX_get_current_cert(self._store_ctx)
-        if _x509 != _ffi.NULL:
-          _cert = _lib.X509_dup(_x509)
-          pycert = X509.__new__(X509)
-          pycert._x509 = _ffi.gc(_cert, _lib.X509_free)
+        _cert = _lib.X509_dup(_x509)
+        pycert = X509.__new__(X509)
+        pycert._x509 = _ffi.gc(_cert, _lib.X509_free)
         return X509StoreContextError(errors, pycert)
 
 
--- a/OpenSSL/crypto.py
+++ b/OpenSSL/crypto.py
@@ -1397,16 +1397,27 @@ class X509StoreContext(object):
     encapsulated in this object includes, but is not limited to, a set of
     trusted certificates, verification parameters and revoked certificates.
 
-    :param store: An :py:class:`X509Store` of trusted certificates.
-    :param cert: An :py:class:`X509` certificate to be validated during a
-        subsequent call to :py:func:`verify_cert`.
+    :ivar _store_ctx: The underlying X509_STORE_CTX structure used by this
+        instance.  It is dynamically allocated and automatically garbage
+        collected.
+
+    :ivar _store: See the ``store`` `__init__`` parameter.
+
+    :ivar _cert: See the ``certificate`` `__init__`` parameter.
     """
 
-    def __init__(self, store, cert):
+    def __init__(self, store, certificate):
+        """
+        :param X509Store store: The certificates which will be trusted for the
+            purposes of any verifications.
+
+        :param X509 certificate: The certificate to be verified.
+        """
         store_ctx = _lib.X509_STORE_CTX_new()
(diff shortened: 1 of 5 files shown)


sholsapp:
Sorry for spamming!

@exarkun this one is good to go, right?

exarkun:
Sorry, I'm a little behind in pyOpenSSL maintenance.  This is on my todo list and I'll get to it ASAP.

OpenSSL/crypto.py
```diff
@@ -25,13 +25,43 @@
...
+        _lib.X509_STORE_CTX_get_error_depth(store_ctx._store_ctx),
+        _native(_ffi.string(_lib.X509_verify_cert_error_string(_lib.X509_STORE_CTX_get_error(store_ctx._store_ctx)))),
+    ]
+    _x509 = _lib.X509_STORE_CTX_get_current_cert(store_ctx._store_ctx)
+    if _x509 != _ffi.NULL:
+      _cert = _lib.X509_dup(_x509)
+      pycert = X509.__new__(X509)
+      pycert._x509 = _ffi.gc(_cert, _lib.X509_free)
+    e = exception_type(errors)
+    e.certificate = pycert
```
exarkun: I wonder if this should just be operating on a new, fixed exception type.  For example, `X509StoreContextError` or `CertificateVerificationError`.  The former is more regular (it's an exception from an X509_STORE_CTX-related API) but sounds sort of nonsensical.  The latter seems like it makes sense but if it turns out later that some API unrelated to certificate verification is going to want to use this X509StoreContext helper function then it will end up being nonsense.

Any thoughts on this?

exarkun: Oh, and related to that, one reason it would be nice to introduce a new exception type here is so the code can be `raise CertificateVerificationError(certificate, errors)` instead of setting an extra attribute on an exception type that doesn't normally have that attribute.

OpenSSL/crypto.py
```diff
@@ -2304,6 +2372,19 @@ def verify(cert, signature, data, digest):
...
+def verify_cert(store_ctx):
+    """
+    Verify a certificate in a context.
+
+    :param store_ctx: The :py:class:`X509StoreContext` to verify.
+    :raises: Error
+    """
+    store_ctx._init()
+    ret = _lib.X509_verify_cert(store_ctx._store_ctx)
+    store_ctx._cleanup()
```
exarkun: I think I'm leaning towards suggestion that `verify_cert` just be a method of `X509StoreContext`.  That way it can be `self._init()` and `self._cleanup()` which is nicer than accessing private methods of a non-self object.

Also possibly the `_cleanup` call should be in a `finally` block - though I can't suggest how to write a unit test that demonstrates that this matters.

OpenSSL/crypto.py
```diff
@@ -25,13 +25,43 @@
...
+    about the failure can be contained from the store context.
+    """
+
+    errors = [
+        _lib.X509_STORE_CTX_get_error(store_ctx._store_ctx),
+        _lib.X509_STORE_CTX_get_error_depth(store_ctx._store_ctx),
+        _native(_ffi.string(_lib.X509_verify_cert_error_string(_lib.X509_STORE_CTX_get_error(store_ctx._store_ctx)))),
+    ]
+    _x509 = _lib.X509_STORE_CTX_get_current_cert(store_ctx._store_ctx)
+    if _x509 != _ffi.NULL:
```
exarkun: `pycert` doesn't get initialized outside of this `if` block.  So if `X509_STORE_CTX_get_current_cert` ever returns NULL then this code will raise a `NameError` a few lines below.

Can this return NULL if the certificate has been set to a non-NULL value?  If so, a test exercising that case would be great.  If not ... maybe the `if` check here is not necessary.  In C that might lead to a NULL dereference (and a possible security issue) if we ever get surprised by a NULL return value.  In Python it just means we'll get an `AttributeError` or a `TypeError` or something on one of the following lines.  For a case we thought was impossible, that's not a terrible failure mode.

exarkun:
Sorry about the delay in getting back to this (had some uninvited surprises over the holidays that left me with much less free time than I expected).

I submitted a PR against this branch for some straightforward, boring changes I think make sense.  I left a few more inline comments about some areas I'm not quite as sure about but that I think merit a little bit more thought.

Thanks again.

OpenSSL/crypto.py
sholsapp: I think that is how we originally thinking about implementing this but stopped once we realized that this is _not_ how the actual OpenSSL API is exposed. That is, the OpenSSL API has this `verify_cert` method as a top-level function.

With that said, I'm happy to move it to see how it looks, just wanted to remind us why we chose to do it this way a couple months ago. :)

OpenSSL/crypto.py
sholsapp: I think this is relevant. =P

> There are only two hard things in Computer Science: cache invalidation and naming things.
> 
> -- Phil Karlton

My gut wants to make us stick with the OpenSSL API and just name this thing a `X509StoreContextError`, but from poking around the OpenSSL code base I can only see this object being used for verification purposes (albeit, verification in lots of different places: `pkcs7`, `ocsp`, `x509_vfy`, ...). I think I should do more research about the naming and use of this thing. I'll let you know what I find.

As for the coding suggestions, they all make sense.

OpenSSL/crypto.py
exarkun: Thanks for pointing that out.  Sorry about the inconsistent feedback (I'm definitely having trouble remembering all of the discussions and suggestions made on this issue so far).

Fortunately it's all written down, hooray.  I went back and read some older comments.  One thing I recognized that somehow I'd overlooked recently is that the wrapped API is `X509_verify_cert`.  Going by the pyOpenSSL convention, that suggests this is actually a method of `X509`.  Too bad that doesn't actually make sense!  Because the only argument accepted is an `X509StoreContext`.  Argh.  I think OpenSSL just named the function wrong and they should have called it `X509_STORE_CTX_verify_cert`.  Alternatively, the X509 itself shouldn't be part of the X509_STORE_CTX and the signature should have been `X509_verify_cert(x509, x509_store_ctx)`.  But I suppose wishing the API and naming made more sense in OpenSSL doesn't help us at all.

I think I prefer the conclusion that `X509_verify_cert` is really an OpenSSL API that concerns itself first and foremost with `X509_STORE_CTX` structures (silly as that structure may be) and so the proper translation of this API into pyOpenSSL is as an `X509StoreContext` method.

There is precedent in pyOpenSSL for methods that are slightly more convenient than the (straightforward Python translation of the) OpenSSL APIs, though (eg `load_privatekey`).  So perhaps some good **follow-up** work would be a top-level pyOpenSSL `verify_cert` (or `verify_certificate` - I kinda dislike ad hoc abbreviations :smile:) that takes a handful of arguments, makes the `X509StoreContext` instance for you and calls its `verify_cert` method.  Most people will probably not care that `X509StoreContext` happens to be a thing (but exposing it is still a good idea for more advanced users).

Does that make sense and sound like a good way forward?

OpenSSL/crypto.py
sholsapp: FYI, until we find that next use case I'm going to add this as `X509StoreContextError` and it will take errors and a certificate as parameters. It will be raised from `X509StoreContext.verify_certificate`.

OpenSSL/crypto.py
sholsapp: I think that this all makes good sense and applied these changes in the latest diff. I'll make a note on my todo list to think about a nice top level verify_certificate method because it does feel strange with the current OpenSSL API. :)

OpenSSL/crypto.py
sholsapp: Oh, jeez, good catch.

Admittedly, I'm not sure how I can trigger this case without some pretty contrived mocking. I'm not inclined to remove the check, though, because the documents at https://www.openssl.org/docs/crypto/X509_verify_cert.html are pretty unclear what "exceptional circumstances" might be.

What if we just do something like

     _x509 = _lib.X509_STORE_CTX_get_current_cert(self._store_ctx)                                                                                                                                             
    if _x509 != _ffi.NULL:                                                                                                                                                                                    
        _cert = _lib.X509_dup(_x509)                                                                                                                                                                            
        pycert = X509.__new__(X509)                                                                                                                                                                             
        pycert._x509 = _ffi.gc(_cert, _lib.X509_free)                                                                                                                                                           
        return X509StoreContextError(errors, pycert) 
    else:
        Error('Exception context error?')

sholsapp:
Sorry it has been taking so long for this, I've been caught up with packaging and "relocatable" virtualenvs at work. :) Let me know what you think of this so far.

And I've somehow dorked up my branch when I merged your changes in ... I'll try fixing that up soon. Sorry!

sholsapp:
Okay, @exarkun, so what's left for me. I have to run, so this is mostly a list for me when I'm back. :)

- Need to update the documentation to use the new `X509StoreContext` method instead of top level `verify_cert`.
- Need to work out the error handling that we want for the uninitialized `pycert` variable.

sholsapp:
@exarkun, I made some last changes, let me know what you think regarding the exception documentation.

Regarding the error checking on the return value of `X509_STORE_CTX_get_current_cert`, I looked at OpenSSL's use of this function and it appears that they don't do any error checking for this function either. I think the reasoning here is that a context error always has to have a certificate associated with it (even if it is the certificate that you're passing to be verified). So, I removed the check, bring `pycert` into scope, and added a inline doc stating the decision. Like always, happy to change where needed if we don't like this approach after all.

Let me know if there is anything else. I'm looking forward to using this new stuff in my tools!

OpenSSL/crypto.py
```diff
@@ -1356,6 +1359,109 @@ def add_cert(self, cert):
...
+
+
+    def verify_certificate(self):
+        """
+        Verify a certificate in a context.
+
+        :param store_ctx: The :py:class:`X509StoreContext` to verify.
+        :raises: Error
+        """
+        self._init()
```
mrjefftang: Why do you wait till the call to verify_certificate() to call _init()?

I'm working on a case where I need to modify the check_time for an expired certificate and the call to X509_STORE_CTX_get0_param returns NULL because it's not initialized yet.

sholsapp: Hi @mrjefftang, this was an artifact of the previous implementation where `verify_cert[ificate]` was a top level function instead of a method on the `X509StoreContext` object. Now, the implementation can and *should* do that.

I'll make the change later tonight. Thanks for using/testing the code!

sholsapp: Hey @mrjefftang, I pushed out a change that initializes the underlying objects at `X509StoreContext` instantiation time. Let me know if this behaves as you would expect.

OpenSSL/crypto.py
```diff
@@ -1356,6 +1359,125 @@ def add_cert(self, cert):
...
+    def verify_certificate(self):
+        """
+        Verify a certificate in a context.
+
+        :param store_ctx: The :py:class:`X509StoreContext` to verify.
+        :raises: Error
+        """
+        # Always re-initialize the store context in case
+        # :py:meth:`verify_certificate` is called multiple times.
+        self._init()
```
mrjefftang: Calling _init() again here throws out any changes to verification parameters made between X509Store() and verify_certificate().

Test case:
```python
import time
import calendar

from OpenSSL.crypto import (load_certificate,
                            FILETYPE_PEM,
                            X509Store,
                            X509StoreContext,
                            X509StoreContextError)
from OpenSSL._util import (ffi as _ffi,
                           lib as _lib)
from OpenSSL.test.test_crypto import (root_cert_pem,
                                      intermediate_cert_pem,
                                      intermediate_server_cert_pem)

root_cert = load_certificate(FILETYPE_PEM, root_cert_pem)
intermediate_cert = load_certificate(FILETYPE_PEM, intermediate_cert_pem)
server_cert = load_certificate(FILETYPE_PEM, intermediate_server_cert_pem)

# Root CA with valid time
store = X509Store()
store.add_cert(root_cert)
store_ctx = X509StoreContext(store, root_cert)
store_ctx.verify_certificate()

# Root CA with invalid time
store = X509Store()
store.add_cert(root_cert)
store_ctx = X509StoreContext(store, root_cert)

expire_time = root_cert.get_notAfter().decode('utf-8')
cert_time = time.strptime(expire_time, '%Y%m%d%H%M%SZ')
invalid_time = calendar.timegm(cert_time)
vfy_param = _lib.X509_STORE_CTX_get0_param(store_ctx._store_ctx)
_lib.X509_VERIFY_PARAM_set_time(vfy_param, invalid_time)
try:
    store_ctx.verify_certificate()
except X509StoreContextError as e:
    assert e.args[0][0] == _lib.X509_V_ERR_CERT_HAS_EXPIRED
    assert e.args[0][1] == 0 
```

I thought about modifying verify_certificate() as follows but it is still not perfect because every time you call verify_certificate() it throws away the parameters at the end of the call but at least it won't crash Python 3.4.2 in a segfault with multiple calls.
```python
    def verify_certificate(self):
        """
        Verify a certificate in a context.

        :param store_ctx: The :py:class:`X509StoreContext` to verify.
        :raises: Error
        """
        try:
            ret = _lib.X509_verify_cert(self._store_ctx)
            if ret <= 0:
                raise self._exception_from_context()
        finally:
            self._cleanup()
            # Always re-initialize the store context in case
            # :py:meth:`verify_certificate` is called multiple times.
            self._init()
```

sholsapp: Yep, it's quite an awkward dance, here.

What makes most sense when one calls `init`... for it to reset the store and all of the parameters you've specified... or for it to remember the store and all of the parameters you've specified? If the latter, the only way I can think to achieve that would be to implement `X509_STORE_CTX_get0_param`,  `X509_STORE_CTX_get0_param` or any other function you're using within `X509StoreContext` class itself so that it can remember and reapply configuration. That is the current approach I'm taking and is why multiple calls to `init` are okay (because the class can reapply the entire configuration every time you call init). It doesn't work for you because you're operating outside of the `X509StoreContext` class itself.

This was actually my original design intention. This commit was just the beginning of a more featureful `X509StoreContext` class that would implement all of the API you're using. The next step would be implementing the methods that you're using in `X509StoreContext` so that it could do what you want.

Thoughts on this approach?

mrjefftang: In that case, your approach makes sense in order to achieve a pythonic approach.  My only concern is that theres quite a few options to set on X509_VERIFY_PARAM and that could open up a rats nest.

I'm kind of straddling the boundary of Python and FFI calls because not all of it has been wrapped in Python.

sholsapp: I agree there are a ton of options. In the case of the verify params one, we can make that fairly generic (these are just flags to pass to the same function). The remaining functions that would need to be wrapped in `X509StoreContext` are:

- X509_STORE_CTX_set_chain
- X509_STORE_CTX_set0_crls
- X509_STORE_CTX_get0_param
- X509_STORE_CTX_set0_param
- X509_STORE_CTX_set_default

We'll also need to have some abstraction over all of the parameters that you can use (https://www.openssl.org/docs/crypto/X509_VERIFY_PARAM_set_flags.html). I think we can come up with a clean solution.

The big question is if we want to admittedly not support your use case until this work is done or not? @exarkun, do you have any opinions on this one?

mrjefftang: I'm ok with it not being supported yet since I can call the bindings.

I think it's more important to be able to work in the normal use case which is building up a certificate chain and verifying it.

kislyuk: :+1:, I can't wait to be able to verify a peer-supplied certificate chain without having to dive into the guts of pyopenssl. (I would also love to have the time setting capability for test suites, but that's a lower priority.)

sholsapp:
@exarkun I'm still here (I know you're busy, just a ping).

I think @mrjefftang and @kislyuk want this, too. Once it gets in I'll start finishing implementation for @mrjefftang asks (see above).

Thanks!

kislyuk:
@exarkun, any chance we can get this approved for merging in?

PR #170, opened 2014-11-12
add `get_digest` method to `X509Req`

bwhmather:
(depends on bwhmather/cryptography@1b55947c).
Writing a simple ca and need to be able verify that a certificate request received by the server is the same as the one sent by the client.  Comparing digests seems like the obvious way to do it.

### DIFF
--- a/src/OpenSSL/crypto.py
+++ b/src/OpenSSL/crypto.py
@@ -962,6 +962,33 @@ class X509Req(object):
 
         return result
 
+    def digest(self, digest_name):
+        """
+        Return the digest of the X509Req object.
+
+        :param digest_name: The name of the digest algorithm to use.
+        :type digest_name: :py:class:`bytes`
+
+        :return: The digest of the object
+        """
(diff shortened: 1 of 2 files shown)


elitest:
There have been a few releases of cryptography since this PR was submitted.  Perhaps we should get this retested?

bwhmather:
https://github.com/pyca/cryptography/pull/1472 has been merged so it should indeed work now.  I figured I would need to wait for the release before bumping here and then forgot...  Sorry

hynek:
This needs to be rebased and a changelog entry.

bwhmather:
Apparently default digest was changed in 0.9.8 to SHA1.  This is causing the tests to fail.  Will figure out what the digest should be and fix the tests.

hynek:

reaperhulk:
At this point this functionality can be obtained via converting to a cryptography object (`to_cryptography`) and computing a digest over the `tbs_certrequest_bytes`. Apologies for mishandling this PR 😞

bwhmather:
Not a problem.  Thank you for following up.

PR #174, opened 2014-11-25
Update build_script

JensTimmerman:
ps make targets don't exist anymore in the new doc Makefile

### DIFF
--- a/rpm/build_script
+++ b/rpm/build_script
@@ -1,1 +1,1 @@
-make -C doc text ps html
+make -C doc text html
--- a/setup.cfg
+++ b/setup.cfg
@@ -7,7 +7,7 @@ universal = 1
 # bdist_rpm settings contributed by Mihai Ibanescu <misa@redhat.com>
 [bdist_rpm]
 release = 1
-build-requires = lynx openssl-devel python-devel python-sphinx
+build-requires = openssl-devel python-devel python-sphinx
 group = Development/Libraries
 build_script = rpm/build_script
 doc-files = doc/_build/html


exarkun:
Thanks.

Removing `ps` makes sense to me.  There is no `ps` target at all and so that can never work.

Removing `man` makes less sense to me.  Maybe `setup.py` should define a `doc` extra that declares a Sphinx >= 1.0 dependency.  On the other hand, pyOpenSSL includes no man pages.  So perhaps it doesn't matter.

On the other hand, I have no idea what `rpm/build_script` is actually used for.  I don't know why it's in the repository or what the benefits of having it working (or the benefit of keeping it at all) are...

Since you noticed this problem, perhaps you know what it's good for? :smile:

JensTimmerman:
In our environment we create rpm's of python packages on a few build servers, using `python setup.py bdist_rpm`
These then get put in our repositories and installed on our servers.

`setup.py bdist_rpm` triggers the run of this build_script, which failed because of the nonexisting ps target.
 
I was able to create a working rpm of my branch, (so without the ps target)
The man wasn't there in the first place, I tried adding it, but it failed because of a dependency, and I didn't want to put time into installing the dependency on our build servers. (and this would indeed require patching of the setup.py)

So, on the down side, using bdist_rpm still doesn't create man pages in the rpm (but I do not really need them, not if there is a text and html version on the system)
on the up side, something that doesn't work without this patch now works.

exarkun:
Thanks for explaining that!

It sounds like another good change to make would be to have travis run `setup.py bdist_rpm` to ensure this continues to work in the future.  I ... think ... I know how to do that easily, I'll submit a PR against this branch to add that change if I can work it out.

> I am willing to rebase these 3 commits into 1 if you prefer this, I was quickly testing stuff using the web interface, resulting in 3 tries.

That's okay.  I generally prefer accurate history to idealized, fictionalized history. :smile:   I can see the benefits to the latter but they mostly require **everyone** to create perfect histories and that's a lot more work I'm not going to ask every pyOpenSSL contributor to do.

hynek:
Could you rebase and add a target to tox.ini that exercises bdist_rpm please?

JensTimmerman:
@hynek : squached and rebased on top of master.
I'm not sure how to create a tox.ini that works on all systems, `python setup.py bdist_rpm` depends on the 'rpm' binary being installed, I'm not familliar enought with tox to figure out how to set the rpm binary as a dependency.

hynek:
or you on an rpm-based OS?  if not, there’s packages for all of them so I would just presume the presence of rpm and add the dep to Travis (which is Ubuntu based).

hynek:
Ugh, can you read that failure?  Unless it’s trivial to fix I think it’s not worth the trouble and we can't have ~~nice things~~ tests for it. :(

JensTimmerman:
I'm not very familliar with the tox.ini/jenkins syntax, it works on my end with the `requires='/'` but the quotes around the `\` seem to disappear for some reason, might need to escape them?

hynek:
ah meh reset to 35204f753e05248efe0098aaf54526525678efce and we’ll merge it.  This amount of complexity is not worth it.  if it breaks, someone will complain (like you did :)).

JensTimmerman:
@hynek reset, I've left it at cc44fcd, since I can't figure out why lynx would be needed, I don't have it on my system and it works.

hynek:
thanks!

Issue #62, opened 2014-03-11
disparity between methods accepting filenames as `bytes` or `str`


tisdall:
I upgraded from 0.13.1 to 0.14 and got several errors. One was that several methods that took filenames have been changed to only accept binary strings instead of regular unicode strings.  So, in py3, `bytes` instead of `str`.

I think 0.13.1 actually had the opposite situation for some of those same methods where it only accepted a unicode string and not a binary string.

I think the underlying C library uses binary strings (ie no encoding enforced), so it makes sense to use that (on *nix).  However, the python wrapper should be able to accept both and translate accordingly.

exarkun:
Considering the direction Python has chosen for handling paths, I think the sensible thing for pyOpenSSL is to:

  * accept `bytes` or `unicode` for any path value
  * automatically encode `unicode` to `bytes` using `sys.getfilesystemencoding()`

It would be nice to have a function which performed the necessary check and encoding for a single value.  This could be used throughout pyOpenSSL anywhere an API accepts a path.

tisdall:
I've been using something like this:
```
file_name_encoding = sys.getfilesystemencoding()
if file_name_encoding is None:
    file_name_encoding = sys.getdefaultencoding()
```

As per the docs saying "Return the name of the encoding used to convert Unicode filenames into system file names, or None if the system default encoding is used." (https://docs.python.org/2/library/sys.html#sys.getfilesystemencoding)

tisdall:
how is this for a utility function:
```python
from six import PY3, binary_type, text_type

def filename_to_bytes(filename):
    if isinstance(filename, binary_type):
        return filename
    elif isinstance(filename, text_type):
        fs_encoding = sys.getfilesystemencoding()
        if fs_encoding is None:
            fs_encoding = sys.getdefaultencoding()
        return filename.encode(fs_encoding)
    else:
        raise TypeError("filenames must either be a string or bytes")
```

exarkun:
I'm not sure it makes sense to apply `sys.getdefaultencoding()` to filesystem paths.  If there's no indication of how path names are represented in the underlying system then it's probably better not to guess or to make a very simple, safe guess (like "ascii").

[...]

Issue #15, opened 2014-01-23
Fix regressions in support for bytes / unicode in certain APIs


adiroiban:
In previous pyopenssl versions the API was fine with accepting strings as Unicode.

For example this was a valid code, which no longer works in latest version. Context.set_cipher_list(u'ALL')

The migration should not be a big deal, but I think that this needs to be documented.

---

Also, handling unicode paths in a cross-platform code is a bit messy since Python API on Linux sometimes accepts Unicode, sometimes byte string (depending on environment), and on Windows you always have to pass the path a Unicode.

Maybe OpenSSL.test.util.TestCase.mktemp should be changed to always return a unicode path, example via mktemp(dir='.', suffix=UNICODE_TEST_MARKER)

Thanks!

abrahammartin:
Is 0.15 going to be release any time soon? 0.14 was released a year ago and there are a lot of bugfixes and new features added since then.

dbrent-amazon:
I am having the same issue. http://stackoverflow.com/questions/28749543/pyopenssl-signature-verification-using-python-3

exarkun:
> Is 0.15 going to be release any time soon?

It's on my todo list.  I had planned to do one over the holidays but got tied up with other (even less fun, I assure you) things.  I have some airport time coming up during which I intend to give this another shot.  Sorry about not getting to this more quickly.

sgallagher:
I was just about to file this as a new bug when I stumbled upon this ticket. I'm working on a script to generate self-signed certificates and it's broken on Python 3. One thing I couldn't discern from the comments above: is there a workaround? Can I convert the strings to something else that would be accepted?

The following code works in Python 2
```python
crypto.X509Extension("subjectKeyIdentifier",
                                 False,
                                 "hash",
                                 subject=ca_cert,
                                 issuer=ca_cert)
```

However, on Python 3, it results in the error:
```
File "/usr/lib/python3.4/site-packages/OpenSSL/crypto.py", line 493, in __init__
    extension = _lib.X509V3_EXT_nconf(_ffi.NULL, ctx, type_name, value)
TypeError: initializer for ctype 'char *' must be a bytes or list or tuple, not str
```

adiroiban:
you can try using `b"subjectKeyIdentifier"` to force bytes... similar for `b"hash"`

exarkun:
[...]

PR #182, opened 2015-01-05
Fix issue with string type check in python3

SamWhited:
The `rand.write_file` and `rand.read_file` methods take a string, however, they excplicitly check for type `bytes`. In Python 3, this check will fail when given a normal UTF-8 string (`isinstance("", bytes)` is false).

One would expect to be able to `rand.write_file("somefilename")`, this fixes that and works in both python 2 and 3.

I didn't see anywhere else this was happening, but it's quite possible that this should be fixed elsewhere as well (in which case using `6` or having a generalized `is_string` utility method or something would probably be a better solution). As it was only two places, I just fixed it in both (this simpler fix reads nicely anyways).

### DIFF
--- a/OpenSSL/rand.py
+++ b/OpenSSL/rand.py
@@ -136,7 +136,7 @@ def load_file(filename, maxbytes=_unspecified):
                      to read the entire file
     :return: The number of bytes read
     """
-    if not isinstance(filename, str) and not isinstance(filename, _builtin_bytes):
+    if not isinstance(filename, (str, _builtin_bytes)):
         raise TypeError("filename must be a string")
 
     if maxbytes is _unspecified:
@@ -155,7 +155,7 @@ def write_file(filename):
     :param filename: The file to write data to
     :return: The number of bytes written
     """
-    if not isinstance(filename, str) and not isinstance(filename, _builtin_bytes):
+    if not isinstance(filename, (str, _builtin_bytes)):
         raise TypeError("filename must be a string")
 
     return _lib.RAND_write_file(filename)
--- a/OpenSSL/rand.py
+++ b/OpenSSL/rand.py
@@ -136,7 +136,7 @@ def load_file(filename, maxbytes=_unspecified):
                      to read the entire file
     :return: The number of bytes read
     """
-    if not isinstance(filename, _builtin_bytes):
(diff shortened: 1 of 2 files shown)


exarkun:
Thanks.  Can you please add unit tests for this change as well?  Also note you can shorten this by using the feature of `isinstance(x, (type_a, type_b)`.

FWIW this occurs in many, many more places in pyOpenSSL (#15, #62) but the small scope of this PR is very good - it means the change is easier to review and merge.

SamWhited:
Used the shorter version of `isinstance` (thanks; good to know that will take a tuple).

I glanced at the tests, but couldn't get them to run right away. Will try again later when I get off the road. A more generall `is_string_or_bytes` utility method might be appropriate since this is happening elsewhere.

SamWhited:
How would you prefer this be tested? The tests which check if the arguments are correct probably shouldn't actually read/write files, but the test of reading/writing files doesn't really need to do so twice (once with the filename as a byte string, one as a normal string). Probably doesn't matter so much; I just like to keep things organized.

Thoughts?

exarkun:
> How would you prefer this be tested? The tests which check if the arguments are correct probably shouldn't actually read/write files

I think that ideally that would be the case but pyOpenSSL's APIs are not particularly well suited to testing argument validation from actually performing the work.

There should probably be another test like `test_files` that uses unicode instead of bytes and makes similar assertions.  You might want to refactor the existing test to use a helper that can be shared by the new test.

exarkun:
> A more generall is_string_or_bytes utility method might be appropriate since this is happening elsewhere.

I've been thinking about this a little bit more as I've been trying to finish #15 / #189 / #201.  The conclusion I've reached is that pyOpenSSL probably needs to be encoding unicode to bytes using `sys.getfilesystemencoding()` anywhere it wants to pass a filename into OpenSSL.  So the helper to introduce should probably be something like `maybe_convert_text_path` which accepts `bytes` and returns them unmodified, accepts `unicode` (aka Python 3 `str`) and returns them encoded properly, and rejects everything else.

If you feel like adding that helper (probably to `OpenSSL/_util.py`) in this PR, great.  If not, just implement that behavior for these `OpenSSL.rand` APIs and then someone else can refactor that implementation into a helper for re-use elsewhere in OpenSSL later.

PR #183, opened 2015-01-08
Fix error handling in Connection.shutdown

### LINKED ISSUE #91: SSL_shutdown error codes are un-discoverable
The `_raise_current_error` in `Connection.shutdown` is incorrect because `SSL_shutdown` can result in `SSL_ERR_WANT_WRITE` or `SSL_ERR_WANT_READ`.  The way you discover this, according to [the official documentation](https://www.openssl.org/docs/ssl/SSL_shutdown.html), is:

> If the underlying BIO is non-blocking, SSL_shutdown() will also return when the underlying BIO could not satisfy the needs of SSL_shutdown() to continue the handshake. In this case a call to SSL_get_error() with the return value of SSL_shutdown() will yield SSL_ERROR_WANT_READ or SSL_ERROR_WANT_WRITE.

The current behavior is that the error code is masked because the return code is lost.

darkrain42:
Fix the error handling in ``Connection.shutdown`` so that underlying socket conditions (``SSL_ERR_WANT_WRITE``, ``SSL_ERR_WANT_WRITE``, or other system-level errors such as ``ECONNRESET``) are propagated properly.

This fixes #91.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,8 @@
+2015-01-08  Paul Aurich <paul@darkrain42.org>
+
+	* OpenSSL/SSL.py: ``Connection.shutdown`` now propagates errors from the
+	  underlying socket.
+
 2014-08-21  Alex Gaynor  <alex.gaynor@gmail.com>
 
 	* OpenSSL/crypto.py: Fixed a regression where calling ``load_pkcs7_data``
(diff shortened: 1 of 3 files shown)


exarkun:
Thanks.  Do you want to add a ChangeLog entry for this fix as well?

darkrain42:
Done!

PR #184, opened 2015-01-11
Add support for querying the negotiated TLS version.

richmoore:
Note this requires https://github.com/pyca/cryptography/pull/1619 to be merged first.

### DIFF
--- a/OpenSSL/SSL.py
+++ b/OpenSSL/SSL.py
@@ -1551,6 +1551,16 @@ class Connection(object):
             return version.decode("utf-8")
 
 
+    def get_protocol_version(self):
+        """
+        Obtain the protocol version of the current connection.
+
+        :returns: The TLS version of the current connection, for example
+            the value for TLS 1.2 would be 0x303.
+        :rtype: :py:class:`int`
+        """
+        version = _lib.SSL_version(self._ssl)
+        return version
 
 ConnectionType = Connection
 
--- a/OpenSSL/test/test_ssl.py
+++ b/OpenSSL/test/test_ssl.py
@@ -2129,6 +2129,20 @@ class ConnectionTests(TestCase, _LoopbackMixin):
         self.assertEqual(server_cipher_bits, client_cipher_bits)
 
 
+    def test_get_protocol_version(self):
+        """
+        :py:obj:`Connection.get_protocol_version` returns a :py:class:`int`
+        giving the protocol version of the current connection.
+        """
+        server, client = self._loopback()
+        server_protocol_version, client_protocol_version = \
+            server.get_protocol_version(), client.get_protocol_version()
+
+        self.assertIsInstance(server_protocol_version, int)
+        self.assertIsInstance(client_protocol_version, int)
+
+        self.assertEqual(server_protocol_version, client_protocol_version)
+
 
 class ConnectionGetCipherListTests(TestCase):
     """
--- a/setup.py
+++ b/setup.py
@@ -34,7 +34,7 @@ setup(name='pyOpenSSL', version=__version__,
       maintainer_email = 'exarkun@twistedmatrix.com',
       url = 'https://github.com/pyca/pyopenssl',
       license = 'APL2',
-      install_requires=["cryptography>=0.5.4", "six>=1.5.2"],
+      install_requires=["cryptography>=0.7.2", "six>=1.5.2"],
       long_description = """\
 High-level wrapper around a subset of the OpenSSL library, includes
  * SSL.Connection objects, wrapping the methods of Python's portable


richmoore:
Failure is expected - the travis run was before the cryptography change was merged (it is now).

exarkun:
Thanks.  I think this change will also need to include a bump to the declared required cryptography version in setup.py (which is easier after cryptography releases the new feature that's required, but I think you can construct a `dev`-style version in the interim if you want to get this merged before the next cryptography release).

reaperhulk:
We'll be releasing 0.7.2 with the binding required for this feature in the next day or two.

richmoore:
I've updated setup.py in my branch.

OpenSSL/SSL.py
```diff
@@ -1551,6 +1551,16 @@ def get_cipher_version(self):
             return version.decode("utf-8")
 
 
+    def get_protocol_version(self):
+        """
+        Obtain the protocol version of the current connection.
+
+        :returns: The TLS version of the current connection, for example
+            the value for TLS 1.2 would be 0x303.
```
exarkun: It may be hard for most people to generalize from "0x303 means TLS 1.2" to an understanding of other values.  Or maybe that's just mean.  Anyway, can you expand this documentation or make it easier to understand these values some other way?  An example of a non-documentation solution might be to make the result self-documenting by defining a collection of symbolic constants and referring to that collection here.  Or another solution might be to refer to some existing OpenSSL documentation about this value.

Which brings me to another point, where is the documentation for `SSL_version`?  I can't find any.  I did find `SSL_get_version` which apparently returns a **string** instead.  Exposing that instead might be another way to solve the documentation issue - since "TLS1.2" doesn't need as much explanation as 0x303 (though just going by the OpenSSL documentation, I don't know if "TLSv1.2" is a value that will ever be returned by `SSL_get_version`!  It only documents "SSLv2", "SSLv3", and "TLSv1" - oh, and, awesomely, "unknown").

OpenSSL/test/test_ssl.py
```diff
@@ -2129,6 +2129,20 @@ def test_get_cipher_bits(self):
...
+        giving the protocol version of the current connection.
+        """
+        server, client = self._loopback()
+        server_protocol_version, client_protocol_version = \
+            server.get_protocol_version(), client.get_protocol_version()
+
+        self.assertIsInstance(server_protocol_version, int)
+        self.assertIsInstance(client_protocol_version, int)
+
+        self.assertEqual(server_protocol_version, client_protocol_version)
```
exarkun: This assertion would be satisfied if the implementation accidentally always returned a constant integer (`0` for example).

Perhaps this test could make a stronger assertion about the value?  It might even be worth having a couple tests for a couple different values.

exarkun:
Thanks.  Can you also add documentation to docs/api/ and an entry to ChangeLog?

exarkun:
Hi richmoore.  Any interest in following up on this or would it be better if someone else finished it up?  Thanks again for your work so far.

elitest:
I took a shot at finishing this one up.  When I switch to using `SSL_get_version`I get this
```
..............................................................................
======================================================================
ERROR: test_get_protocol_version (__main__.ConnectionTests)
----------------------------------------------------------------------
  File "OpenSSL/test/test_ssl.py", line 2755, in test_get_protocol_version
    server.get_protocol_version(), client.get_protocol_version()
  File "build/bdist.linux-x86_64/egg/OpenSSL/SSL.py", line 1894, in get_protocol_version
    version = _lib.SSL_get_version(self._ssl)
AttributeError: 'FFILibrary' object has no attribute 'SSL_get_version'

----------------------------------------------------------------------
Ran 193 tests in 6.109s

FAILED (errors=1)
```
When I try to print it from a simple ssl connect I get `None` as the value.  Is this a problem in this not being implemented in cffi?

alex:
`SSL_get_version` doesn't appear to be bound in cryptography, adding it
should be a pretty simple PR.

On Sun, Apr 26, 2015 at 12:22 PM, elitest <notifications@github.com> wrote:

> I took a shot at finishing this one up. When I switch to using
> SSL_get_versionI get this
>
> ..............................................................................
> ======================================================================
> ERROR: test_get_protocol_version (__main__.ConnectionTests)
> ----------------------------------------------------------------------
> Traceback (most recent call last):
>   File "OpenSSL/test/test_ssl.py", line 2755, in test_get_protocol_version
>     server.get_protocol_version(), client.get_protocol_version()
>   File "build/bdist.linux-x86_64/egg/OpenSSL/SSL.py", line 1894, in get_protocol_version
>     version = _lib.SSL_get_version(self._ssl)
> AttributeError: 'FFILibrary' object has no attribute 'SSL_get_version'
>
> ----------------------------------------------------------------------
> Ran 193 tests in 6.109s
>
> FAILED (errors=1)
>
> When I try to print it from a simple ssl connect I get None as the value.
> Is this a problem in this not being implemented in cffi?
>
> —
> <https://github.com/pyca/pyopenssl/pull/184#issuecomment-96404516>.
>



-- 

elitest:
We can probably close?  These changes were (effectively) merged in #244

PR #187, opened 2015-01-20
Remove EGD

Sp1l:
EGD was only necessary for some commercial UNIX systems, versions that needed it all reached end of life.
EGD needed until        OS release date
IRIX 6.5.19   feb 2003
Solaris 2.6                     jul 1997
AIX     5.2                     oct 2002
Tru64   5.1B                    sep 2002
HP-UX   11i v2                  sep 2003
https://en.wikipedia.org/wiki//dev/random#EGD_as_an_alternative

### DIFF
--- a/OpenSSL/rand.py
+++ b/OpenSSL/rand.py
@@ -93,29 +93,6 @@ def status():
 
 
 
-def egd(path, bytes=_unspecified):
-    """
-    Query an entropy gathering daemon (EGD) for random data and add it to the
-    PRNG. I haven't found any problems when the socket is missing, the function
-    just returns 0.
-
-    :param path: The path to the EGD socket
-    :param bytes: (optional) The number of bytes to read, default is 255
-    :returns: The number of bytes read (NB: a value of 0 isn't necessarily an
-              error, check rand.status())
-    """
-    if not isinstance(path, _builtin_bytes):
-        raise TypeError("path must be a byte string")
-
-    if bytes is _unspecified:
-        bytes = 255
-    elif not isinstance(bytes, int):
-        raise TypeError("bytes must be an integer")
-
-    return _lib.RAND_egd_bytes(path, bytes)
-
-
-
 def cleanup():
     """
     Erase the memory used by the PRNG.
--- a/OpenSSL/test/test_rand.py
+++ b/OpenSSL/test/test_rand.py
@@ -103,43 +103,6 @@ class RandTests(TestCase):
         self.assertTrue(rand.status() in (1, 2))
 
 
-    def test_egd_wrong_args(self):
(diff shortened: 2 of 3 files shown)


public:
I think you might need to delete more stuff to keep coverage up?

alex:
Just deleting this seems like a backwards compatibility issue; I think it needs to be deprecated (even though EGD is toatlly stupid)

Sp1l:
@alex If anyone still wants to run pyopenssl on the platforms mentioned in the pull request they have an immense number of other compatibility issues. I expect they'll be needing to run all kinds of ancient versions of Python/OpenSSL/pyOpenSSL/py-cryptography in line with their museum-worthy hardware and OS. Even Windows has a proper source of random for many years now.

exarkun:
Thanks, Sp1l.  I had no idea EGD was this obsolete.

Nevertheless, this change could break working programs using pyOpenSSL.  Consider a program that just does:

    from OpenSSL.rand import egd

and never proceeds to use `egd`.  Clearly this is easily fixed by the maintainer of the program (just delete the import).  However, if we release a version of pyOpenSSL that simply removes egd then end-users may install it and use it with programs the maintainers of which haven't even had a chance to learn that this API has been removed.  An easy fix for a maintainer can be a total mystery to an end-user.  I'd prefer not to introduce this sort of failure for users of pyOpenSSL-using software.

The thing to do here is deprecate the APIs being removed (by emitting warnings, probably), release a version of pyOpenSSL that includes those deprecations, wait a while, then remove the deprecated APIs.

Sp1l:
@exarkun If you give me an example of how you mark features as deprecated I'll get you an updated patch marking deprecation.
Be warned that Python 2.7/3.x from the next patch version onwards will conditionally use EGD and thus will come without EGD support completely. Any code importing EGD will break anyway as a result of that!

exarkun:
> Be warned that Python 2.7/3.x from the next patch version onwards will conditionally use EGD and thus will come without EGD support completely. Any code importing EGD will break anyway as a result of that!

I'm not sure I follow this.  pyOpenSSL's EGD is totally distinct from any Python standard library EGD support.  I don't think changes to one will make any difference to users of the other.

> @exarkun If you give me an example of how you mark features as deprecated I'll get you an updated patch marking deprecation.

This deprecation from Twisted demonstrates one approach - http://twistedmatrix.com/trac/browser/trunk/twisted/web/server.py#L69

Unfortunately it depends on a deprecation API provided by Twisted (as the standard library doesn't do much to help with the deprecation process).  I wouldn't mind if the helper was copied from Twisted into pyOpenSSL's util module (with proper attribution, etc).

Thanks again.

PR #189, opened 2015-02-04
#15 Fix regressions in support for bytes / unicode in certain APIs

abrahammartin:
#15 Fix regressions in support for bytes / unicode in certain APIs 
Added a check for _text_type, raise a DeprecationWarning and encode the str to utf-8 or asciii. This should preserve 0.13 API calls.
I used the 0.13 test suite with the 0.14 code and made the necessary changes to make 0.13 and 0.14 tests pass with python 2.X and 3.X using travis.

### DIFF
--- a/OpenSSL/SSL.py
+++ b/OpenSSL/SSL.py
@@ -1,4 +1,3 @@
-import six
 from warnings import warn
 from sys import platform
 from functools import wraps, partial
@@ -8,6 +7,7 @@ from errno import errorcode
 
 from six import text_type as _text_type
 from six import integer_types as integer_types
+from six import PY2 as _PY2
 
 from OpenSSL._util import (
     ffi as _ffi,
@@ -315,7 +315,7 @@ class Context(object):
 
         # Backward compatibility
         if isinstance(cafile, _text_type):
-            if six.PY2:
+            if _PY2:
                 warn("unicode in cafile is no longer accepted, use bytes", DeprecationWarning)
             else:
                 warn("str in cafile is no longer accepted, use bytes", DeprecationWarning)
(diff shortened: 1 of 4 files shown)


exarkun:
Thaks for your work on this.  The reason these regressions were introduced in the first place was the lack of proper testing.  Can you also update the test suite to exercise these cases?

OpenSSL/SSL.py
```diff
@@ -310,6 +310,12 @@ def load_verify_locations(self, cafile, capath=None):
         :param capath: In which directory we can find the certificates
         :return: None
         """
+
+        # Backward compatibility
+        if isinstance(cafile, _text_type):
+            DeprecationWarning("str object in cafile is no longer accepted, use bytes")
```
exarkun: This doesn't actually emit a warning.  It just creates the warning object and then discards it.  Warnings are typically emitted using `warnings.warn`.

Also, this message might be confusing on Python 2 where `str` and `bytes` are synonyms.  I suggest saying "text" instead of "str object" or, perhaps better, emit different messages on Python 2 and Python 3: say "unicode" on Python 2 and "str" on Python 3.

OpenSSL/SSL.py
```diff
@@ -310,6 +311,17 @@ def load_verify_locations(self, cafile, capath=None):
...
         """
+
+        # Backward compatibility
+        if isinstance(cafile, _text_type):
+            if version_info.major == 2:
+                warn("unicode in cafile is no longer accepted, use bytes", DeprecationWarning)
+            elif version_info.major == 3:
+                warn("str in cafile is no longer accepted, use bytes", DeprecationWarning)
+            else:
+                warn("text in cafile is no longer accepted, use bytes", DeprecationWarning)
```
exarkun: You can assume that the only versions of Python this code needs to handle are 2 and 3.  Also you might find `six.PY3` a more convenient spelling of the version check.

abrahammartin:
I can't see tests for bio_write in the ConnectionBIOTests class, only for bio_read. Are they located somewhere else?

exarkun:
Looks like there aren't really any direct unit tests (which is unfortunate).  `bio_write` (and `bio_read`) get exercised a lot by the `_interactInMemory` helper, though, which is used by a bunch of tests for other things.

abrahammartin:
Tests added. If you need something else, let me know.

PR #193, opened 2015-02-13
Compatibility for OpenSSL >= 1.0.1i

### LINKED ISSUE #192: X509.has_expired() returns wrong value on OS X / OpenSSL 1.0.2
I'm putting this here to track down some weird behavior I'm seeing when I was testing the verify-chain branch.

pyOpenSSL: 0.14
cryptography: 0.7.2

Test fails on: OpenSSL: 1.0.2 22 Jan 2015 (Homebrew OpenSSL)
Test passes on: OpenSSL: 0.9.8zc 15 Oct 2014 (OS X 10.10.2 OpenSSL)

```python
from OpenSSL.crypto import load_certificate
from OpenSSL.crypto import FILETYPE_PEM
from OpenSSL.test.test_crypto import root_cert_pem

cert = load_certificate(FILETYPE_PEM, root_cert_pem)
assert cert.has_expired() == False
```

mrjefftang:
ASN1_UTCTIME_cmp_time_t is returning -2 on OpenSSL 1.0.2

Addresses pyca/pyopenssl#192

### DIFF
--- a/OpenSSL/crypto.py
+++ b/OpenSSL/crypto.py
@@ -1,5 +1,6 @@
-from time import time
+from time import time, strptime
 from base64 import b16encode
+from calendar import timegm
 from functools import partial
 from operator import __eq__, __ne__, __lt__, __le__, __gt__, __ge__
 from warnings import warn as _warn
@@ -1161,10 +1162,10 @@ class X509(object):
         :return: True if the certificate has expired, false otherwise
         """
         now = int(time())
-        notAfter = _lib.X509_get_notAfter(self._x509)
-        return _lib.ASN1_UTCTIME_cmp_time_t(
-            _ffi.cast('ASN1_UTCTIME*', notAfter), now) < 0
+        notAfter = self.get_notAfter().decode('utf-8')
+        notAfterSecs = timegm(strptime(notAfter, '%Y%m%d%H%M%SZ'))
 
+        return now > notAfterSecs
 
     def _get_boundary_time(self, which):
         return _get_asn1_time(which(self._x509))
--- a/OpenSSL/crypto.py
+++ b/OpenSSL/crypto.py
@@ -1,4 +1,4 @@
-from time import time 
+from time import time
 from base64 import b16encode
 from functools import partial
 from operator import __eq__, __ne__, __lt__, __le__, __gt__, __ge__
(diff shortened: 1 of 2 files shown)


mrjefftang:
Looks like TravisCI isn't accepting new requests for OS X test builds.  I'm not sure how to get this tested as there's no Ubuntu OpenSSL 1.0.2 package either.

mrjefftang:
Final error arises from ```test_set_tmp_ecdh``` when it calls ```context.set_tmp_ecdh(curve)``` for EC curves ```Oakley-EC2N-4``` and ```Oakley-EC2N-3```.

mrjefftang:
@exarkun It's not clear to me why the ```Context.set_tmp_ecdh()``` call for ```Oakley-EC2N-4``` and ```Oakley-```EC2N-3``` cause an OpenSSL exception but it looks like it's also not being raised properly either since it's the ```TestCase.tearDown()``` cleanup that is discovering the exception.

reaperhulk:
@mrjefftang could we get you to rebase this now? Thanks for your work so far!

mrjefftang:
@reaperhulk Done.

OpenSSL/crypto.py
```diff
@@ -1161,10 +1162,10 @@ def has_expired(self):
         :return: True if the certificate has expired, false otherwise
         """
         now = int(time())
-        notAfter = _lib.X509_get_notAfter(self._x509)
-        return _lib.ASN1_UTCTIME_cmp_time_t(
-            _ffi.cast('ASN1_UTCTIME*', notAfter), now) < 0
+        notAfter = self.get_notAfter().decode('utf-8')
```
hynek: After a discussion with Paul I think we should not do that.  pyOpenSSL should offer as little abstraction as necessary and since this bug has been fixed in 1.0.2a, we shouldn’t add baggage to our codebase.

hynek:
Thanks for this and your responsiveness!  Please revert the one change and we’re good to go if the CI stays green. :)

mrjefftang:
@hynek Done

reaperhulk:
The fix that will stop the notAfter bug has not yet shipped (should be in 1.0.2b actually), but we'll merge this now since it shows the issue. Thanks again!

hynek:
Thanks to all involved and sorry for the delays!

PR #195, opened 2015-02-19
Add Elliptic Curve support to PKey

mrjefftang:
This is a rewrite of an older PR that adds EC support to PKey to incorporate the inclusion of _EllipticCurve objects.

### DIFF
--- a/OpenSSL/crypto.py
+++ b/OpenSSL/crypto.py
@@ -27,6 +27,10 @@ FILETYPE_TEXT = 2 ** 16 - 1
 
 TYPE_RSA = _lib.EVP_PKEY_RSA
 TYPE_DSA = _lib.EVP_PKEY_DSA
+if _lib.Cryptography_HAS_EC:
+    TYPE_EC = _lib.EVP_PKEY_EC
+else:
+    TYPE_EC = None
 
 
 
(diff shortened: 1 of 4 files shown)


OpenSSL/crypto.py
```diff
@@ -172,15 +176,18 @@ def generate_key(self, type, bits):
         """
         Generate a key of a given type, with a given number of a bits
 
-        :param type: The key type (TYPE_RSA or TYPE_DSA)
-        :param bits: The number of bits
+        :param type: The key type (TYPE_RSA or TYPE_DSA or TYPE_EC)
+        :param bits: The number of bits or EllipticCurve
```
elitest: If I was specifying a curve, I'm not sure if I would know how I would do it, based on this. If I was using RSA I would specify something like "2048".  If I was specifying a curve, I would  specify what? "secp384r1"?  I'm not sure if I have a better alternative though or a suggestion for where you should take it.

mrjefftang: I agree, it's not the greatest approach.

The current implementation is used as follows:
```
curve = OpenSSL.crypto.get_elliptic_curve('secp384r1')
pkey = PKey()
pkey.generate(key, TYPE_EC, curve)
```

Honestly I think the best approach would be to subclass `PKey` into  `RSAPKey`, `DSAPKey`, and `ECPKey` but that is getting away from the thin abstraction on top of OpenSSL.

hynek:
I believe this has been fixed via #439.  Sorry for the inappropriate process!

Issue #190, opened 2015-02-12
Hello Request breaks pyOpenSSL


mhils:
Hi there,

I tried to hunt down https://github.com/mitmproxy/mitmproxy/issues/472 today and encountered a strange pyOpenSSL bug. If the server

1. requests renegotiation (`Hello Request`),
2. (unsuccessfully) asks for a client certificate during renegotation and then
3. sends some encrypted data, I cannot read that data using pyOpenSSL.

Here's my testcase - I think I'm doing nothing special:
```python
from OpenSSL import SSL
import socket

ctx = SSL.Context(SSL.SSLv23_METHOD)
sock = SSL.Connection(ctx, socket.socket(socket.AF_INET, socket.SOCK_STREAM))
sock.connect(("exchangedev.taskbox.co", 443))
sock.do_handshake()

def dump_creds():
    crand = sock.client_random().encode("hex")
    masterkey = sock.master_key().encode("hex")
    with open("keys","ab") as f:
        f.write("CLIENT_RANDOM {} {}\r\n".format(crand, masterkey))

dump_creds()

sock.send("GET /Microsoft-Server-ActiveSync HTTP/1.1\r\n"
		  "Host: exchangedev.taskbox.co\r\n"
		  "\r\n"
		  "\r\n")

try:
    print(sock.recv(1))
finally:
    dump_creds() # renegotiated master key
```
Here's the output:
```
C:\Users\user\git\mitmproxy\test\>python test.py
Traceback (most recent call last):
  File "test.py", line 23, in <module>
    print(sock.recv(1))
  File "C:\Python27\lib\site-packages\OpenSSL\SSL.py", line 995, in recv
    self._raise_ssl_error(self._ssl, result)
  File "C:\Python27\lib\site-packages\OpenSSL\SSL.py", line 847, in _raise_ssl_error
    raise WantReadError()
OpenSSL.SSL.WantReadError
```
Now, that clearly looks like I'm not getting any data back. However, looking at Wireshark, I see this:

![Wireshark Output](https://maximilianhils.com/upload/2015-02/2015-02-12_03-00-31.png)

Similarly, using `openssl s_client -connect exchangedev.taskbox.co:443`, I get the response.

[...]

PR #196, opened 2015-02-28
Add SSL_MODE_* constants

mhils:
Make the `SSL_MODE_*` constants first class citizens. (refs #190)


exarkun:
Thanks.  `PARTIAL_WRITE` is required to make the existing test suite pass - so the part of this change covering that constant is tested (because `OpenSSL.test.test_ssl.MemoryBIOTests.test_outgoingOverflow` fails if you get that value wrong).

Do you think you could add some additional tests for the other constants here, demonstrating that the desired behavior can be achieved by using them?

Thanks.  Please re-open the PR when that's done.

PR #199, opened 2015-03-15
Implement Connection.recv_into

exarkun:


OpenSSL/SSL.py
```diff
@@ -1030,6 +1030,45 @@ def recv(self, bufsiz, flags=None):
...
+        """
+        if nbytes is None:
+            nbytes = len(buffer)
+        else:
+            nbytes = min(nbytes, len(buffer))
+
+        # We need to create a temporary buffer. This is annoying, it would be
+        # better if we could pass memoryviews straight into the SSL_read call,
+        # but right now we can't. Revisit this if CFFI gets that ability.
+        buf = _ffi.new("char[]", nbytes)
```
alex: FWIW with cffi 0.9 the the `ffi.from_buffer` method would let you skip this allocation

exarkun: Looks like only in some cases.  Not if the given buffer is a `bytearray`, for example?

alex: Yes, I believe that's correct.

On Sun, Mar 15, 2015 at 6:05 PM, Jean-Paul Calderone <
notifications@github.com> wrote:

> In OpenSSL/SSL.py
> <https://github.com/pyca/pyopenssl/pull/199#discussion_r26453014>:
>
> > +            buffer. If not present, defaults to the size of the buffer. If
> > +            larger than the size of the buffer, is reduced to the size of the
> > +            buffer.
> > +        :param flags: (optional) Included for compatibility with the socket
> > +            API, the value is ignored.
> > +        :return: The number of bytes read into the buffer.
> > +        """
> > +        if nbytes is None:
> > +            nbytes = len(buffer)
> > +        else:
> > +            nbytes = min(nbytes, len(buffer))
> > +
> > +        # We need to create a temporary buffer. This is annoying, it would be
> > +        # better if we could pass memoryviews straight into the SSL_read call,
> > +        # but right now we can't. Revisit this if CFFI gets that ability.
> > +        buf = _ffi.new("char[]", nbytes)
>
> Looks like only in some cases. Not if the given buffer is a bytearray,
> for example?
>
> —
> <https://github.com/pyca/pyopenssl/pull/199/files#r26453014>.
>



-- 

PR #209, opened 2015-04-13
Support bytes and unicode for filenames

exarkun:
Update of #182 to factor the common logic into a helper and apply it to the filename-accepting APIs in `OpenSSL.SSL` as well.


alex:
patch looks ok, assuming tests pass

exarkun:
Thanks!

OpenSSL/_util.py
```diff
@@ -68,6 +70,23 @@ def native(s):
...
+    path and which can be passed into an OpenSSL API accepting a filename.
+
+    :param s: An instance of :py:class:`bytes` or :py:class:`unicode`.
+
+    :return: An instance of :py:class:`bytes`.
+    """
+    if isinstance(s, binary_type):
+        return s
+    elif isinstance(s, text_type):
+        return s.encode(sys.getfilesystemencoding())
```
tisdall: Are there any issue here with `sys.getfilesystemencoding()` returning `None` in Py2?  The docs say that it can do that.  That's why I suggested `return filename.encode(sys.getfilesystemencoding() or 'ascii')` in #62.

However, it seems like `sys.getfilesystemencoding()` returns `None` if it's the "system default" and `string.encode()` can be passed nothing to use the default.  So maybe it should test for `None` and if it's `None` just call `.encode()` with no arguments.

tisdall: Looking over #62 again, I think `return filename.encode(sys.getfilesystemencoding() or 'ascii')` is the best option until we can divinate `open` more.  Just using `.encode()` with no argument is probably equivalent to `.encode(sys.getdefaultencoding())` which you didn't think was the right way (I'm not really sure).

With out the `or 'ascii'` Py2 users may get a `TypeError: encode() argument 1 must be string, not None` which will be rather cryptic.

PR #210, opened 2015-04-13
travis docker support + osx support ~~madness~~

reaperhulk:
I'll iterate on this until it works.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,40 +6,40 @@ matrix:
   - language: generic
     os: osx
     env: TOXENV=py27
-  - python: 2.6 # these are just to make travis's UI a bit prettier
+  - python: "2.6" # these are just to make travis's UI a bit prettier
     env: TOXENV=py26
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -1,83 +1,126 @@
...
+  - python: 3.4
+    env:
+      CRYPTOGRAPHY_GIT_MASTER=true
+  - python: pypy
+    env:
+      CRYPTOGRAPHY_GIT_MASTER=true
+  - python: 2.7
+    env:
       OPENSSL=0.9.8
+    addons:
```
exarkun: Can you add a comment here explaining this section needs to match the above equivalent section in the matrix?

exarkun:
Thanks!  This is a really nice upgrade to the travis config.  Please just address the one minor inline comment and also change the implicit strings back to explicit strings and then I'm happy (and we'll see what happens when it runs on travis).

PR #215, opened 2015-04-13
submit coverage whether or not travis actually succeeds.

reaperhulk:


alex:
Is this better than `after_script`? (No opinions, just asking)

reaperhulk:
I have no opinion either.

alex:
I lied, I very very slightly think "after_script" is more appropriate

reaperhulk:
Well since we're still waiting on CI let's just change it now :)

alex:
A+

PR #216, opened 2015-04-13
Remove TOXENV env vars

alex:
We don't actually use tox in these tests


hynek:
That’s true but I’d rather switch Travis to tox such that local tests are run the same way as on Travis.

reaperhulk:
#230 is going to start using tox so this is obsolete now.

PR #225, opened 2015-04-14
Don't randomly fail with KeyError in shutdown().

### LINKED ISSUE #224: 0.15 raises a KeyError on calling shutdown twice
Previously this raised an `OpenSSL.SSL.Error` with the empty list.

glyph:
Fixes #224


hynek:
Bots are happy; please add a changelog entry and we’ll release 0.15.1 ASAP.

PR #230, opened 2015-04-14
Use tox in CI

### LINKED ISSUE #226: Migrate to tox for travis
This will make it simpler to replicate the behavior of the CI locally.

hynek:
tox should be used for running tests such that local and Travis tests are run the same way.

fixes #226

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ dist
 __pycache__
 .tox
 doc/_build/
+.coverage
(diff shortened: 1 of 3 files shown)


hynek:
@reaperhulk is there a way to mark all `CRYPTOGRAPHY_GIT_MASTER=true` as allow_failures without enumerating them all? :-/

PR #233, opened 2015-04-15
Use py.test test runner

### LINKED ISSUE #227: Switch tests to py.test
@hynek loves pytest.

hynek:
Fixes #227

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -3,15 +3,13 @@ envlist = pypy,py26,py27,py32,py33,py34
 
 [testenv]
 deps =
-    setuptools>=7.0
+    setuptools>=7.0  # older setuptools pollute CWD with egg files of dependencies
     coverage
-    pytest
 setenv =
     # Do not allowed the executing environment to pollute the test environment
     # with extra packages.
     PYTHONPATH=
-# The standard library unittest module can run tests on Python 2.7 and newer
 commands =
     python -c "import OpenSSL.SSL; print(OpenSSL.SSL.SSLeay_version(OpenSSL.SSL.SSLEAY_VERSION))"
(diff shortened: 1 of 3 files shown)


tox.ini
```diff
@@ -12,5 +13,5 @@ setenv =
 # The standard library unittest module can run tests on Python 2.7 and newer
 commands =
     python -c "import OpenSSL.SSL; print(OpenSSL.SSL.SSLeay_version(OpenSSL.SSL.SSLEAY_VERSION))"
-    coverage run --branch --source=OpenSSL setup.py test
+    coverage run --branch --source=OpenSSL {envdir}/bin/py.test OpenSSL
```
alex: You can use `-m pytest` instead of the `{envdir}` thing

tox.ini
```diff
@@ -12,5 +13,5 @@ setenv =
 # The standard library unittest module can run tests on Python 2.7 and newer
```
exarkun: This comment is being out-dated, it appears.

tox.ini
```diff
@@ -12,5 +13,5 @@ setenv =
...
 commands =
     python -c "import OpenSSL.SSL; print(OpenSSL.SSL.SSLeay_version(OpenSSL.SSL.SSLEAY_VERSION))"
-    coverage run --branch --source=OpenSSL setup.py test
+    coverage run --branch --source=OpenSSL -m pytest OpenSSL
```
exarkun: Do you want to leave the test suite information in `setup.py` even though it's no longer being used by the CI system?

hynek: I will make `setup.py test` work again once I’ve verified that the test suite actually works.

PR #234, opened 2015-04-15
set string mask utf8only

reaperhulk:
replaces #115


alex:
If this has been the default for forever, why do we need to set the global state?

Spindel:
It was only default in the openssl config file. Utf8 was made default in august last year inside the openssl library. But not on all branches.

hynek:
@alex it has been explained to me that the current default simply garbles strings.

@reaperhulk I feel like we should add a doc about how to re-instantiate the default behavior?

reaperhulk:
We could I suppose, but where should that live?

PR #235, opened 2015-04-16
Do not litter working dir

hynek:
This PR cleans up the handling of temporary files such that the CWD is not filled with them.

It changes the approach to creating a directory which it kills afterwards instead of keeping track of single files.

Some adjacent lines have been also cleaned up.

No code except tests is touched by this PR.

### DIFF
--- a/OpenSSL/test/test_rand.py
+++ b/OpenSSL/test/test_rand.py
@@ -205,8 +205,7 @@ class RandTests(TestCase):
         Random data can be saved and loaded to files with paths specified as
         bytes.
         """
-        path = self.mktemp()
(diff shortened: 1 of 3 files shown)


hynek:
One thing I would suggest is to add the snowman unconditionally to all tmp files to simplify the code some more.

hynek:
I’ve simplified a bunch, r4r!

reaperhulk:
LGTM. The way `SSL_CTX_load_verify_locations` works is terrible. I haven't dug deeply yet but I bet there's a way to do that without requiring files on disk.

reaperhulk:
(But that is an issue for a future PR)

PR #243, opened 2015-04-25
Updating certificate generation in examples

elitest:
Hi all,

I was looking through the examples and noticed that they had not been looked at in a while, I thought that perhaps we should update them to be more in line with what is considered contemporary good practice.  In this pull request:
* md5->sha256 for signature
* 1024->2048 for key length
* added print statements to inform the user what keys were going where

### DIFF
--- a/examples/mk_simple_certs.py
+++ b/examples/mk_simple_certs.py
@@ -7,15 +7,15 @@ from certgen import *   # yes yes, I know, I'm lazy
 cakey = createKeyPair(TYPE_RSA, 2048)
 careq = createCertRequest(cakey, CN='Certificate Authority')
 cacert = createCertificate(careq, (careq, cakey), 0, (0, 60*60*24*365*5)) # five years
-print('Creating Certificate Authority private key in \"simple/CA.pkey\"')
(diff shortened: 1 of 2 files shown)


examples/mk_simple_certs.py
```diff
@@ -4,14 +4,18 @@
 
 from OpenSSL import crypto
 from certgen import *   # yes yes, I know, I'm lazy
-cakey = createKeyPair(TYPE_RSA, 1024)
+cakey = createKeyPair(TYPE_RSA, 2048)
 careq = createCertRequest(cakey, CN='Certificate Authority')
 cacert = createCertificate(careq, (careq, cakey), 0, (0, 60*60*24*365*5)) # five years
+print('Creating Certificate Authority private key in \"simple/CA.pkey\"')
 open('simple/CA.pkey', 'w').write(crypto.dump_privatekey(crypto.FILETYPE_PEM, cakey))
+print('Creating Certificate Authority certificate in \"simple/CA.cert\"')
```
alex: The `\` are not needed here, since you're using single quotes.

alex:
Besides that comment, this looks great. Thanks for doing this!

PR #244, opened 2015-04-27
Add support for querying the negotiated TLS version.  The Quickening

elitest:
This PR hopefully finished up the work done by @richmoore in #184.  Thanks to @alex for helping me figure out how to get SSL_get_version implemented.  This PR requires the merge that happened in pyca/cryptography#1869, so do we wait until the next release of cryptography and then bump the required version number in setup.py?

I'm happy to hear feedback on whether the change to strings requires better/different testing?

Also the [OpenSSL docs](https://www.openssl.org/docs/ssl/SSL_get_version.html) have been updated since the original PR in January, and this does actually support TLS1.1 and TLS1.2 :)

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -7,6 +7,12 @@
 
 2015-04-15  Paul Kehrer <paul.l.kehrer@gmail.com>
 
+	* OpenSSL/SSL.py, : Add ``get_protocol_version()`` to Connection
+	  Based on work from Rich Moore
+	* OpenSSL/test/test_crypto.py: tests for ``get_protocol_version``
+
+2015-04-15  Paul Kehrer <paul.l.kehrer@gmail.com>
+
 	* OpenSSL/crypto.py, OpenSSL/test/test_crypto.py: Switch to utf8string
 	  mask by default. OpenSSL formerly defaulted to a T61String if there
 	  were UTF-8 characters present.  This was changed to default to
--- a/ChangeLog
+++ b/ChangeLog
@@ -7,9 +7,11 @@
 
 2015-04-15  Paul Kehrer <paul.l.kehrer@gmail.com>
 
-	* OpenSSL/SSL.py, : Add ``get_protocol_version()`` to Connection
+	* OpenSSL/SSL.py, : Add ``get_protocol_version()`` and 
+	  ``get_protocol_version_name()`` to Connection
 	  Based on work from Rich Moore
-	* OpenSSL/test/test_crypto.py: tests for ``get_protocol_version``
+	* OpenSSL/test/test_crypto.py: tests for ``get_protocol_version()``
+	  and ``get_protocol_version_name()``
 
 2015-04-15  Paul Kehrer <paul.l.kehrer@gmail.com>
 
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,8 +1,8 @@
 2015-05-27  Jim Shaver <dcypherd@gmail.com>
 
 	* OpenSSL/SSL.py, : Add ``get_protocol_version()`` and 
-	  ``get_protocol_version_name()`` to Connection
-	  Based on work from Rich Moore
+	  ``get_protocol_version_name()`` to ``Connection``.
+	  Based on work from Rich Moore.
 
 2015-05-02  Jim Shaver <dcypherd@gmail.com>
 
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,17 +1,17 @@
-2015-05-02  Jim Shaver <dcypherd@gmail.com>
-
-	* .travis.yml, setup.py, tox.ini: Removed support for Python 3.2.
-	  This version is rarely used and is now deprecated by a major
-	  dependency of pyOpenSSL (cryptography).  Affected users should upgrade
-	  to Python 3.3+.
-
-2015-04-15  Paul Kehrer <paul.l.kehrer@gmail.com>
+2015-05-27  Jim Shaver <dcypherd@gmail.com>
 
 	* OpenSSL/SSL.py, : Add ``get_protocol_version()`` and 
 	  ``get_protocol_version_name()`` to Connection
 	  Based on work from Rich Moore
 	* OpenSSL/test/test_crypto.py: tests for ``get_protocol_version()``
 	  and ``get_protocol_version_name()``
+
+2015-05-02  Jim Shaver <dcypherd@gmail.com>
+
+	* .travis.yml, setup.py, tox.ini: Removed support for Python 3.2.
+	  This version is rarely used and is now deprecated by a major
+	  dependency of pyOpenSSL (cryptography).  Affected users should upgrade
+	  to Python 3.3+.
 
 2015-04-15  Paul Kehrer <paul.l.kehrer@gmail.com>
 
--- a/ChangeLog
+++ b/ChangeLog
@@ -3,8 +3,6 @@
 	* OpenSSL/SSL.py, : Add ``get_protocol_version()`` and 
 	  ``get_protocol_version_name()`` to Connection
 	  Based on work from Rich Moore
-	* OpenSSL/test/test_crypto.py: tests for ``get_protocol_version()``
-	  and ``get_protocol_version_name()``
 
 2015-05-02  Jim Shaver <dcypherd@gmail.com>
 
--- a/OpenSSL/SSL.py
+++ b/OpenSSL/SSL.py
@@ -1888,11 +1888,12 @@ class Connection(object):
         Obtain the protocol version of the current connection.
 
         :returns: The TLS version of the current connection, for example
-            the value for TLS 1.2 would be ``b'TLSv1.2'``.
-        :rtype: :py:class:`bytes`
+            the value for TLS 1.2 would be ``TLSv1.2``or ``Unknown``
+            for connections that were not successfully.
+        :rtype: :py:class:`unicode`
         """
         version = _ffi.string(_lib.SSL_get_version(self._ssl))
-        return version
+        return version.decode("utf-8")
 
 
     def get_protocol_version(self):
--- a/OpenSSL/SSL.py
+++ b/OpenSSL/SSL.py
@@ -1883,16 +1883,16 @@ class Connection(object):
             return version.decode("utf-8")
 
 
-    def get_protocol_version(self):
+    def get_protocol_version_name(self):
         """
         Obtain the protocol version of the current connection.
 
         :returns: The TLS version of the current connection, for example
-            the value for TLS 1.2 would be ``TLSv1.2``.
+            the value for TLS 1.2 would be ``b'TLSv1.2'``.
         :rtype: :py:class:`unicode`
         """
-        version = _ffi.string(_lib.SSL_get_version(self._ssl))
-        return version.decode("utf-8")
+        version = _lib.SSL_get_version(self._ssl)
+        return version
 
 
     @_requires_npn
--- a/OpenSSL/SSL.py
+++ b/OpenSSL/SSL.py
@@ -1938,6 +1938,17 @@ class Connection(object):
         return _ffi.buffer(data[0], data_len[0])[:]
 
 
+    def get_protocol_version(self):
+        """
+        Obtain the protocol version of the current connection.
+
+        :returns: The TLS version of the current connection, for example
+            the value for TLS 1.2 would be 0x303.
+        :rtype: :py:class:`int`
+        """
+        version = _lib.SSL_version(self._ssl)
+        return version
+
 
 ConnectionType = Connection
 
--- a/OpenSSL/SSL.py
+++ b/OpenSSL/SSL.py
@@ -1891,7 +1891,7 @@ class Connection(object):
             the value for TLS 1.2 would be ``b'TLSv1.2'``.
         :rtype: :py:class:`unicode`
         """
-        version = _lib.SSL_get_version(self._ssl)
+        version = _ffi.string(_lib.SSL_get_version(self._ssl))
         return version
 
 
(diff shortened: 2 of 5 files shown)


doc/api/ssl.rst
```diff
@@ -598,6 +598,11 @@ Connection objects have the following methods:
     but not it returns the entire list in one go.
 
 
+.. py:method:: Connection.get_protocol_version()
+
+    Retrieve the version of the SSL or TLS protocol used by the Connection
```
alex: This should include an example of what the result looks like. Also should end with a `.` :-)

OpenSSL/test/test_ssl.py
```diff
@@ -2745,6 +2745,20 @@ def test_get_cipher_bits(self):
         self.assertEqual(server_cipher_bits, client_cipher_bits)
 
 
+    def test_get_protocol_version(self):
+        """
+        :py:obj:`Connection.get_protocol_version()` returns a string
+        giving the protocol version of the current connection.
+        """
+        server, client = self._loopback()
+        server_protocol_version, client_protocol_version = \
```
hynek: Please split this into two proper lines.  The current version takes the same amount of lines and is less readable.

OpenSSL/SSL.py
```diff
@@ -1883,6 +1883,18 @@ def get_cipher_version(self):
...
+    def get_protocol_version(self):
+        """
+        Obtain the protocol version of the current connection.
+
+        :returns: The TLS version of the current connection, for example
+            the value for TLS 1.2 would be 0x303.
+        :rtype: :py:class:`unicode`
+        """
+        version = _ffi.string(_lib.SSL_get_version(self._ssl))
+        return version.decode("utf-8")
```
hynek: I’m not sure about this one.  I believe we decided to keep pyOpenSSL’s APIs bytes-based except for paths.

So unless someone can enlighten me on what I’m missing, I’m gonna claim this ought to be bytes.

alex: Dunno, the name of the protocol version definitely seems textual to me.

On Mon, Apr 27, 2015 at 10:16 AM, Hynek Schlawack <notifications@github.com>
wrote:

> In OpenSSL/SSL.py
> <https://github.com/pyca/pyopenssl/pull/244#discussion_r29149506>:
>
> > @@ -1883,6 +1883,18 @@ def get_cipher_version(self):
> >              return version.decode("utf-8")
> >
> >
> > +    def get_protocol_version(self):
> > +        """
> > +        Obtain the protocol version of the current connection.
> > +
> > +        :returns: The TLS version of the current connection, for example
> > +            the value for TLS 1.2 would be 0x303.
> > +        :rtype: :py:class:`unicode`
> > +        """
> > +        version = _ffi.string(_lib.SSL_get_version(self._ssl))
> > +        return version.decode("utf-8")
>
> I’m not sure about this one. I believe we decided to keep pyOpenSSL’s APIs
> bytes-based except for paths.
>
> So unless someone can enlighten me on what I’m missing, I’m gonna claim
> this ought to be bytes.
>
> —
> <https://github.com/pyca/pyopenssl/pull/244/files#r29149506>.
>



-- 

elitest: If I understand your argument correctly, you are saying that when you query for TLS version you should receive something like 0x769 as a response?  The whole python3 bytes v. unicode thing is still new to me...  What was the thinking behind trying to stick with bytes?

Plenty of things in SSL.py are unicode strings though, including much of the related cipher information.  Should a developer really have to get out their OpenSSL decoder ring everytime she wants to query a cipher?  I think ``TLS_RSA_WITH_AES_256_CBC_SHA256`` makes a lot more sense than ``0x3d``, as long as the upstream doesn't change the formatting of those strings.

OpenSSL/test/test_ssl.py
elitest: Excellent point and done in 50d7d15.

OpenSSL/SSL.py
hynek: For the first paragraph; what I’m saying is that the docstring is plain wrong.  There’s no way you get anything resembling “0x303” back.  You get a string like `"TLSv1"` back.  Try it and pdb in your test. :)

elitest: My apologies, I should have looked at the code you were commenting on more closely!  I thought I had updated it from the way it was initially proposed in #184.  Fixed in 6ac2e2a.

elitest:
When does Travis update the version of cryptography it tests against?  Based on the way I read the .travis.yml, Travis is supposed to pull from cryptography's master branch?  If that is the case why would the test fail?

hynek:
Good question.  @reaperhulk any ideas?

reaperhulk:
When we switched to using tox in travis we didn't change it over to using the master branch properly, so it's installing it and then not using it.

elitest:
Is that something we want to fix?  If I understand [.travis.yml#L71-83](https://github.com/pyca/pyopenssl/blob/master/.travis.yml#L71-83) properly the problem is that we aren't installing cryptography's master into the virtualenv?  So when it installs pyOpenSSL into the virtualenv via setup.py it installs the latest released cryptography from PyPI?

reaperhulk:
Yeah that needs to be fixed. Right now it installs it into a venv, but then invokes tox (which builds its own venv and doesn't use the master branch).

hynek:
Please rebase on master; this issue should be fixed now.

hynek:
Also, until cryptography 0.9 is out, we can bikeshed on the return type.

My impression is, that pyOpenSSL is bytes only.  If you’d like to argue for unicode I would like to ask you for precedents.

pyOpenSSL is supposed to by a low-level layer around OpenSSL so I’m very apprehensive to introduce unicode-based APIs.  We’ve just added a ton of warning if certain methods are called with unicode.

elitest:
@hynek have you had a chance to look at the [original PR](https://github.com/pyca/pyopenssl/pull/184)?  Would you prefer that instead?

hynek:
Oi, that's where the incorrect doc string came from. :)

That's not what I meant (I just want to get rid of the decode call) but we should definitely expose both ways. I'm on my phone for the next hours but I'll look into it when I'm back home.

It'll be a different PR though. so don't worry about it here.

elitest:
So I might have some time to work on this during the coming weekend in.  Just to be clear, when you say we should expose both ways you mean exposing both ```SSL_get_version```(a.k.a. "TLV1.2" in bytes) and ```SSL_version```(a.k.a. 0x303)?  In #184 it was discussed how to make it clear what is being returned to you, in the documentation, do you have any suggestions there?

hynek:
So I’m saying we should expose both but we start with the string-based version.  And I’m also saying it should return bytes (not unicode) unless you can make a good case for unicode based on existing APIs. :)

hynek:
Can you rebase please?

elitest:
**WIP**
I'm still having issues wrapping my mind around the whole bytes thing... So tests currently fail.  Right now I'm getting stuff like:

>AssertionError: \<cdata 'char *' 0x7fbe700c6c96\> is not an instance of \<class 'bytes'\>

do I need to wrap the result that I'm returning in ```bytes()``` or something?

reaperhulk:
`SSL_get_version` returns a `const char *`. cffi represents that as a proxy object. If you want to obtain the value you'll need to use `ffi.buffer()` or `ffi.string()`. The former requires you either know the length of the `char *` or own all the bytes while the latter reads until it encounters a null (as it assumes a null terminated string). In this case `ffi.string` is the appropriate method

elitest:
That is what I had in whatever 41ac0c4 was before the rebase.  Are we saying that we just don't decode it to utf-8 when we return?

elitest:
Tests should now pass.

reaperhulk:
needs a rebase for travis to test it

reaperhulk:
@hynek this LGTM but since you had some API opinions around bytes earlier I'll let you merge if you think it's ready :)

doc/api/ssl.rst
```diff
@@ -598,6 +598,22 @@ Connection objects have the following methods:
     but not it returns the entire list in one go.
 
 
+.. py:method:: Connection.get_protocol_version()
+
+    Retrieve the version of the SSL or TLS protocol used by the Connection.
+    For example, it will return ``0x303`` for connections made over TLS
+    version 1.2, or ``Unknown`` for connections that were not successfully
```
hynek: that seems unlikely :)

OpenSSL/SSL.py
```diff
@@ -1883,6 +1883,18 @@ def get_cipher_version(self):
             return version.decode("utf-8")
 
 
+    def get_protocol_version_name(self):
+        """
+        Obtain the protocol version of the current connection.
+
+        :returns: The TLS version of the current connection, for example
+            the value for TLS 1.2 would be ``b'TLSv1.2'``.
+        :rtype: :py:class:`unicode`
```
hynek: I believe that’s not true anymore?

OpenSSL/SSL.py
```diff
@@ -1938,6 +1950,17 @@ def get_alpn_proto_negotiated(self):
         return _ffi.buffer(data[0], data_len[0])[:]
 
 
+    def get_protocol_version(self):
```
hynek: please group those two methods together

ChangeLog
```diff
@@ -1,3 +1,11 @@
+2015-05-27  Jim Shaver <dcypherd@gmail.com>
+
+	* OpenSSL/SSL.py, : Add ``get_protocol_version()`` and
```
hynek: extra comma plus space

OpenSSL/SSL.py
```diff
@@ -1938,6 +1950,17 @@ def get_alpn_proto_negotiated(self):
...
+        """
+        Obtain the protocol version of the current connection.
+
+        :returns: The TLS version of the current connection, for example
+            the value for TLS 1.2 would be 0x303.
+        :rtype: :py:class:`int`
+        """
+        version = _lib.SSL_version(self._ssl)
+        return version
+
```
hynek: this needs three empty lines

ChangeLog
```diff
@@ -1,3 +1,11 @@
+2015-05-27  Jim Shaver <dcypherd@gmail.com>
+
+	* OpenSSL/SSL.py, : Add ``get_protocol_version()`` and 
+	  ``get_protocol_version_name()`` to Connection
+	  Based on work from Rich Moore
+	* OpenSSL/test/test_crypto.py: tests for ``get_protocol_version()``
```
hynek: just add  OpenSSL/test/test_crypto.py to the list above.  it’s implicit that we add tests for new code. :)

ChangeLog
```diff
@@ -1,3 +1,11 @@
+2015-05-27  Jim Shaver <dcypherd@gmail.com>
+
+	* OpenSSL/SSL.py, : Add ``get_protocol_version()`` and 
+	  ``get_protocol_version_name()`` to Connection
```
hynek: I’d like Connection to be in `` too.  Also please add some periods after sentences. :)

hynek:
I didn’t have opinions as much as I’m trying to figure out what the correct way to do is. :)  As far as I can see, bytes is the way to go but if anyone can show me similar APIs that take/return unicode, I’ll change my mind.

elitest:
I'm fine either way.  There is some [Cipher](https://github.com/pyca/pyopenssl/blob/master/OpenSSL/SSL.py#L1839-1883) stuff that is returned in unicode for example.  But if a discussion has happened around it and that is the direction we are taking additions I'm fine with that.

doc/api/ssl.rst
elitest: removed unknown in 208438c

OpenSSL/SSL.py
elitest: 208438c now bytes

OpenSSL/SSL.py
elitest: moved in 208438c

OpenSSL/SSL.py
elitest: there are three lines at the bottom, it has been moved 208438c

ChangeLog
elitest: 208438c

ChangeLog
elitest: a923e93

hynek:
Ugghhh, and those methods are actually pretty analogue to these ones. :-/  Same for Curve.name

*sigh*

Unicode it is then. Sorry.

elitest:
LOL!  It is my own fault I should have pointed it out earlier.

OpenSSL/SSL.py
```diff
@@ -1883,6 +1883,31 @@ def get_cipher_version(self):
             return version.decode("utf-8")
 
 
+    def get_protocol_version_name(self):
+        """
+        Obtain the protocol version of the current connection.
+
+        :returns: The TLS version of the current connection, for example
+            the value for TLS 1.2 would be ``TLSv1.2``or ``Unknown``
+            for connections that were not successfully.
```
hynek: successfully what?

doc/api/ssl.rst
```diff
@@ -598,6 +598,20 @@ Connection objects have the following methods:
...
+.. py:method:: Connection.get_protocol_version()
+
+    Retrieve the version of the SSL or TLS protocol used by the Connection.
+    For example, it will return ``0x769`` for connections made over TLS
+    version 1.
+
+
+.. py:method:: Connection.get_protocol_version_name()
+
+    Retrieve the version of the SSL or TLS protocol used by the Connection.
```
hynek: “…as an unicode string.” would be nice to differentiate it from the other method, no?

hynek:
Thank you for your work and your patience!

elitest:
Thanks for your guidance!

PR #245, opened 2015-04-29
Begin making examples Python 3 compatible

elitest:
This PR makes a few of the examples work under Python 3 in addition to Python 2.  I ran some tests under 2.7.8 and 3.4.2.

I also found a couple of md5s being used during signing in certgen.py, I'm proposing that they be bumped up to SHA-256.

### DIFF
--- a/examples/certgen.py
+++ b/examples/certgen.py
@@ -45,14 +45,14 @@ def createCertRequest(pkey, digest="md5", **name):
     req = crypto.X509Req()
     subj = req.get_subject()
 
-    for (key,value) in name.items():
+    for (key,value) in list(name.items()):
         setattr(subj, key, value)
 
     req.set_pubkey(pkey)
     req.sign(pkey, digest)
     return req
 
-def createCertificate(req, (issuerCert, issuerKey), serial, (notBefore, notAfter), digest="md5"):
+def createCertificate(req, issuerCertKey, serial, validityPeriod, digest="md5"):
     """
     Generate a certificate given a certificate request.
 
@@ -67,6 +67,8 @@ def createCertificate(req, (issuerCert, issuerKey), serial, (notBefore, notAfter
                digest     - Digest method to use for signing, default is md5
     Returns:   The signed certificate in an X509 object
     """
+    (issuerCert, issuerKey) = issuerCertKey
+    (notBefore, notAfter) = validityPeriod
     cert = crypto.X509()
     cert.set_serial_number(serial)
     cert.gmtime_adj_notBefore(notBefore)
--- a/examples/certgen.py
+++ b/examples/certgen.py
@@ -45,7 +45,7 @@ def createCertRequest(pkey, digest="sha256", **name):
     req = crypto.X509Req()
     subj = req.get_subject()
 
-    for (key,value) in list(name.items()):
+    for key,value in name.items():
         setattr(subj, key, value)
 
     req.set_pubkey(pkey)
@@ -67,8 +67,8 @@ def createCertificate(req, issuerCertKey, serial, validityPeriod, digest="sha256
                digest     - Digest method to use for signing, default is sha256
     Returns:   The signed certificate in an X509 object
     """
-    (issuerCert, issuerKey) = issuerCertKey
-    (notBefore, notAfter) = validityPeriod
+    issuerCert, issuerKey = issuerCertKey
+    notBefore, notAfter = validityPeriod
     cert = crypto.X509()
     cert.set_serial_number(serial)
     cert.gmtime_adj_notBefore(notBefore)
--- a/examples/certgen.py
+++ b/examples/certgen.py
@@ -25,7 +25,7 @@ def createKeyPair(type, bits):
     pkey.generate_key(type, bits)
     return pkey
 
-def createCertRequest(pkey, digest="md5", **name):
+def createCertRequest(pkey, digest="sha256", **name):
     """
     Create a certificate request.
 
@@ -52,7 +52,7 @@ def createCertRequest(pkey, digest="md5", **name):
     req.sign(pkey, digest)
     return req
 
-def createCertificate(req, issuerCertKey, serial, validityPeriod, digest="md5"):
+def createCertificate(req, issuerCertKey, serial, validityPeriod, digest="sha256"):
     """
     Generate a certificate given a certificate request.
 
--- a/examples/certgen.py
+++ b/examples/certgen.py
@@ -45,7 +45,7 @@ def createCertRequest(pkey, digest="sha256", **name):
     req = crypto.X509Req()
     subj = req.get_subject()
 
-    for key,value in name.items():
+    for key, value in name.items():
         setattr(subj, key, value)
 
     req.set_pubkey(pkey)
--- a/examples/certgen.py
+++ b/examples/certgen.py
@@ -30,7 +30,7 @@ def createCertRequest(pkey, digest="sha256", **name):
     Create a certificate request.
 
     Arguments: pkey   - The key to associate with the request
-               digest - Digestion method to use for signing, default is md5
+               digest - Digestion method to use for signing, default is sha256
                **name - The name of the subject of the request, possible
                         arguments are:
                           C     - Country name
@@ -64,7 +64,7 @@ def createCertificate(req, issuerCertKey, serial, validityPeriod, digest="sha256
                             starts being valid
                notAfter   - Timestamp (relative to now) when the certificate
                             stops being valid
-               digest     - Digest method to use for signing, default is md5
+               digest     - Digest method to use for signing, default is sha256
     Returns:   The signed certificate in an X509 object
     """
     (issuerCert, issuerKey) = issuerCertKey
--- a/examples/mk_simple_certs.py
+++ b/examples/mk_simple_certs.py
@@ -7,11 +7,15 @@ from certgen import *   # yes yes, I know, I'm lazy
 cakey = createKeyPair(TYPE_RSA, 1024)
 careq = createCertRequest(cakey, CN='Certificate Authority')
 cacert = createCertificate(careq, (careq, cakey), 0, (0, 60*60*24*365*5)) # five years
-open('simple/CA.pkey', 'w').write(crypto.dump_privatekey(crypto.FILETYPE_PEM, cakey))
-open('simple/CA.cert', 'w').write(crypto.dump_certificate(crypto.FILETYPE_PEM, cacert))
+open('simple/CA.pkey', 'w').write(crypto.dump_privatekey(crypto.FILETYPE_PEM, cakey).decode('utf-8')
+)
+open('simple/CA.cert', 'w').write(crypto.dump_certificate(crypto.FILETYPE_PEM, cacert).decode('utf-8')
+)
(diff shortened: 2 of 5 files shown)


examples/certgen.py
```diff
@@ -45,14 +45,14 @@ def createCertRequest(pkey, digest="sha256", **name):
...
-    for (key,value) in name.items():
+    for (key,value) in list(name.items()):
         setattr(subj, key, value)
 
     req.set_pubkey(pkey)
     req.sign(pkey, digest)
     return req
 
-def createCertificate(req, (issuerCert, issuerKey), serial, (notBefore, notAfter), digest="md5"):
+def createCertificate(req, issuerCertKey, serial, validityPeriod, digest="sha256"):
```
elitest: According to [PEP 3113](https://www.python.org/dev/peps/pep-3113/) tuple unpacking is deprecated.

examples/certgen.py
```diff
@@ -45,14 +45,14 @@ def createCertRequest(pkey, digest="sha256", **name):
     req = crypto.X509Req()
     subj = req.get_subject()
 
-    for (key,value) in name.items():
+    for (key,value) in list(name.items()):
```
hynek: Why do you convert them into a list? Please don’t. :)

Also let’s just write `key, value` instead of `(key,value)`.

examples/certgen.py
```diff
@@ -64,9 +64,11 @@ def createCertificate(req, (issuerCert, issuerKey), serial, (notBefore, notAfter
                             starts being valid
                notAfter   - Timestamp (relative to now) when the certificate
                             stops being valid
-               digest     - Digest method to use for signing, default is md5
+               digest     - Digest method to use for signing, default is sha256
     Returns:   The signed certificate in an X509 object
     """
+    (issuerCert, issuerKey) = issuerCertKey
+    (notBefore, notAfter) = validityPeriod
```
hynek: again, no need for parens

examples/mk_simple_certs.py
```diff
@@ -7,15 +7,18 @@
 cakey = createKeyPair(TYPE_RSA, 2048)
 careq = createCertRequest(cakey, CN='Certificate Authority')
 cacert = createCertificate(careq, (careq, cakey), 0, (0, 60*60*24*365*5)) # five years
+
 print('Creating Certificate Authority private key in "simple/CA.pkey"')
-open('simple/CA.pkey', 'w').write(crypto.dump_privatekey(crypto.FILETYPE_PEM, cakey))
+open('simple/CA.pkey', 'w').write(crypto.dump_privatekey(crypto.FILETYPE_PEM, cakey).decode('utf-8'))
```
hynek: this should be converted  into a context manager

examples/mk_simple_certs.py
```diff
@@ -7,15 +7,18 @@
...
 print('Creating Certificate Authority certificate in "simple/CA.cert"')
-open('simple/CA.cert', 'w').write(crypto.dump_certificate(crypto.FILETYPE_PEM, cacert))
+open('simple/CA.cert', 'w').write(crypto.dump_certificate(crypto.FILETYPE_PEM, cacert).decode('utf-8'))
```
hynek: this should be converted  into a context manager

examples/mk_simple_certs.py
```diff
@@ -7,15 +7,18 @@
...
+
 for (fname, cname) in [('client', 'Simple Client'), ('server', 'Simple Server')]:
     pkey = createKeyPair(TYPE_RSA, 2048)
     req = createCertRequest(pkey, CN=cname)
     cert = createCertificate(req, (cacert, cakey), 1, (0, 60*60*24*365*5)) # five years
+
     print('Creating Certificate %s private key in "simple/%s.pkey"' % (fname, fname))
-    open('simple/%s.pkey' % (fname,), 'w').write(crypto.dump_privatekey(crypto.FILETYPE_PEM, pkey))
+    open('simple/%s.pkey' % (fname,), 'w').write(crypto.dump_privatekey(crypto.FILETYPE_PEM, pkey).decode('utf-8'))
```
hynek: this should be converted  into a context manager

examples/mk_simple_certs.py
```diff
@@ -7,15 +7,18 @@
...
     print('Creating Certificate %s certificate in "simple/%s.cert"' % (fname, fname))
-    open('simple/%s.cert' % (fname,), 'w').write(crypto.dump_certificate(crypto.FILETYPE_PEM, cert))
+    open('simple/%s.cert' % (fname,), 'w').write(crypto.dump_certificate(crypto.FILETYPE_PEM, cert).decode('utf-8'))
```
hynek: this should be converted  into a context manager.

examples/simple/client.py
```diff
@@ -13,11 +13,11 @@
 
 def verify_cb(conn, cert, errnum, depth, ok):
     # This obviously has to be updated
-    print 'Got certificate: %s' % cert.get_subject()
+    print('Got certificate: %s' % cert.get_subject())
```
hynek: please make the % arguments a tuple since you’re touching the line

examples/simple/server.py
```diff
@@ -14,11 +14,11 @@
 
 def verify_cb(conn, cert, errnum, depth, ok):
     # This obviously has to be updated
-    print 'Got certificate: %s' % cert.get_subject()
+    print('Got certificate: %s' % cert.get_subject())
```
hynek: same here, tuple please

examples/simple/server.py
```diff
@@ -57,14 +57,14 @@ def dropClient(cli, errors=None):
 
 while 1:
     try:
-        r,w,_ = select.select([server]+clients.keys(), writers.keys(), [])
+        r,w,_ = select.select([server]+list(clients.keys()), list(writers.keys()), [])
```
hynek: please add missing spaces after `,` and around `+`

examples/simple/server.py
```diff
@@ -88,13 +88,13 @@ def dropClient(cli, errors=None):
...
-        except SSL.Error, errors:
+        except SSL.Error as errors:
             dropClient(cli, errors)
         else:
             writers[cli] = writers[cli][ret:]
             if writers[cli] == '':
                 del writers[cli]
 
-for cli in clients.keys():
+for cli in list(clients.keys()):
```
hynek: don’t convert it to a list please.

examples/mk_simple_certs.py
elitest: Thanks for the feedback @hynek.  I thought contexts were for creating connections with certs that had already been created.  Or am I missing something?

alex: He meant a context manager:

```py
with open(...) as f:
    f.write(...)
```

On Wed, Apr 29, 2015 at 10:25 PM, elitest <notifications@github.com> wrote:

> In examples/mk_simple_certs.py
> <https://github.com/pyca/pyopenssl/pull/245#discussion_r29399782>:
>
> >  print('Creating Certificate Authority private key in "simple/CA.pkey"')
> > -open('simple/CA.pkey', 'w').write(crypto.dump_privatekey(crypto.FILETYPE_PEM, cakey))
> > +open('simple/CA.pkey', 'w').write(crypto.dump_privatekey(crypto.FILETYPE_PEM, cakey).decode('utf-8'))
>
> Thanks for the feedback @hynek <https://github.com/hynek>. I thought
> contexts were for creating connections with certs that had already been
> created. Or am I missing something?
>
> —
> <https://github.com/pyca/pyopenssl/pull/245/files#r29399782>.
>



-- 

examples/simple/server.py
```diff
@@ -8,17 +8,19 @@
...
+from OpenSSL import SSL, crypto
 import sys, os, select, socket
 
 
 def verify_cb(conn, cert, errnum, depth, ok):
     # This obviously has to be updated
-    print 'Got certificate: %s' % cert.get_subject()
+    certsubject = crypto.X509Name(cert.get_subject())
+    commonname = certsubject.commonName
+    print(('Got certificate: ' + commonname))
```
alex: Why are these changes needed, particularly the `crypto.X509Name(...)` bit?

elitest: I should have made a comment to clarify :).

This is what you get on master:
`Got certificate: <X509Name object '/CN=Certificate Authority'>`
`Got certificate: <X509Name object '/CN=Simple Server'>`

This is what you get with that:
`Got certificate: Certificate Authority`
`Got certificate: Simple Server`

I think that is the direction the orig author was going in when he said: "# This obviously has to be updated".  I can remove the comment if you agree.  I suppose there are less spoofable things that we could use to identify the certs in addition to the CN, like the fingerprint.

examples/certgen.py
elitest: Fixed in 473fe6a

examples/certgen.py
elitest: Fixed in 473fe6a

examples/mk_simple_certs.py
elitest: Fixed in aab9ddd

examples/mk_simple_certs.py
elitest: Fixed in aab9ddd

examples/mk_simple_certs.py
elitest: Fixed in aab9ddd

examples/mk_simple_certs.py
elitest: Fixed in aab9ddd

examples/simple/client.py
elitest: Fixed in b2ff5be

examples/simple/server.py
elitest: Fixed in b2ff5be

examples/simple/server.py
elitest: lists removed(4852ef4) and then re-added(b2ff5be) as will not work under python3.  Spacing fixed in those commits.

examples/simple/server.py
elitest: Fixed in 473fe6a

examples/certgen.py
hynek: it’s actually even *removed* in Python 3 (which kind of sucks when doing Twisted)

examples/simple/server.py
hynek: yeah please get rid of the comment.  there’s nothing obvious about it.

elitest:
Thanks to @hynek's work in #253 this now passes.  Cheers.

hynek:
So, stupid question: is there currently an easy way to actually test the examples?

elitest:
the 'easy' way I used to test them from the examples directory was:
run 'python mk_simple_certs.py'
run 'python simple/server.py 8443'
run 'python simple/client.py localhost 8443'

then type some text in, and server.py should echo it back to you, as well as tell you what certificate it received.  It actually just gives the CN as opposed to something that couldn't be spoofed like a fingerprint.  Then I ran the same against python3.

hynek:
I see, thanks!

Please add `examples/simple/*.cert` and `examples/simple/*.pkey` to .gitignore then.

hynek:
Thanks!

PR #246, opened 2015-04-30
Add del_extension needed by Zorp

endreszabo:
[Balabit Zorp](https://github.com/balabit/zorp) application level firewall [relies on the del_extension function](https://github.com/balabit/zorp/blob/master/pylib/Zorp/Keybridge.py#L350) to remove attributes like CRL pathes from the mimicked  certificates when doing man-in-the-middle traffic filtering.

Also needs the ffi lib (python-cryptograpy at hazmat/bindings/openssl/x509.py) to have the definition of:

``` c
X509_EXTENSION *X509_delete_ext(X509 *, int);
```

### DIFF
--- a/OpenSSL/crypto.py
+++ b/OpenSSL/crypto.py
@@ -1340,6 +1340,19 @@ class X509(object):
         extension = _lib.X509_EXTENSION_dup(ext._extension)
         ext._extension = _ffi.gc(extension, _lib.X509_EXTENSION_free)
         return ext
+        
+    def del_extension(self, index):
+        """
+        Delete a specific extension of the certificate by index.
+
+        :param index: The index of the extension to delete.
+        :return: The X509Extension object deleted at the specified index.
+        """
+        ext = X509Extension.__new__(X509Extension)
+        ext._extension = _lib.X509_delete_ext(self._x509, index)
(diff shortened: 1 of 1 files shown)


OpenSSL/crypto.py
```diff
@@ -1340,6 +1340,19 @@ def get_extension(self, index):
...
+        
+    def del_extension(self, index):
+        """
+        Delete a specific extension of the certificate by index.
+
+        :param index: The index of the extension to delete.
+        :return: The X509Extension object deleted at the specified index.
+        """
+        ext = X509Extension.__new__(X509Extension)
+        ext._extension = _lib.X509_delete_ext(self._x509, index)
```
reaperhulk: This will leak memory. It needs to use `X509_EXTENSION_free` in conjunction with `ffi.gc`. Take a look at the method right above this for an example.

reaperhulk:
This also needs unit tests. Thanks for your contribution so far!

OpenSSL/crypto.py
endreszabo: Ah, really! Thanks for pointing that out.

hynek:

reaperhulk:
Closing due to lack of activity.

PR #252, opened 2015-05-04
fix a leak in _subjectAltNameString

reaperhulk:


hynek:
When was that leak introduced? It may warrant a change log entry.

reaperhulk:
I haven't looked at the history but I'd guess 0.14.

PR #253, opened 2015-05-04
Fix CI tests against cryptography master

hynek:
Trying to fix the problem we ran into in #244 – need to open a PR so Travis runs.


alex:
slick

reaperhulk:
It's hard (impossible?) to tell whether the proper version of cryptography is being installed into the tox env. Do you think it makes sense to add a new line to the command list like this?
```bash
python -c "import cryptography; print(cryptography.__version__)"
```

hynek:
I completely agree.  Print added.

PR #257, opened 2015-05-08
Allow for using the raw OpenSSL verify callback

Lukasa:
PyOpenSSL has the ability to set the verify callback, but it tries to be all 'helpful' and 'user friendly' and all that hippy crap, so it hides the terrible OpenSSL nonsense from me.

Sadly, for shazow/urllib3#607 I need access to the X509_STORE_CTX that PyOpenSSL is hiding. This patch allows me to set a _different_ verify callback that does that.

This is only a proposed API, I'm not wedded to it at all, so feedback on the API and naming is totally welcomed.

### DIFF
--- a/OpenSSL/SSL.py
+++ b/OpenSSL/SSL.py
@@ -231,6 +231,36 @@ class _VerifyHelper(_CallbackExceptionHelper):
             "int (*)(int, X509_STORE_CTX *)", wrapper)
 
 
+class _ExplicitVerifyHelper(_CallbackExceptionHelper):
+    """
+    Wrap a callback such that it can be used as a certificate verification
+    callback without the extra wrapping of ``_VerifyHelper``.
+    """
+    def __init__(self, callback):
+        _CallbackExceptionHelper.__init__(self)
+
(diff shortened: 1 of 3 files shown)


hynek:
I recall running into the problems from this “helpfulness” too once.  But why explicit?  Sounds more like “raw” or something? (please don’t fix yet ;))

Lukasa:
:+1: Explicit is definitely a *bad* name, but I honestly couldn't think of a better one. Raw might work, but I'll happily take suggestions.

hynek:
I think this would be a good thing to add in general.  So unless @alex or @reaperhulk see anything questionable, I’d say to go forward, rename it to raw and make sure it shows up in the docs + changelog.

Lukasa:
@hynek I think I've updated this to do what you wanted, let me know if you want further changes. =)

hynek:
Sorry lost track of this.  Would you mind bringing it up to shape or do you now care anymore?

reaperhulk:
Going to go ahead and close this due to inactivity. We can always reopen later.

PR #260, opened 2015-05-14
Fix Travis CI

hynek:
`sys.getfilesystemencoding()` seems to be returning `ascii` now which breaks all builds.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,8 @@
 sudo: false
 language: python
+env:
+  global:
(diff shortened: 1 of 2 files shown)


alex:
`tox` doesn't pass environment variables (besides `PATH`) into the test process. You'd have to add `LC_ALL` to `passenv` in `tox.ini`

hynek:
thanks! next attempt.

hynek:
Tests are green again; please review/merge.

Issue #239, opened 2015-04-18
Make test suite pass on 0.9.8


hynek:
For example https://travis-ci.org/pyca/pyopenssl/jobs/59016116

- [x] PKCS12Tests.test_export_without_args
- [x] PKCS12Tests.test_export_without_bytes
- [x] PKCS12Tests.test_export_without_mac
- [x] PKCS12Tests.test_friendly_name
- [x] PKCS12Tests.test_various_empty_passphrases
- [x] FunctionTests.test_dump_certificate
- [x] FunctionTests.test_dump_certificate_request
- [x] CRLTests.test_export_text
- [x] ConnectionTests.test_shutdown_closed (passes on OS X)
- [x] ConnectionTests.test_shutdown_truncated (passes on OS X)
- [x] ContextTests.test_set_default_verify_paths

I think it’s safe to assume that these are three actual issues tops that need to be actually fixed.

hynek:
Side-note: the shutdown tests pass on OS X.

hynek:
They actually pass *all* on OS X now we have a 0.9.8 builder.

(One of?) the background(s?) probably is that the verification is done by running an openssl in the current PATH.  If you run the tests with an pyOpenSSL linked against 0.9.8 but verify with pyOpenSSL 1.0.2, it failes.  If both are ran with 0.9.8 they pass.

hynek:
To fix this test, we have to find out how to use additionally installed openssl binary on Ubuntu.  Could be actually easy.

Issue #240, opened 2015-04-18
Make test suite pass on 1.0.2


hynek:
E.g. https://travis-ci.org/pyca/pyopenssl/jobs/59016103

- [x] X509Tests.test_root_has_not_expired (actually an OpenSSL 1.0.2(a) bug, we could add a skip though?) cf #192
- [x] PKCS12Tests.test_export_without_args
- [x] PKCS12Tests.test_export_without_bytes
- [x] PKCS12Tests.test_export_without_mac
- [x] PKCS12Tests.test_friendly_name
- [x] PKCS12Tests.test_various_empty_passphrases
- [x] FunctionTests.test_dump_certificate
- [x] FunctionTests.test_dump_certificate_request
- [x] CRLTests.test_export_text
- [x] ContextTests.test_set_tmp_ecdh
- [x] ConnectionTests.test_state_string

Pretty sure most of them will go away if we fix #239 properly.

hynek:
The OpenSSL bug aside, the only problem left is:

```
________________________ ContextTests.test_set_tmp_ecdh ________________________
OpenSSL/test/util.py:194: in tearDown
    "Left over errors in OpenSSL error queue: " + repr(e)
E   AssertionError: Left over errors in OpenSSL error queue: Error([('bignum routines', 'BN_mod_inverse', 'no inverse'), ('bignum routines', 'BN_mod_inverse', 'no inverse')],)
```

FYI, `test_set_tmp_ecdh` iterates over `get_elliptic_curves()` and sets them on a context.

Ideas/suggestions?

hynek:
A new one:

```
______________________________________________________________________ ConnectionTests.test_state_string ______________________________________________________________________

self = <OpenSSL.test.test_ssl.ConnectionTests testMethod=test_state_string>

    def test_state_string(self):
        """
            :py:obj:`Connection.state_string` verbosely describes the current
            state of the :py:obj:`Connection`.
            """
        (server, client) = socket_pair()
        server = self._loopbackServerFactory(server)
        client = self._loopbackClientFactory(client)

        self.assertEqual('before/accept initialization',
                         server.state_string().decode())
        self.assertEqual('before/connect initialization',
                         client.state_string().decode())

        for conn in [server, client]:
            try:
                conn.do_handshake()
            except WantReadError:
                pass

        self.assertEqual('SSLv3 read client hello B',
                         server.state_string().decode())
        self.assertEqual('SSLv3 read server hello A',
                         client.state_string().decode())

[...]

PR #263, opened 2015-05-14
Ensure OpenSSL CLI matches lib on homebrew

hynek:
Some of the test failures might be caused by just the CLI version mismatch.

cf. https://github.com/pyca/pyopenssl/issues/239#issuecomment-102052023


hynek:
This fixes indeed most of the failures from #240!

please review/merge

Issue #181, opened 2014-12-30
CRL parameters missing from CRL objects


adulau:
pyopenssl doesn't provide the following parameters when parsing CRL objects (not the revoked objects)

```
    Issuer
    Last Update
    Next Update
    CRL extensions
        X509v3 CRL Number
```

Do you plan to add it?

Thank you very much.

PS: It seems to be a recurring issue cf http://stackoverflow.com/questions/13694984/pyopenssl-get-a-crls-last-update-and-next-update-fields

schlenk:
You can get those things via pyasn1 (the pyasn1.util.rfc2459 module has the needed ASN.1 definitions...). But it is pretty ugly to do it that way and having access to the CRL data via pyopenssl (or Cryptography below it) would be useful.

exarkun:
This would be a fine addition to pyOpenSSL.  If someone would like to submit a PR (with tests and documentation), I'd be happy to review it.

eldoon:
Can we expect to see this in 16 Dev?

reaperhulk:
When parsing CRLs use cryptography: https://cryptography.io/en/latest/x509/reference/#cryptography.x509.CertificateRevocationList

PR #264, opened 2015-05-15
Expose some CRL attributes (#181)

carletes:
This pull request addresses pyca/pyopenssl#181, exposing some useful attributes of CRL objects:
- Issuer
- Last Update
- Next Update
- CRL extensions

### DIFF
--- a/OpenSSL/crypto.py
+++ b/OpenSSL/crypto.py
@@ -2081,6 +2081,80 @@ class CRL(object):
             _raise_current_error()
 
         return _bio_to_string(bio)
+
+
+    def get_issuer(self):
+        """
+        Return the issuer of this CRL.
+
+        :return: A :py:class:`X509Name` object.
+        """
+        name = X509Name.__new__(X509Name)
+        name._name = _lib.X509_CRL_get_issuer(self._crl)
+        if name._name == _ffi.NULL:
+            # TODO: This is untested.
+            _raise_current_error()
+
+        # The name is owned by the CA structure.  As long as the X509Name
+        # Python object is alive, keep the CA Python object alive.
+        name._owner = self
+
+        return name
+
+
+    def get_last_update(self):
+        """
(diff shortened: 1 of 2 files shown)


sholsapp:
Setting some of these attributes available in  https://github.com/pyca/pyopenssl/issues/281

hynek:
Since I’m woefully uninformed about CRLs, I would like @reaperhulk to comment on this.

braingu-gil:
Is there an update on the status of this pull request?  This feature would be extremely useful for anyone using CRLs.

hynek:

carletes:
@hynek: No apologies needed! Except for mine, for taking this long to answer, that is.

I understand from @glyph's https://mail.python.org/pipermail/cryptography-dev/2015-December/000542.html and the other messages that pyOpenSSL should be eventually be deprecated, and we should all be using Cryptography. Is that correct?

I'm fine with this, and I can already do that for my code (in fact I submitted this pull request because I was not aware of this). If my understanding is correct, I'm happy to close this pull request and switch to Cryptography entirely. What's your advice?

And thanks a lot for taking the burden of maintaining pyOpenSSL, by the way!

glyph:
@carletes My personal opinion; pyOpenSSL is still definitely maintained, but if you *can* switch over to purely using Cryptography's X509 stuff, without using pyOpenSSL at all, you probably should.  If you are implementing TLS, pyOpenSSL is still probably a better way to go for now.

apeduru:
What is the status of this PR? I find this feature extremely useful

reaperhulk:
Our general policy at this point is that if it's possible to do in cryptography developers should do it there (where we have more modern APIs and a more active community). Since this is possible to do in cryptography (you can even convert to a cryptography CRL object by using `to_cryptography` on your existing pyOpenSSL CRL object) I'm going to go ahead and close this PR.

PR #272, opened 2015-05-28
Use RTD theme

hynek:
We should really use the much nicer theme by RTD.

According to the [docs](http://docs.readthedocs.org/en/latest/theme.html), not setting a theme should do the trick.


reaperhulk:
check-manifest is upset with this :)

hynek:
yeah I forgot to actually commit the fix (just staged it and didn’t read the output of `git push` m(

hynek:
Passes now.

PR #278, opened 2015-06-10
Clean up X509StoreContext documentation

sholsapp:
This change cleans up the X509StoreContext and X509StoreContextError
documentation to be more aligned with the other documentation.

Tested using ``tox -e docs``.

### DIFF
--- a/OpenSSL/crypto.py
+++ b/OpenSSL/crypto.py
@@ -1503,12 +1503,11 @@ X509StoreType = X509Store
 
 class X509StoreContextError(Exception):
     """
-    An error occurred while verifying a certificate using
-    `OpenSSL.X509StoreContext.verify_certificate`.
+    An exception raised when an error occurred while verifying a certificate
+    using `OpenSSL.X509StoreContext.verify_certificate`.
 
     :ivar certificate: The certificate which caused verificate failure.
-    :type cert: :class:`X509`
-
+    :type certificate: :class:`X509`
     """
     def __init__(self, message, certificate):
         super(X509StoreContextError, self).__init__(message)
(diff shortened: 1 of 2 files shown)


OpenSSL/crypto.py
```diff
@@ -1608,8 +1612,13 @@ def verify_certificate(self):
         """
         Verify a certificate in a context.
 
+        .. versionadded:: 0.15
+
         :param store_ctx: The :py:class:`X509StoreContext` to verify.
-        :raises: Error
+
+        :raises X509StoreContextError: If an error occured when validating a
+          certificate in the context. Sets ``certifcate`` attribute to indicate
```
hynek: typo in certificate, also please put two spaces after periods, thanks. :)

OpenSSL/crypto.py
```diff
@@ -1520,11 +1519,15 @@ class X509StoreContext(object):
     An X.509 store context.
 
     An :py:class:`X509StoreContext` is used to define some of the criteria for
-    certificate verification.  The information encapsulated in this object
+    certificate verification. The information encapsulated in this object
```
hynek: two spaces after periods are on purpose, please don’t remove them. :)

OpenSSL/crypto.py
sholsapp: Woops, vim's spell works but my eyeballs don't. Fixed both issues.

OpenSSL/crypto.py
sholsapp: *Reallllllly*? That's pretty old school... but fixed nevertheless. :) Will try to keep in mind but that's a habit I had to break a long time ago to escape from school.

reaperhulk: Haha, nice @hynek. iOS retrained me since double space turns into a single space and a period. Prior to 2007 I was a double space after period loyalist though!

hynek:
Thanks!  Our docs need love badly.

Issue #256, opened 2015-05-08
X509StoreContext should accept CRLs


bgaifullin:
Now, there is no possibility to use CRLS with verify_certificate method

sholsapp:
Hey @bgaifullin, you're right, it should. It also should accept quite a few other options, too. I talked about this a little bit in https://github.com/pyca/pyopenssl/pull/155#discussion_r24637816 which is the original commit that added the `verify_certificate` method.

As a stop gap while we make these changes, you can take an approach similar to that of @mrjefftang in https://github.com/pyca/pyopenssl/pull/155#discussion_r24635378 to implement the missing functionality directly in your client code.

I'll get to this soon, sorry!

tisdall:
I was trying to add CRL to my twisted project and found this issue...  I'm surprised this hasn't been more of an issue.  There's code to make CRL objects and Revoked objects, but what can you do with those if you can't use them to verify a certificate?

carletes:
While this gets implemented I'm doing something like this:

```python
from OpenSSL import crypto

def verify(ca_cert_pem, crl_pem, cert_pem):
    store = crypto.X509Store()
    store.add_cert(crypto.load_certificate(crypto.FILETYPE_PEM, ca_cert_pem))
    cert = crypto.load_certificate(crypto.FILETYPE_PEM, cert_pem)
    ctx = crypto.X509StoreContext(store, cert)
    ctx.verify_certificate()

    # Until `X509StoreContext` accepts CRLs
    crl = crypto.load_crl(crypto.FILETYPE_PEM, crl_pem)
    revoked = crl.get_revoked() or []
    for r in revoked:
        r_serial = r.get_serial()
        c_serial = "%X" % (cert.get_serial_number(),)
        if r_serial == c_serial:
            raise Exception("Certificate revoked")
```

schlenk:
Currently you can do nearly nothing with CRL objects. You can manually compare the get_revoked() results (in the contorted way seen above by @carletes ) with the serial number of the cert.
But the recent cryptography 0.9 added the needed low level APIs to access a few more things in those objects and to add them to a verify stack.

sholsapp:
@schlenk is anyone working on using these low level APIs to enhance the CRL objects? After taking a peek at this to get started, I realized as you pointed out that you can do nearly nothing with the CRL objects. I think at a minimum we need a way to get a handle on a X509 object that represents the revoked certificate.

[...]

Issue #258, opened 2015-05-12
CRL API gotchas in get_revoked() and serial number


schlenk:
The pyopenssl API for CRLs is kind of weird.

Case 1:  crl.get_revoked() returns None for empty CRLs

The natural use case for this is to iterate over the revoked objects like this:

```
for revoked in crl.get_revoked():
     ...
```

This idiom blows up if you use a CRL with no revoked objects in it, as crl.get_revoked() returns None instead of an empty tuple/list.

Case 2: Revoked.get_serial() returns a hex encoded long integer, but X509 serial returns the long integer directly.

To check if a certificate is revoked, you need to compare the serial number of the X509 with the serial number inside the Revoked object.

```
if x509.get_serial_number() == revoked.get_serial():
   print("OOPS, doesn't work..., wrong types")
```

Only this works:

```
if x509.get_serial_number() == long(revoked.get_serial(), 16):
   print("Obvious...")
```

This is just insane. 

The API should be more natural to use in a pythonic way, so get_revoked() should return an empty iterable and the get_serial() method should return a long int.

PR #281, opened 2015-06-22
Enable use of CRL (and more) in verify context

sholsapp:
This change adds a number of features to the X509Store class in order to
allow use of certificate revocation lists (and more) in verification
contexts.

Fixes https://github.com/pyca/pyopenssl/issues/256

### DIFF
--- a/OpenSSL/crypto.py
+++ b/OpenSSL/crypto.py
@@ -1472,8 +1472,34 @@ X509Type = X509
 
 class X509Store(object):
     """
-    An X509 certificate store.
+    An X.509 store.
+
+    An X.509 store is used to describe a context in which to verify a
+    certificate. A description of a context may include a set of certificates
+    to trust, a set of certificate revocation lists, verification flags and
+    more.
+
+    An X.509 store, being only a description, cannot be used by itself to verify
+    a certificate. To carry out the actual verification process, see
+    :py:class:`X509StoreContext`.
     """
+
+    CRL_CHECK = _lib.X509_V_FLAG_CRL_CHECK
+    CRL_CHECK_ALL = _lib.X509_V_FLAG_CRL_CHECK_ALL
+    IGNORE_CRITICAL = _lib.X509_V_FLAG_IGNORE_CRITICAL
+    X509_STRICT = _lib.X509_V_FLAG_X509_STRICT
+    ALLOW_PROXY_CERTS = _lib.X509_V_FLAG_ALLOW_PROXY_CERTS
+    POLICY_CHECK = _lib.X509_V_FLAG_POLICY_CHECK
+    EXPLICIT_POLICY = _lib.X509_V_FLAG_EXPLICIT_POLICY
+    # FLAG_INHIBIT_ANY = _lib.X509_V_FLAG_FLAG_INHIBIT_ANY
+    INHIBIT_MAP = _lib.X509_V_FLAG_INHIBIT_MAP
+    NOTIFY_POLICY = _lib.X509_V_FLAG_NOTIFY_POLICY
+    # USE_DELTAS = _lib.X509_V_FLAG_USE_DELTAS
+    CHECK_SS_SIGNATURE = _lib.X509_V_FLAG_CHECK_SS_SIGNATURE
+    CB_ISSUER_CHECK = _lib.X509_V_FLAG_CB_ISSUER_CHECK
+    # NO_ALT_CHAINS = _lib.X509_V_FLAG_NO_ALT_CHAINS
+
+
     def __init__(self):
         store = _lib.X509_STORE_new()
         self._store = _ffi.gc(store, _lib.X509_STORE_free)
@@ -1481,9 +1507,10 @@ class X509Store(object):
 
     def add_cert(self, cert):
         """
-        Adds the certificate :py:data:`cert` to this store.
+        Adds a trusted certificate to this store.
 
-        This is the Python equivalent of OpenSSL's ``X509_STORE_add_cert``.
+        Adding a certificate with this method adds this certificate as a
+        *trusted* certificate.
 
         :param X509 cert: The certificate to add to this store.
         :raises TypeError: If the certificate is not an :py:class:`X509`.
@@ -1498,7 +1525,51 @@ class X509Store(object):
             _raise_current_error()
 
 
+    def add_crl(self, crl):
+        """
+        Add a certificate revocation list to this store.
+
+        The certificate revocation lists added to a store will only be used if
+        the associated flags are configured to check certificate revocation
+        lists.
+
+        .. versionadded:: 0.16
+
+        :param CRL crl: The certificate revocation list to add to this store.
+        :return: :py:data:`None` if the certificate revocation list was added successfully.
+        """
+        if _lib.X509_STORE_add_crl(self._store, crl._crl) == 0:
+            _raise_current_error()
+
+
+    def set_flags(self, flags):
+        """
+        Set verification flags to this store.
+
+        Verification flags can be combined by oring them together.
+
+        .. note::
+
+          Setting a verification flag sometimes requires clients to add
+          additional information to the store, otherwise a suitable error will
+          be raised.
+
+          For example, in setting flags to enable CRL checking a
+          suitable CRL must be added to the store otherwise an error will be
+          raised.
+
+        .. versionadded:: 0.16
+
+        :param int flags: The verification flags to set on this store.
+        :return: :py:data:`None` if the verification flags were successfully set.
+        """
+        if _lib.X509_STORE_set_flags(self._store, flags) == 0:
+            _raise_current_error()
+
+
+
 X509StoreType = X509Store
+
 
 
 class X509StoreContextError(Exception):
@@ -1518,29 +1589,19 @@ class X509StoreContext(object):
     """
     An X.509 store context.
 
-    An :py:class:`X509StoreContext` is used to define some of the criteria for
-    certificate verification.  The information encapsulated in this object
-    includes, but is not limited to, a set of trusted certificates,
-    verification parameters, and revoked certificates.
-
-    .. note::
-
-      Currently, one can only set the trusted certificates on an
-      :py:class:`X509StoreContext`.  Future versions of pyOpenSSL will expose
-      verification parameters and certificate revocation lists.
+    An X.509 store context is used to carry out the actual verification process
+    of a certificate in a described context. For describing such a context, see
+    :py:class:`X509Store`.
 
     :ivar _store_ctx: The underlying X509_STORE_CTX structure used by this
         instance.  It is dynamically allocated and automatically garbage
         collected.
-
     :ivar _store: See the ``store`` ``__init__`` parameter.
-
     :ivar _cert: See the ``certificate`` ``__init__`` parameter.
-
     :param X509Store store: The certificates which will be trusted for the
         purposes of any verifications.
-
     :param X509 certificate: The certificate to be verified.
+
     """
 
     def __init__(self, store, certificate):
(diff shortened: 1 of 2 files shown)


sholsapp:
Hey all, I submitted the PR mostly to get the testing/coverage goodness and to give others a chance to submit feedback before I spend too much longer on this. I'll ad some comments to parts I specifically have questions about.

OpenSSL/crypto.py
```diff
@@ -1472,8 +1472,34 @@ def get_extension(self, index):
...
+    An X.509 store is used to describe a context in which to verify a
+    certificate. A description of a context may include a set of certificates
+    to trust, a set of certificate revocation lists, verification flags and
+    more.
+
+    An X.509 store, being only a description, cannot be used by itself to verify
+    a certificate. To carry out the actual verification process, see
+    :py:class:`X509StoreContext`.
+    """
+
```
sholsapp: Not all of the flags below are *tested*. What do folks prefer I do here: 1) expose all of the flags below and let people use them even though they're not tested or 2) only expose the flags that have been specifically tested?

OpenSSL/crypto.py
```diff
@@ -1472,8 +1472,34 @@ def get_extension(self, index):
...
+    CRL_CHECK = _lib.X509_V_FLAG_CRL_CHECK
+    CRL_CHECK_ALL = _lib.X509_V_FLAG_CRL_CHECK_ALL
+    IGNORE_CRITICAL = _lib.X509_V_FLAG_IGNORE_CRITICAL
+    X509_STRICT = _lib.X509_V_FLAG_X509_STRICT
+    ALLOW_PROXY_CERTS = _lib.X509_V_FLAG_ALLOW_PROXY_CERTS
+    POLICY_CHECK = _lib.X509_V_FLAG_POLICY_CHECK
+    EXPLICIT_POLICY = _lib.X509_V_FLAG_EXPLICIT_POLICY
+    # FLAG_INHIBIT_ANY = _lib.X509_V_FLAG_FLAG_INHIBIT_ANY
```
sholsapp: Some flags are newer than others. How do I tell what flags are supported by my underlying OpenSSL implementation?

OpenSSL/crypto.py
```diff
@@ -2011,27 +2065,123 @@ def add_revoked(self, revoked):
...
+        if digest_obj == _ffi.NULL:
+            raise ValueError("No such digest method")
+        _lib.X509_CRL_set_issuer_name(self._crl, _lib.X509_get_subject_name(issuer_cert._x509))
+        _lib.X509_CRL_sort(self._crl)
+        sign_result = _lib.X509_CRL_sign(self._crl, issuer_key._pkey, digest_obj)
+        if not sign_result:
+            _raise_current_error()
+
 
+    def export(self, cert, key, type=FILETYPE_PEM, days=100,
```
sholsapp: This function is bad but I cannot fix it to use the new functions added in this PR without changing the public API of pyOpenSSL. I would appreciate some guidance from the maintainers on ways they want to handle public API changes. E.g., do we do semantic versioning, do we just do bust clients, something else?

This same problem is present in https://github.com/pyca/pyopenssl/issues/258. The solution is simple, but it requires a public API change or the additional of new methods.

OpenSSL/test/test_crypto.py
```diff
@@ -3431,6 +3434,66 @@ def test_modification_pre_verify(self):
...
+            # that has been reported in
+            # https://github.com/pyca/pyopenssl/issues/258
+            serial = hex(cert.get_serial_number())[2:]
+            revoked.set_serial(serial)
+            revoked.set_reason(b'unspecified')
+            revoked.set_rev_date(b'20140601000000Z')
+            crl.add_revoked(revoked)
+        crl.set_version(1)
+        crl.set_lastUpdate(b'20140601000000Z')
+        crl.set_nextUpdate(b'20180601000000Z')
```
sholsapp: I need to fix this to automatically detect the year, else we'll start failing sometime around 2018.

Sorry, but I can't resist:

what-year-is-it

sholsapp:
Also, it appears that I need the latest version of cryptography for this PR to work. How can I do that?

reaperhulk:
@sholsapp The Travis CI builds run against cryptography master in the allowed failures section. They abruptly error, which is usually a sign of a memory management problem: https://travis-ci.org/pyca/pyopenssl/jobs/67910194

sholsapp:
@reaperhulk That's what I thought re: the building against master. I'll find some time to see if I can hunt down a memory error. Thanks for the tip.

sholsapp:
@reaperhulk Any additional tip/pointers would be appreciated here. I can only generate this error by doing `TOXENV=py26 tox`. When I setup my virtual environment by developing pyOpenSSL and manually installing head of my cryptography, all seems to work as expected and py.test completes without failures.

sholsapp:
I deleted my last comment couple comments and I think this patch is a red herring.

I'm finding that building and testing pyOpenSSL against head of cryptography, even *without* this patch, is failing at the same location. I narrowed it down to failing on OpenSSL/test/test_ssl.py:361 when instantiating `Context(method)`.

Can someone else reproduce?

reaperhulk:
@sholsapp Yes I can reproduce. Will have to investigate what's going on tomorrow though.

sholsapp:
Thanks @reaperhulk. If I find anything I'll speak up.

reaperhulk:
@sholsapp calling `SSLv2_method` is causing an immediate segfault in master. This does not happen in 0.9.1. Not sure why right now.

OpenSSL/crypto.py
```diff
@@ -2011,27 +2065,123 @@ def add_revoked(self, revoked):
...
+        :return: :py:const:`None`
+        """
+        return self._set_boundary_time(_lib.X509_CRL_get_nextUpdate, when)
+
+
+    def sign(self, issuer_cert, issuer_key, digest='sha1'):
+        """
+        Sign the CRL.
+
+        Signing a CRL enables clients to assosciate the CRL itself with an
```
schlenk: typo in 'associate'

OpenSSL/crypto.py
```diff
@@ -2011,27 +2065,123 @@ def add_revoked(self, revoked):
...
+
+        .. versionadded:: 0.16
+
+        :param bytes when: A timestamp string.
+        :return: :py:const:`None`
+        """
+        return self._set_boundary_time(_lib.X509_CRL_get_lastUpdate, when)
+
+
+    def set_nextUpdate(self, when):
```
schlenk: Any good reason why this does not take a datetime.datetime object instead of a string?

OpenSSL/crypto.py
```diff
@@ -1472,8 +1472,34 @@ def get_extension(self, index):
...
+    CRL_CHECK_ALL = _lib.X509_V_FLAG_CRL_CHECK_ALL
+    IGNORE_CRITICAL = _lib.X509_V_FLAG_IGNORE_CRITICAL
+    X509_STRICT = _lib.X509_V_FLAG_X509_STRICT
+    ALLOW_PROXY_CERTS = _lib.X509_V_FLAG_ALLOW_PROXY_CERTS
+    POLICY_CHECK = _lib.X509_V_FLAG_POLICY_CHECK
+    EXPLICIT_POLICY = _lib.X509_V_FLAG_EXPLICIT_POLICY
+    # FLAG_INHIBIT_ANY = _lib.X509_V_FLAG_FLAG_INHIBIT_ANY
+    INHIBIT_MAP = _lib.X509_V_FLAG_INHIBIT_MAP
+    NOTIFY_POLICY = _lib.X509_V_FLAG_NOTIFY_POLICY
+    # USE_DELTAS = _lib.X509_V_FLAG_USE_DELTAS
```
schlenk: Why not add  Delta CRL support?

OpenSSL/crypto.py
```diff
@@ -1481,9 +1507,10 @@ def __init__(self):
 
     def add_cert(self, cert):
         """
-        Adds the certificate :py:data:`cert` to this store.
+        Adds a trusted certificate to this store.
```
schlenk: This isn't entirely correct. It is also used to add intermediate certificates (at least in your tests), which are not 'trusted' unless linked to a trusted self signed root certificate.

If X509_V_FLAG_PARTIAL_CHAINS is specified, it would be true, but that has other dangers. Currently the API to add 'chain' certificates to the StoreCTX that aren't trusted isn't used.

sholsapp: Good points. I'll update the working here later today.

OpenSSL/crypto.py
sholsapp: I found that some of these flags were added in different versions of OpenSSL. I don't know how to conditionally add them if they're present. Do you have any suggestions on conditionally adding the flag if it is available in the underlying OpenSSL?

OpenSSL/crypto.py
sholsapp: The only "good" reason I have is that other API in this module use this format. I'd insist that we keep it consistent, but I would also be happy to change the other API to use a datetime if the maintainers are okay with a backwards incompatible API change.

OpenSSL/crypto.py
sholsapp: Thanks, will fix.

reaperhulk:
As an update, this segfault is a bug in cryptography master related to conditional names not being stripped. See: https://github.com/pyca/cryptography/issues/2061

sholsapp:
Thanks for tracking that down @reaperhulk! :smile:

reaperhulk:
https://github.com/pyca/cryptography/issues/2061 has been fixed. I restarted your builds and it passes on py26 and py27 but there are some issue with the PR on py3.

sholsapp:
Thanks @reaperhulk I'll check it out soon.

hynek:
FYI, we usually don’t look at PRs unless  codecov/commit is 100% (and not 86.36%).  IOW the checks must be all green.

sholsapp:
@hynek sure thing, that's great that we're at 100% code coverage. I'll get to this soonish.

hynek:
JFTR, we're not at 100% project-wide; but we require new patches to be fully covered so we catch up one day. :)

dsully:
@hynek - I believe that 100% code coverage is an unrealistic expectation, especially given that patches like this would require the installation of bad versions of OpenSSL to do negative testing.

dsully:
@sholsapp - can you rebase this, and let's see where the code coverage is at? Likely only minor changes needed.

hynek:
Please also do #478 on it.  That should pretty much solve all coverage problems.

dsully:
I have a working patch based on @sholsapp's change, rebased to HEAD & using _openssl_assert().

What's the best way to get that change into this PR? Or do I need a new PR by it's nature?

alex:
New PR is probably the easiest approach.

On Fri, Jun 3, 2016 at 8:48 PM, Dan Sully <notifications@github.com> wrote:

> I have a working patch based on @sholsapp <https://github.com/sholsapp>'s
> change, rebased to HEAD & using _openssl_assert().
>
> What's the best way to get that change into this PR? Or do I need a new PR
> by it's nature?
>
> —
> You are receiving this because you are subscribed to this thread.
> Reply to this email directly, view it on GitHub
> <https://github.com/pyca/pyopenssl/pull/281#issuecomment-223725784>, or mute
> the thread
> <https://github.com/notifications/unsubscribe/AAADBFEvyVy7Kl10XVrLb0E8uectnwstks5qIMtvgaJpZM4FJU_P>
> .
>



-- 
GPG Key fingerprint: D1B3 ADC0 E023 8CA6

dsully:
Ok, #483 is good to go, passed Travis.

PR #286, opened 2015-07-02
Add EC support

aalba6675:
Begin to add EC support to PKey and its methods.


hynek:
I believe this has been fixed via #439.  Sorry for the inappropriate process!

PR #287, opened 2015-07-03
Move to codecov, rejigger badges

hynek:
Let’s say goodbye to 4 comments on each commit.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -61,7 +61,7 @@ matrix:
 
   - python: "2.7"
     env: TOXENV=docs
-    
+
   # Let the cryptography master builds fail because they might be triggered by
   # cryptography changes beyond our control.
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -104,8 +104,11 @@ script:
     fi
     ~/.venv/bin/tox
 
+before_install:
+  - pip install codecov
```
alex: This isn't running inside of a venv on OS X. I'd just leave this where the `pip install coveralls` was.

.travis.yml
```diff
@@ -104,8 +104,11 @@ script:
     fi
     ~/.venv/bin/tox
 
+before_install:
+  - ~/.venv/bin/pip install codecov
```
alex: The venv is created in `install`, this is too late. Why don't you just install `codecov` in the same place `coveralls` used to be installed?

PR #288, opened 2015-07-03
Add coverage to pip install

hynek:
Codecov needs it.


alex:
oof, you'll need to activate the venv first so that it can find `coverage` on `$PATH`

PR #289, opened 2015-07-03
Enabling caching of pip downloads and wheels

alex:
Makes stuff more vroom vroom on travis

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,10 @@
 sudo: false
 language: python
+
+cache:
+  directories:
+    - $HOME/.cache/pip
+
 env:
   global:
     - LC_ALL=en_US.UTF-8


hynek:
Are you sure it's working?  The build of this PR was just 20s faster than of mine (20 min 25 sec vs 20 min 4 sec).

reaperhulk:
First time through cache has no impact since it still has to build it to cache it. But it also looks like the docker images Travis is using don't have pip7 yet so no wheels

hynek:
Maybe add a `pip install -U pip setuptools wheel`?

alex:
As always, I prefer to keep orthagonal issues seperate; can we land this as is and do a different patch to upgrade pip if that's what we want?

hynek:
My understanding is that this change is useless without a newer pip?  As in: no caching happens?  Or does it mean that it will cache downloads but no wheels?

If the latter, I’m happy to merge.

alex:
downloads are still cached; we're just missing the wheels bit

On Fri, Jul 3, 2015 at 1:43 PM, Hynek Schlawack <notifications@github.com>
wrote:

> My understanding is that this change is useless without a newer pip? As
> in: no caching happens? Or does it mean that it will cache downloads but no
> wheels?
>
> If the latter, I’m happy to merge.
>
> —
> <https://github.com/pyca/pyopenssl/pull/289#issuecomment-118400029>.
>



-- 

hynek:
*vroom vroom*

PR #292, opened 2015-07-18
Add new function `crypto.pbkdf2_hmac()` that implements key stretching.

ldx:
This adds `crypto.pbkdf2_hmac()` for key stretching using OpenSSL's PKCS5_PBKDF2_HMAC(). It takes the name of the pseudorandom function as a parameter, and can use any of OpenSSL's EVP_*()  message digest functions, such as 'md5', 'sha1', 'sha256' etc.


santiycr:
+1!

alex:
I'm not sure there's a comeplling reason to put this in PyOpenSSL, rather than just telling people to use pyca/cryptography. What do you think @hynek, @reaperhulk ?

hynek:
It’s basically https://cryptography.io/en/latest/hazmat/primitives/key-derivation-functions/?highlight=stretching#cryptography.hazmat.primitives.kdf.pbkdf2.PBKDF2HMAC , right?

In general, I’m -0 on adding APIs to pyOpenSSL that are not TLS-related.  `cryptography` is a better home for them IMHO.

ldx:
@alex @hynek I wasn't aware `PBKDF2HMAC()` was already in cryptography. This works for my original use case, so closing this PR. Thanks!

hynek:
Cool!  Maybe we should add something to the pyOpenSSL docs. *ponder*

Issue #266, opened 2015-05-26
Multiple tests have race conditions


hynek:
They compare outputs and if the data is generate around the boundary of a second, they don’t match because of that.

Currently known perpetrators:
- [ ] `CRLTests.test_export_text`
- [ ] `X509Tests.test_gmtime_adj_notBefore`
- [ ] `test_ssl.ConnectionTests.test_state_string`

alexwlchan:
Duplicate of #102?

alexwlchan:
I’ve been thinking about this a bit, because I’ve been racking up a lot of CI runs lately.

The most “pure”, robust approach would be to modify the test to detect datestamps in the two strings, parse as datetime, get a timedelta object, re-run as necessary… which all seems rather fiddly and complicated for a window condition that usually disappears if we re-run the test.

Alternatively, how would you feel about using [flaky](https://pypi.python.org/pypi/flaky) to re-run those tests a second time if we detect an AssertionError thrown on the first run? That’s easy to do and would probably catch the vast majority of cases.

hynek:
I *think* I’d prefer the pure approach TBH.  Although I’m a bit afraid, the timestamp format might change with OpenSSL versions (just a hunch based on history)?

alex:
`test_export_text` is probably the most frequent offender here. I'm sorely tempted to slap `flaky` on it. How are other folks feeling?

reaperhulk:
I'm 👍 on flaky. Tired of pyOpenSSL failing parts of cryptography's CI with its races...

PR #295, opened 2015-07-25
fix race condition in tests

mhils:
Fix race condition in tests which surfaced here: https://travis-ci.org/pyca/pyopenssl/jobs/72581102#L222

### DIFF
--- a/OpenSSL/test/test_crypto.py
+++ b/OpenSSL/test/test_crypto.py
@@ -1507,9 +1507,11 @@ WpOdIpB8KksUTCzV591Nr1wd
         the current time plus the number of seconds passed in.
         """
         cert = load_certificate(FILETYPE_PEM, self.pemData)
-        now = datetime.utcnow() + timedelta(seconds=100)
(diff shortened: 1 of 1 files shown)


hynek:
Would you mind elaborating on your approach?  That may expedite the review. :)

mhils:
Sure. In a nutshell, the previous code was
````python
now = datetime.utcnow() + timedelta(seconds=100)
cert.gmtime_adj_notBefore(100)
self.assertEqual(cert.get_notBefore(), b(now.strftime("%Y%m%d%H%M%SZ")))
```
This fails if there is the second changes between the first and the second statement, as in the travis job above. The workaround is to capture a timestamp before and after the call to `cert.gmtime_adj_notBefore(100)` and make sure that `cert.get_notBefore()` is within that range.
Does this clear things up? :smiley:

hynek:
Makes sense thanks.

hynek:
You also may be interested in #266. ;)

PR #296, opened 2015-08-04
Refactor and fix X509 certificate serial_number functions

socketpair:
And also allow to pass raw bytes to X509.set_serial_number()

### DIFF
--- a/src/OpenSSL/crypto.py
+++ b/src/OpenSSL/crypto.py
@@ -1127,34 +1127,36 @@ class X509(object):
         if not isinstance(serial, _integer_types):
             raise TypeError("serial must be an integer")
 
-        hex_serial = hex(serial)[2:]
-        if not isinstance(hex_serial, bytes):
-            hex_serial = hex_serial.encode('ascii')
+        if serial < 0:
+            raise ValueError('Serial number should not be negative')
 
-        bignum_serial = _ffi.new("BIGNUM**")
-
-        # BN_hex2bn stores the result in &bignum.  Unless it doesn't feel like
-        # it.  If bignum is still NULL after this call, then the return value
-        # is actually the result.  I hope.  -exarkun
-        small_serial = _lib.BN_hex2bn(bignum_serial, hex_serial)
-
-        if bignum_serial[0] == _ffi.NULL:
-            set_result = _lib.ASN1_INTEGER_set(
-                _lib.X509_get_serialNumber(self._x509), small_serial)
-            if set_result:
-                # TODO Not tested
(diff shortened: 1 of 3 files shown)


socketpair:
Unfortunatelly I cannot run tests for python2 since `cryptography` can not be installed on my system...

StyleEX:
+1

hynek:
How would you feel about adding a helper function that you’d use together with `set_serial_number_hex`?  It’s easy to test and generally speaking dispatch on types is a rather bad idea.

socketpair:
Not sure I understand your right. Please be verbose :(

hynek:
Right, because I made a mistake. :)

Another attempt to get it right: 

Instead of adding a new method and changing the existing `set_serial_number`, I’m arguing for something like: `x509.set_serial_number(hexstr_to_int(b"fffe"))` which doesn’t change the existing class and much simplifies the testing of the added code.

It’s more explicit in its intent and doesn’t hide possible bugs.

socketpair:
Thinking that serial number is monotonically incrementing value is a bad idea. Actually it is not used in secure environment since  https://en.wikipedia.org/wiki/Predictable_serial_number_attack . In real scenarios, serial number is a random binary string of arbitrary length (someone use 8 bytes, someone 64).

Example of such vulnerability:
http://blog.erratasec.com/2008/12/not-all-md5-certs-are-vulnerable.html

So, it is much better to create new function (accepting raw bytes), and consider function accepting plain integer obsolete. And more over, some users who use obsolete technologies to generate serial numbers should call `set_serial_number_hex(hexlify(42))`.

So, for the first time ( 1-2 years?) it will be nice to support deprecated function, but with this:
https://docs.python.org/2/library/warnings.html

And also, it may be suitable to add `.set_random_serial_number()`, but it is a question about which random we should use....so it is not good API.

socketpair:
For example

https://reborn.ru
serial number is  26:41:8d:fd:83:aa:85:a5:af:86:8f:64:45:75:a8:d9


So, I think, it is reasonable to switch to new API with some time. So, I added new function, and consider to make old API to be obsolete.

socketpair:
Re-worked branch completely. Will not change API, but remove VERY WRONG code about small serial numbers. BN_hex2bn NEVER return original value, even if it was small.

socketpair:
I'm tired to find and fix previous bugs, especially when tests are FAILED. I cannot fix bugs since code coverage find things that I did not change (just change indentation because code refactoring).

This fuck me hard, sorry. If you can't merge that, just say NO, and I will decline that pul request.

PR #303, opened 2015-08-27
Fixed AttributeError when calling get_app_data()

chakatodd:


alex:
Can you add a test case for this?

PR #304, opened 2015-08-27
Fixed AttributeError when calling get_app_data()

chakatodd:


OpenSSL/test/test_ssl.py
```diff
@@ -2376,6 +2376,7 @@ def test_app_data(self):
         :py:obj:`Connection.get_app_data`.
         """
         conn = Connection(Context(TLSv1_METHOD), None)
+        self.assertIsNone(conn.get_app_data())
```
kjav: This should be self.assertIs(None, conn.get_app_data())

chakatodd: I'm relatively new to python. Is this because of the minimum supported python version? If so where would I find that?

chakatodd:
Updated with requested change.

OpenSSL/test/test_ssl.py
kjav: I'm not sure, sorry - I just saw there was a build failure and looked elsewhere in the file for an example of asserting for values equal to None.

hynek:
Um this kind of got lost.  Could you rebase please?

We’ve also updated our https://github.com/pyca/pyopenssl/blob/master/CONTRIBUTING.rst file, it would be nice if you could make this patch conform to it (mostly the assert as far as I see.  please convert the others in the same test method too).

Sorry for the delays.

chakatodd:
I don't understand why the build is failing.

hynek:
Most failures are due to a newer version of flake8 I think and were fixed in #409.

However your code has one problem that flake8 is complaining about too:  the equivalent of `assertIdentical` is not `==` but `is`.

So rebasing + fixing that line should make everything green.

Thanks!

hynek:
Thank you and sorry it took so long.

Issue #142, opened 2014-07-17
renegotiate method on SSL.Connection silently does nothing


alex:
https://github.com/pyca/pyopenssl/blob/master/OpenSSL/SSL.py#L1091

kjav:
How can I perform a renegotiation? I've looked at the _lib object and cannot find any relevant functions, apart from the SSL_get_secure_renegotiation_support function. I'd like to close this issue.

PR #305, opened 2015-08-28
Implements previously empty renegotiate & renegotiate_pending functions

kjav:
This fixes bug https://github.com/pyca/pyopenssl/issues/142, and uses bindings committed in https://github.com/pyca/cryptography/pull/2303

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -65,7 +65,7 @@ source_suffix = '.rst'
 # The encoding of source files.
 #source_encoding = 'utf-8-sig'
 
-# The master doctree document.
+# The master toctree document.
 master_doc = 'index'
 
 # General information about the project.
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -65,7 +65,7 @@ source_suffix = '.rst'
 # The encoding of source files.
 #source_encoding = 'utf-8-sig'
 
-# The master toctree document.
+# The master doctree document.
 master_doc = 'index'
 
 # General information about the project.
--- a/OpenSSL/test/test_ssl.py
+++ b/OpenSSL/test/test_ssl.py
@@ -3210,8 +3210,8 @@ class ConnectionRenegotiateTests(TestCase, _LoopbackMixin):
         """
         server, client = self._loopback()
 
-        server.send("hello world")
-        self.assertEquals(client.recv(len("hello world")), "hello world")
+        server.send(b"hello world")
+        self.assertEquals(client.recv(len(b"hello world")), b"hello world")
 
         self.assertEquals(server.total_renegotiations(), 0)
         self.assertTrue(server.renegotiate())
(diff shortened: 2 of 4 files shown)


kjav:
How do I get the checks to be run again? The most recent version of the cryptography bindings were not merged until the checks had already failed.

kjav:
Would committing an empty commit work?

alex:
Yup, an empty commit will work, one of the core developers can also reset
travis.

On Fri, Aug 28, 2015 at 12:52 PM, kjav <notifications@github.com> wrote:

> Would committing an empty commit work?
>
> —
> <https://github.com/pyca/pyopenssl/pull/305#issuecomment-135830897>.
>



-- 

kjav:
Don't worry - I managed to find a typo in a comment, so I commited a fix for that.

doc/conf.py
```diff
@@ -65,7 +65,7 @@ def find_version(*file_paths):
 # The encoding of source files.
 #source_encoding = 'utf-8-sig'
 
-# The master toctree document.
+# The master doctree document.
```
alex: I don't think this is actually correct -- I believe this comment indicates that "index" is the root in teh table of contents.

alex:
This will need some tests before it can be landed. Thanks for working on it!

kjav:
Why is codecov 'waiting for Cl to finish' when it has finished already?

kjav:
I think a test has already been written for this function, but commented out. https://github.com/kjav/pyopenssl/blob/master/OpenSSL/test/test_ssl.py#L3208

kjav:
Why is this build failing? It says the function "SSL_renegotiate_pending" is not a member of the _lib module, but it was added in https://github.com/pyca/cryptography/commit/c6a73cb16f00889b53d368359059141d75a26be0

kjav:
Perhaps the cryptography module needs to be rebuild on the build machine (using `pip install -e .` in the root directory)?

reaperhulk:
Because it's not in a released version. The cryptographyMaster builds (which pull from github) are passing.

I've backported your binding into our 1.0.x tree but until we do the 1.0.1 release this won't pass (and then you'll need to update the PR to require cryptography 1.0.1 in PyOpenSSL anyway).

kjav:
Ah, that makes sense. Would I do that using an #ifdef statement/?

reaperhulk:
No, it will be the version restriction in setup.py. Pure Python. But you shouldn't update this PR again until 1.0.1 is out since that's the blocker now :)

kjav:
ok, I'll remember to do that :)

kjav:
Looks like the latest release of cryptography (1.0.1) has my new bindings in it! Is it acceptable to increase the cryptography requirement from 0.7 to 1.0.1?

reaperhulk:
I'm fine with that, but cryptography 1.0+ made some big installation changes (requiring cffi 1.1+) so we should be aware that some users may have some trouble during the transition.

hynek:
I guess that move is inevitable.  We’ll just have to wait until Travis supports an appropriate PyPy to avoid another huge influx of angry bug reports.

hynek:
It appears that we’re already depending on cryptography 1.0 so if you polish it up we can merge this.

hynek:
@kjav are you still interested in working on this?

PR #307, opened 2015-08-31
download, compile, and use a local 0.9.8 in travis when required

reaperhulk:
This is experimental and will likely fail the first few (million) tries.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,8 @@ language: python
 
 cache:
   directories:
-    - $HOME/.cache/pip
+#    - $HOME/.cache/pip
+    - $HOME/ossl-098
 
 env:
   global:
@@ -44,12 +45,6 @@ matrix:
   - python: "2.7"
     env:
       OPENSSL=0.9.8 TOXENV=py27
-    addons:
-      apt:
-        sources:
-          - lucid
-        packages:
-          - libssl-dev/lucid
 
   # Meta
   - python: "2.7"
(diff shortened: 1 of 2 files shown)


alex:
There's a syntax error in the bash.

.travis.yml
```diff
@@ -110,6 +105,22 @@ script:
       export CFLAGS="-I/usr/local/opt/openssl/include"
       export PATH="/usr/local/opt/openssl/bin:$PATH"
     fi
+    # This section potentially downloads, compiles and installs openssl 0.9.8zg
+    if [[ "${OPENSSL}" == "0.9.8"]]; then
+      # download, compile, and install if it's not already present via travis cache
+      if [ ! -f "$HOME/ossl-098/bin/openssl" ]; then
```
alex: Missing a `[` and `]` here.

hynek:
Looks great! `OpenSSL 0.9.8zg 11 Jun 2015`

If I’m not mistaken, just remove it from allowed failures and we’re good to go?

reaperhulk:
This isn't actually behaving itself yet. You see 0.9.8zg but then it's linked against 0.9.8o. I think that might be libssl0.9.8 from Ubuntu so I probably need a version spec for ld

reaperhulk:
Got it loading the right library now. Unfortunately a unit test fails right now. Without looking into it I'm guessing it might have something to do with trust roots...

hynek:
I’m willing to say to call it a day on this ticket and solve the new problem in a new one.

Might be that 0.9.8 passes on OS X just because of my favorite CVE.

What do y’all think?

reaperhulk:
There's a good chance that's true. I'm neutral on merging it (mostly because I don't have the ambition to sort out this failure right now).

PR #308, opened 2015-09-01
Add support for Elliptic Curve Pkey

r0ro:
This is largely inspired by https://github.com/pyca/pyopenssl/pull/195/ but with a different approach regarding generate_key arguments.
Instead of keeping the 'bits' argument, I replaced it with a 'param' argument.
When called, generate_key will check the 'param' value according to the Pkey 'type'.
When 'type' is RSA or DSA, 'param' must be an int value specifying the key bits length.
When 'type' is EC, 'param' must be the name of an elliptic curve.


r0ro:
I'm not sure how I'm supposed to improve coverage, since the untested code cannot be reached if openssl is compiled with ec support. Any ideas ?

hynek:
So my idea would be to not use `_lib.Cryptography_HAS_EC` in the methods for determining EC support, but looking at `TYPE_EC` instead.

That way you can easily monkey patch the module before running tests.

***

Please note though, that you can’t rename the `bits` parameter as that would break our backwards compatibility (plenty people use always kwargs).  And re-using parameters is poor style anyway.

I personally would tend to suggest to deprecate `generate_key` and create three new methods called `generate_(rsa|dsa|ec)_key` with unambiguous parameters.

Opinions @reaperhulk @alex?

reaperhulk:
I'd prefer the separated key generation functions as well.

r0ro:
Ok, I'll modify this asap

r0ro:
I've added the `generate_(rsa|dsa|ec)_key` functions, updated the tests using the new assert format.
The only thing I can't improve is the coverage. 
The unchecked code comes from unreachable _raise_current_error() and I don't see any way to trigger those errors on purpose.

r0ro:
any chance it gets merged ?

hynek:
just a few things that spring to my eye:

- you need to rebase
- if we deprecate generate_key, it als has to raise a warning (and needs a test for it) – ou can draw inspiration from https://github.com/pyca/pyopenssl/commit/80d005f68754f268da419540a593db83286f3d0c
- please get rid of the `:py:` prefixes.  Also `:py:data:` doesn’t really make sense for arguments.

r0ro:
I've rebased to latest version.
Note that I had to force the use of pytest 2.8.2 in tox.ini since the current release is broken (see https://github.com/pytest-dev/pytest/issues/1238)

Also the build fails on osx with system openssl, it seems to be a version mismatch somewhere :
```
/Users/travis/build/pyca/pyopenssl$ /usr/bin/openssl version 
OpenSSL 0.9.8za 5 Jun 2014

print(OpenSSL.SSL.SSLeay_version(OpenSSL.SSL.SSLEAY_VERSION)) 
OpenSSL 1.0.2d 9 Jul 2015
```

It also affects other pull requests, so I guess the current master won't pass the test

It's my final attempt to get this merged. If there's only cosmetic changes, I'll let you handle it.

hynek:
The failure on OS X/System was fallout from pyca/cryptography#2503 before but now it’s indeed failing due to a version mismatch.  I’ve created #397 for that.

I’m sorry the contribution process is frustrating to you especially since the main effort we put since April was to make the project friendlier to contributions.

But there’s only so much we can do on our unpaid free time on a security critical and 15 years old project.  Usually we get yelled at for moving both too slow and too fast by two different sets of people.  Additionally we get regularly stopped in tracks by the OpenSSL project’s regressions or some distribution like Fedora makes our life extra miserable like they did the other day.

If you have a solution to this situation I’ll be glad to hear it, until then I’m afraid we’ll all have to live in misery together.

r0ro:
hynek, sorry if I was disrespectful, I was indeed frustrated because my original pull request is now 3 month old, and I never got any feedback before asking for it.
Thanks for your feedback on OS X build failure.

hynek:
I did not interpret as disrespectful, I just wanted to acknowledge your frustration and add context.

atusingh:
@r0ro Does it make sense to support explicit paramater encoding when generating ec key? Maybe adding a arg in `generate_ec_key` (`explicit=False` or `param_encoding="named_curve"`) then passing it to `_to_EC_KEY`?

If compatibility with 15.1 needs to be maintained, the new arg could be defaulted to `"explicit"` encoding. But openssl's default is `"named_curve"`.

hynek:

PR #310, opened 2015-09-02
getattr now works when no socket is wrapped

kjav:
When using a Connection object with no socket, _sock is None and the connection throws unhelpful exceptions when an attribute exception should be thrown. This change fixes this issue.

### DIFF
--- a/OpenSSL/SSL.py
+++ b/OpenSSL/SSL.py
@@ -1146,7 +1146,10 @@ class Connection(object):
         Look up attributes on the wrapped socket object if they are not found on
         the Connection object.
         """
-        return getattr(self._socket, name)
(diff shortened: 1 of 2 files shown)


kjav:
Anyone have any comments?

hynek:
I’m not sure about the error message.  Wouldn’t it be more helpful to tell the user that no socket is wrapped yet?

kjav:
You can use the libraries without wrapping a socket at all, by using the bio_write and bio_read functions.

kjav:
I've found this useful when using PyOpenSSL to test a proprietary OpenSSL library

hynek:
oh I see now, thanks.

PR #313, opened 2015-09-04
get_elliptic_curve require 'name' parameter

zakkie:


hynek:
Thanks!  Could you surround the word `name` in `Return a single curve object selected by name.` with `*`s?

alex:
@hynek I'm going to merge this, since it corrects a bug. I'll send a follow up PR to add the `*`.

PR #323, opened 2015-09-04
Fixed #322 -- removed code that already exists in cryptography

### LINKED ISSUE #322: Delete locking callback code, just use cryptography's

alex:


hynek:
Do we have any thread safety tests that can make us slightly confident this isn’t gonna break stuff?

alex:
@hynek There's a couple of files in `leakcheck/` which you can run manually that use threads.

hynek:
This is a bit too hairy for me. Can @reaperhulk have a look maybe?

reaperhulk:
Two comments:

* I think hoisting the `init_static_locks` call up into _util.py right after the binding is initialized would be a more logical place for it to live.
* the cryptography implementation does not set `CRYPTO_set_id_callback`, so we may need to investigate what that really means. https://www.openssl.org/docs/manmaster/crypto/threads.html mentions that `CRYPTO_THREADID` deprecated `CRYPTO_set_id_callback` in 1.0.0+, but it's not clear to me what the consequence of not setting this callback is.

@public, you originally wrote the locking callback for cryptography, do you know if/why `CRYPTO_set_id_callback` (or the `CRYPTO_THREADID_set_callback`) are needed?

alex:
@reaperhulk as far as I can tell from the openssl docs, `set_id_callback` and friends are totally deprecated, I don't think they do anything.

reaperhulk:
`CRYPTO_THREADID_set_callback` is deprecated? It is the successor to `CRYPTO_set_id_callback` right?

alex:
@reaperhulk sorry, it's not deprecated, but there's a default behavior which is perfectly fine, we're not adding any value by setting this callback.

reaperhulk:
Ah, perfect. Then I'm :+1: on this once tests pass and will merge.

PR #324, opened 2015-09-04
Get rid of flaky shields.io badge

hynek:
It’s been error’ing for weeks and months now.  No real use for it either.


alex:
Is there something we coudl replace this with?

hynek:
I don’t find it particularly useful (and it already was a replacement for another flaky badge provider).  How about just a link to PyPI?

PR #334, opened 2015-09-05
add OP_SINGLE_ECDH_USE

zakkie:
to improve forward secrecy...


OpenSSL/SSL.py
```diff
@@ -65,6 +65,7 @@ class _buffer(object):
     pass
 
 OP_SINGLE_DH_USE = _lib.SSL_OP_SINGLE_DH_USE
+OP_SINGLE_ECDH_USE = SSL_OP_SINGLE_ECDH_USE
```
alex: This needs `_lib.` before it.

Also, I suspect this attribute doesn't always exist, so it might need a conditional.

zakkie: oops, sorry. fixed it.

`SSL_OP_SINGLE_ECDH_USE` was introduced at same time as SSL_CTX_set_tmp_ecdh: openssl/openssl@ea262260469e49149cb10b25a87dfd6ad3fbb4ba
So, I think conditionally catching exception is unnecessary

alex: Ok, sounds good.

Can you merge master into this branch so that it's up to date and then we
can merge.

On Sat, Sep 5, 2015 at 11:26 AM, Akihiro YAMAZAKI <notifications@github.com>
wrote:

> In OpenSSL/SSL.py
> <https://github.com/pyca/pyopenssl/pull/334#discussion_r38811783>:
>
> > @@ -65,6 +65,7 @@ class _buffer(object):
> >      pass
> >
> >  OP_SINGLE_DH_USE = _lib.SSL_OP_SINGLE_DH_USE
> > +OP_SINGLE_ECDH_USE = SSL_OP_SINGLE_ECDH_USE
>
> oops, sorry. fixed it.
>
> SSL_OP_SINGLE_ECDH_USE was introduced at same time as
> SSL_CTX_set_tmp_ecdh: openssl/openssl@ea26226
> <https://github.com/openssl/openssl/commit/ea262260469e49149cb10b25a87dfd6ad3fbb4ba>
> So, I think conditionally catching exception is unnecessary
>
> —
> <https://github.com/pyca/pyopenssl/pull/334/files#r38811783>.
>



-- 

hynek:
Also the doc string doesn’t quite fit anymore.  Could change it to `Constants` and ` When these options re used`?

OpenSSL/SSL.py
zakkie: sure, merged master

PR #339, opened 2015-09-05
Make flake8 happy about test/util.py

hynek:


OpenSSL/test/util.py
```diff
@@ -119,7 +121,8 @@ def format_leak(p):
 
             # Figure the last interesting frame will always be CRYPTO_malloc,
             # since that's where we hooked in to things.
-            while c_stack and 'CRYPTO_malloc' not in c_stack[0] and 'CRYPTO_realloc' not in c_stack[0]:
+            while c_stack and 'CRYPTO_malloc' not in c_stack[0] and \
+                    'CRYPTO_realloc' not in c_stack[0]:
```
alex: Can you use parenthesis here, we've avoided using `\` line continuations everywhere else.

OpenSSL/test/util.py
```diff
@@ -105,7 +106,8 @@ def format_leak(p):
             # /home/exarkun/Projects/cpython/2.7/python(PyEval_EvalFrameEx+0x753b) [0x4d0e1e]
             # /home/exarkun/Projects/cpython/2.7/python() [0x4d6419]
             # ...
-            #
+            """  # noqa
```
alex: Ugh, this makes me super sad :-(

hynek: yeah me too but that seems the best way to silence it selectively?

alex: Can we remove the quotes from this section, leave it failing, and come back to it at the end, I want some time to think about the best approach here.

PR #354, opened 2015-09-06
Skip test_set_default_verify_paths on own 0.9.8

hynek:
I had a closer look at the new fail on 0.9.8 and in the light of limited visibility within Travis it seems clear to me that the problem is that our self-compiled OpenSSL doesn't know about the system's trust store.

There's probably some compile option to make it use it?  Either that, of we can just skip the test as done in this PR.  I don't really care either way and leave it to @reaperhulk who knows much more, how much effort it would be to smarten up our OpenSSl.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,10 +6,6 @@ cache:
 #    - $HOME/.cache/pip
     - $HOME/ossl-098
 
-env:
-  global:
-    - LC_ALL=en_US.UTF-8
-
 matrix:
   include:
   - language: generic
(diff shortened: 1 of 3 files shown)


hynek:
Paaauuul?  Did you play with the compiler settings yet? :)

reaperhulk:
I did but didn't come to any decent resolution on this thus far. We should probably just skip for now.

hynek:
OK I’m confused, why does https://travis-ci.org/pyca/pyopenssl/jobs/84618334 fail?

PR #358, opened 2015-09-07
Implemented state_string function

kjav:


kjav:
This is the last function in the Connection class that is not yet implemented. There are still functions in the Context class which are not implemented, but do not throw NotImplemented exceptions (e.g. load_client_ca)

alex:
Looks ok to me, @hynek can you confirm?

hynek:
Needs a changelog entry, otherwise seems ok.

kjav:
Is the format of that entry ok?

ChangeLog
```diff
@@ -1,3 +1,9 @@
+2015-09-07 Sam Lee <sam.rr.lee@gmail.com>
+
+	* OpenSSL/SSL.py, OpenSSL/test/test_ssl.py: Implemented
+	  the empty ``Connection.state_string()`` function, using
```
alex: Can you remove the word "empty", I think it's confusing unless you know the full history of this bug.

kjav:
OK, that makes sense. I have removed it.

hynek:
thanks!

PR #359, opened 2015-09-07
add Context.set_ecdh_auto()

zakkie:
for using ECDH easily


reaperhulk:
This branch currently has a conflict so it will need to be rebased. OpenSSL 1.1.0 (which is in alpha now) also removes this function in favor of (essentially) enabling it by default. If a user wants to disable ECDH they can do so by excluding it in the cipher lists. Is there any reason we shouldn't just do the same (e.g. enable this by default and not add a new API)?

Lukasa:
I think we should do that @reaperhulk, :+1:.

hynek:
:+1:

PR #368, opened 2015-10-13
factor out function to dump crl

ddcc:


OpenSSL/crypto.py
```diff
@@ -2576,6 +2560,30 @@ def verify(cert, signature, data, digest):
...
+        ret = _lib.i2d_X509_CRL_bio(bio, crl._crl)
+    elif type == FILETYPE_TEXT:
+        ret = _lib.X509_CRL_print(bio, crl._crl)
+    else:
+        raise ValueError(
+            "type argument must be FILETYPE_PEM, FILETYPE_ASN1, or "
+            "FILETYPE_TEXT")
+
+    assert ret == 1
+    return _bio_to_string(bio)
```
reaperhulk: Two blank lines here for PEP8.

OpenSSL/crypto.py
```diff
@@ -2576,6 +2560,30 @@ def verify(cert, signature, data, digest):
     if verify_result != 1:
         _raise_current_error()
```
reaperhulk: Two blank lines here for PEP8.

reaperhulk:
So what's the goal of refactoring this out? Is it the intent to have `dump_crl` be part of the public API? If so, this needs docs (and a changelog entry).

Otherwise the core code looks fine to me, but I don't have any opinions on whether this should be a public API or not.

ddcc:
Yeah, I'd like to add it as part of the API, since most other objects support both loading and dumping. For my use case, I'd also like to be able to export CRL's that are within PKCS#7 data.

OpenSSL/test/test_crypto.py
```diff
@@ -3205,6 +3205,14 @@ def test_load_crl_bad_data(self):
         """
         self.assertRaises(Error, load_crl, FILETYPE_PEM, b"hello, world")
 
+    def test_dump_crl(self):
+        """
+        Dump a known CRL and ensure it is output correctly.
+        """
+        crl = load_crl(FILETYPE_PEM, crlData)
+        buf = dump_crl(FILETYPE_PEM, crl)
+        self.assertEqual(crlData, buf)
```
hynek: please `assert crlData == buf`

doc/api/crypto.rst
```diff
@@ -90,6 +90,11 @@ Private keys
 Certificate revocation lists
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
+.. py:function:: dump_crl(type, crl)
```
hynek: please use autodoc here.  currently the docstring is more verbose than the documentation. :)

ddcc:
@hynek Ok, I think I have made the changes you mentioned. Never used autodoc before, so let me know if it needs to be changed.

hynek:
Looks good

just:

- add respective data types (buffer is not a data type…) using Sphinx directives
- surround the constants with double ` to make the fixed width
- the Returns string doesn’t have to be that verbose
- finish sentences with a period. :)

bonus points if you do it also to the remaining three CRL-related functions (completely optional).

thanks!

OpenSSL/test/test_crypto.py
```diff
@@ -3205,6 +3205,14 @@ def test_load_crl_bad_data(self):
         """
         self.assertRaises(Error, load_crl, FILETYPE_PEM, b"hello, world")
 
+    def test_dump_crl(self):
+        """
+        Dump a known CRL and ensure it is output correctly.
```
hynek: https://jml.io/pages/test-docstrings.html Please

hynek:
(please note that I didn’t see that you’ve added commits; please bump the ticket once you feel that your branch is ready for review)

Getting closer!

- please rebase
- `Check that the dumped CRL matches the original input.` should be just `The dumped CRL matches the original input.`

ddcc:
Ok, done.

hynek:
The failures are unrelated. thanks!

PR #371, opened 2015-10-16
Move package into src

hynek:
So, I don’t want to push this through so it’s more of an RFC and I would like it only merged if both @alex and @reaperhulk agree.

We have a lot of pain with heisenbugs and packaging weirdnesses so this PR cleans the structure up to PyCA standards:

- it moves the package into a src directory so it’s not accidentally imported when running tests (and the `pip install -e .` ’ed version is used)
- it moves the test directory out of the package because I find that terrible style with nothing but downsides (you can perfectly run separate tests in production; you even get the assurance that they are testing code within site-packages)
- I have *no* idea how to use `runtests.py` since it only fails for me.  I feel like we should get rid of it, but how does one do memory debugging?  But that’s for another day.
- These changes also enable us to collect coverage over the whole tox run so you get your local codecov everyone always dreamed of.

Let me know what you think.


setup.py
```diff
@@ -46,82 +44,45 @@ def find_meta(meta):
...
-    ],
-    test_suite="OpenSSL",
-    tests_require=[
-        "pytest",
-    ],
-    cmdclass={
-        "test": PyTest,
-    }
-)
+if __name__ == "__main__":
```
reaperhulk: Speaking as someone ignorant of the true mysteries of setup.py, what's the advantage of doing this?

reaperhulk:
This change appears to break using pytest for `python setup.py test`. I don't necessarily object to that, although we maintain compatibility for that in cryptography because some packagers seem to think that's the only way to run tests.

hynek:
I don't like/use `setup.py test` because it doesn't use pip and it's terrible all around. If we don't use it ourselves it's pointless to keep.

NB: we haven't released a version of pyOpenSSL  with setup:py test support.

setup.py
hynek: Nothing in particular.  I just like to be able to import it for testing around.

hynek:
Now I’ve added @Lukasa to the committer team (will write an e-mail about that soonish), I’d like to weight him in too.

But seriously,I’m asking about opinions whether to go forward, not a code review.  We can bike shed later.  I just want feedback now *poke* @alex .

Lukasa:
I'm +1 on this change. I don't care about `python setup.py test`, particularly as in all my testing all it managed to do was segfault my interpreter.

hynek:
So Alex expressed a lack of opinions on IRC.   So one of you gentlemen (or @lvh) please review or merge, thanks. :)

reaperhulk:
I'm happy with this -- we can make tweaks as required in future PRs.

PR #374, opened 2015-10-19
Add a twisted target to CI

hynek:
Twisted is our most important downstream user.  In order to avoid another
regression release like 0.15.1, we add it to our CI

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -47,7 +47,7 @@ matrix:
 
   # Make sure we don't break Twisted
   - python: "2.7"
-    env: TOXENV=py27-twisted
+    env: TOXENV=py27-twistedMaster
 
 
   # Also run at least a little bit against an older version of OpenSSL.
--- a/.travis.yml
+++ b/.travis.yml
@@ -45,6 +45,11 @@ matrix:
   - python: "pypy"
     env: TOXENV=pypy-cryptographyMaster
 
+  # Make sure we don't break Twisted
+  - python: "2.7"
+    env: TOXENV=py27-twisted
+
+
   # Also run at least a little bit against an older version of OpenSSL.
   - python: "2.7"
     env:
--- a/tox.ini
+++ b/tox.ini
@@ -23,10 +23,7 @@ commands =
 deps =
     # [tls,conch] syntax doesn't work here so we enumerate all dependencies.
     git+https://github.com/twisted/twisted
-    gmpy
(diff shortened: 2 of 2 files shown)


Lukasa:
> Twisted is our most important downstream user.

As a Requests core maintainer, I'll try not to be too offended by that statement. :wink:

More seriously, is it sensible to test *all* of Twisted here? That'll take a while. Can we restrict the tests we run?

reaperhulk:
Good error too:

```exceptions.ValueError: invalid literal for int() with base 10: 'dev0'```

:dancer:

**Comment by hynek** [comment]
Yeah, it fails because Twisted makes assumptions about our versions. :(((

But it would kind of suck if we found out about breakage not before I’ll try to release, huh?

**Comment by hynek** [comment]
```python
    major, minor = list(int(part) for part in lib.__version__.split("."))[:2]
```


pretty sure the fix is to just move the `[:2]` into the parentheses…

hynek:
Working on it under http://twistedmatrix.com/trac/ticket/8056

hynek:
FYI it landed on Twisted trunk.

So we have two options:

1. test against Twisted trunk
1. wait for the next release

Option 1 seems nice for me for practical reasons: we’ll also have a feedback loop for twisted if they break something.

hynek:
Moar fail is being had:

```
[FAIL]
  File "/home/travis/build/hynek/pyopenssl/.tox/py27-twistedMaster/lib/python2.7/site-packages/twisted/test/test_sslverify.py", line 2285, in test_oneOrTwoDots
    self.assertIn(self.OpenSSL.__version__.count("."), (1, 2))
  File "/home/travis/build/hynek/pyopenssl/.tox/py27-twistedMaster/lib/python2.7/site-packages/twisted/trial/_synctest.py", line 497, in assertIn
    % (containee, container))
twisted.trial.unittest.FailTest: 3 not in (1, 2)

twisted.test.test_sslverify.RealOpenSSLTests.test_oneOrTwoDots
```

Fixing at https://twistedmatrix.com/trac/ticket/8060

Lukasa:
LGTM.

hynek:
you gotta click dat button Cory =)

Lukasa:
I keep forgetting I have the commit bit now.

Y'all are going to regret this *so much*. :shipit: :cat2: :tropical_fish:

Issue #191, opened 2015-02-12
Figure out how to cope with EGD going away


public:
It's likely that EGD support will slowly start to disappear from things that PyOpenSSL depends on in the future. e.g. The libcrypto that it links to on some platforms. pyca/cryptography is considering removing the bindings because of this https://github.com/pyca/cryptography/pull/1636 but obviously breaking PyOpenSSL would be a bad thing.

How do we go about deprecating these APIs in a way that causes the minimum of problems for people?

hynek:
FWIW, it seems that LibreSSL has already cut out EGD?  I tried to do some work in s_i and pre 0.14 pyOpenSSL doesn't install du to compile errors mentioning EDG. 


Can someone give me a short summary on what this is about?

reaperhulk:
Yep, Libre has cut out EGD. The entropy gathering daemon is an outdated and no longer useful daemon that tried to provide entropy by monitoring various sources independently of the kernel.

hynek:
OK I’ve had a look and it seems we use EGD in exactly one place: https://github.com/pyca/pyopenssl/blob/master/src/OpenSSL/rand.py#L94-L113

Suggestion:

1. we make `OpenSSL.rand.egd()` raise a `DeprecationWarning`
2. we replace the call RAND_egd_bytes() by a call of `seed(os.urandom(bytes))`

That way we:

1. don’t compromise the user’s security (arguably we increase it)
2. get rid of the API binding completely
3. start a deprecation cycle so we can get rid of it in a year completely

Opinions?

Lukasa:
Sounds reasonable to me. =)

reaperhulk:
I'm fine with that. Plus it gives us a potential date to remove the binding from cryptography.

PR #375, opened 2015-10-20
Replace the only EGD call by urandom

hynek:
Also harmonize documentation.  #boyScout

### DIFF
--- a/doc/api/rand.rst
+++ b/doc/api/rand.rst
@@ -1,79 +1,43 @@
 .. _openssl-rand:
 
-:py:mod:`rand` --- An interface to the OpenSSL pseudo random number generator
-=============================================================================
+:mod:`rand` --- An interface to the OpenSSL pseudo random number generator
+==========================================================================
+
+.. warning::
+   Functions from this methods shouldn't be used.
+   `Use urandom <http://sockpuppet.org/blog/2014/02/25/safely-generate-random-numbers/>`_ instead.
+
 
 .. py:module:: OpenSSL.rand
    :synopsis: An interface to the OpenSSL pseudo random number generator
 
 
-This module handles the OpenSSL pseudo random number generator (PRNG) and
-declares the following:
+This module handles the OpenSSL pseudo random number generator (PRNG) and declares the following:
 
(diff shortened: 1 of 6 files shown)


hynek:
Fixes #191

doc/api/rand.rst
```diff
@@ -1,79 +1,43 @@
...
-.. py:module:: OpenSSL.rand
-   :synopsis: An interface to the OpenSSL pseudo random number generator
-
-
-This module handles the OpenSSL pseudo random number generator (PRNG) and
-declares the following:
-
-.. py:function:: add(string, entropy)
+.. warning::
+   Functions from this methods shouldn't be used.
```
reaperhulk: s/methods/module right?

doc/api/rand.rst
```diff
@@ -1,79 +1,43 @@
...
+   `Use urandom <http://sockpuppet.org/blog/2014/02/25/safely-generate-random-numbers/>`_ instead.
```
Lukasa: Is it worth adding to the documentation what the deprecation cycle is?

src/OpenSSL/version.py
```diff
@@ -11,7 +11,7 @@
     "__title__", "__uri__", "__version__",
 ]
 
-__version__ = "0.16.dev0"
+__version__ = "15.2.0.dev0"
```
Lukasa: =O Version scheme changes!

Lukasa:
Some small notes (suggestion for more docs, notes on the version number change), otherwise I'm happy. =)

doc/api/rand.rst
hynek: We're not deprecating the whole module, that's just good-natured advice. :)

I wanna write a backwards compat doc before release (2 year after deprecation it goes to hell).

src/OpenSSL/version.py
hynek: Yep, I've discussed that on #cryptography-dev.   SemVer just doesn't make any sense for pyOpenSSL.

Lukasa:
\o/

PR #376, opened 2015-10-20
Fix codecov

hynek:
I’m afraid we’ve (I) broken codecov.  Let’s fix it again.

pyroma is btw gone because it couldn’t handle our structure.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -61,9 +61,6 @@ matrix:
     env: TOXENV=flake8
 
   - python: "2.7"
-    env: TOXENV=pyroma
-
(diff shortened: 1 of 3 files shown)


hynek:
Dunno why the coverage went down.  I think that might be a consequence of the refactoring.  Before the coverage was diluted by the test files which isn’t the case anymore.

IOW this is our *actual* coverage.

reaperhulk:
I like having coverage on the test files too since sometimes you end up with conditionals that aren't hit in tests, but we can discuss that in a separate issue. I can merge this now.

PR #380, opened 2015-10-24
Add py.test settings

hynek:
We add _trial_temp to ignores because now we run the Twisted test suite as
part of tox, py.test tries to recurse into it.

We also require py.test 2.8.2 because older version don't properly handle
deprecated_call which we use.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,3 +1,8 @@
+[pytest]
+minversion = 2.8.2
+strict = true
+norecursedirs = build dist .* *.egg doc _trial_temp
+
 [sdist]
 force_manifest=1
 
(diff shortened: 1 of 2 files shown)


setup.cfg
```diff
@@ -1,3 +1,8 @@
+[pytest]
+minversion = 2.8.2
```
Lukasa: Should this requirement not also (or entirely) be expressed by tox.ini?

Lukasa:
One note about versioning. Either accept and edit accordingly, or reject because I'm crazy.

hynek:
done

Lukasa:
\o/

PR #381, opened 2015-10-25
Switch to a nicer changelog format

hynek:
The old format is tedious to read and write.  Let’s do better from now on.

Also document the general backward compatibility.


Lukasa:
LGTM. :cake:

PR #382, opened 2015-10-27
Support for serializing/deserializing public keys

Lukasa:
This is needed for HPKP in urllib3, to avoid using lots of cryptography private functions. It also adds symmetry to the equivalent private key functions.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -40,6 +40,8 @@ In accordance with our backward compatibility policy :func:`OpenSSL.rand.egd` wi
 Changes:
 ^^^^^^^^
 
+- Added :func:`OpenSSL.crypto.dump_publickey` to dump :class:`OpenSSL.crypto.PKey` objects that represent public keys, and :func:`OpenSSL.crypto.load_publickey` to load such objects from serialized representations.
+  [`#382 <https://github.com/pyca/pyopenssl/pull/382>`_]
 - Added :func:`OpenSSL.crypto.dump_crl` to dump a certificate revocation list out to a string buffer.
   [`#368 <https://github.com/pyca/pyopenssl/pull/368>`_]
 - Added :meth:`OpenSSL.SSL.Connection.state_string` using the OpenSSL binding ``state_string_long``.
@@ -54,10 +56,6 @@ Changes:
   This will default us to the setting that actually works.
   To revert this you can call ``OpenSSL.crypto._lib.ASN1_STRING_set_default_mask_asc(b"default")``.
   [`#234 <https://github.com/pyca/pyopenssl/pull/234>`_]
-- Added :func:`OpenSSL.crypto.dump_publickey` to dump :class:`OpenSSL.crypto.PKey` objects that represent public keys.
-  [`#382 <https://github.com/pyca/pyopenssl/pull/382>`_]
-- Added :func:`OpenSSL.crypto.load_publickey` to load :class:`OpenSSL.crypto.PKey` objects that represent public keys from serialized representations.
-  [`#382 <https://github.com/pyca/pyopenssl/pull/382>`_]
 
 
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -13,7 +13,7 @@ We have made *many* changes to make local development more pleasing.
 The test suite now passes both on Linux and OS X with OpenSSL 0.9.8, 1.0.1, and 1.0.2.
 It has been moved to `py.test <http://pytest.org/latest/>`_, all CI test runs are part of `tox <https://testrun.org/tox/>`_ and the source code has been made fully `flake8 <https://flake8.readthedocs.org/en/>`_ compliant.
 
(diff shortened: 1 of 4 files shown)


reaperhulk:
This needs a changelog entry :)

Lukasa:
@hynek Thoughts on how to hit that last line of the diff? It's going to be tricky to feed something that will cause `write_bio` to fail. @reaperhulk has some opinions he'd like to share.

reaperhulk:
Covering these failure conditionals is hard and we should really do it with a function like openssl_assert in cryptography :)

hynek:
You can put a pragma no cover on it for now.  Also please get rid of those dreadful `:py:` prefixes. :)

Lukasa:
Hooray, all green! \o/

Lukasa:
@hynek Done.

hynek:
Cory, what was the very first review comment by Paul? ;)

Lukasa:
@hynek I don't read Paul's comments. ;)

src/OpenSSL/crypto.py
```diff
@@ -1623,6 +1623,31 @@ def dump_certificate(type, cert):
     return _bio_to_string(bio)
 
 
+def dump_publickey(type, pkey):
+    """
+    Dump a public key to a buffer.
+
+    :param type: The file type (one of :py:data:`FILETYPE_PEM` or
```
hynek: Still :py:

src/OpenSSL/crypto.py
```diff
@@ -2404,6 +2429,36 @@ def _read_passphrase(self, buf, size, rwflag, userdata):
             return 0
 
 
+def load_publickey(type, buffer):
+    """
+    Load a public key from a buffer.
+
+    :param type: The file type (one of :py:data:`FILETYPE_PEM`,
+        :data:`FILETYPE_ASN1`).
+    :param buffer: The buffer the key is stored in.
```
hynek: Please add an explanation what type buffer can be.

src/OpenSSL/crypto.py
```diff
@@ -2404,6 +2429,36 @@ def _read_passphrase(self, buf, size, rwflag, userdata):
             return 0
 
 
+def load_publickey(type, buffer):
+    """
+    Load a public key from a buffer.
+
+    :param type: The file type (one of :py:data:`FILETYPE_PEM`,
```
hynek: Still py

Lukasa:
Updated.

src/OpenSSL/crypto.py
```diff
@@ -2404,6 +2429,37 @@ def _read_passphrase(self, buf, size, rwflag, userdata):
             return 0
 
 
+def load_publickey(type, buffer):
+    """
+    Load a public key from a buffer.
+
+    :param type: The file type (one of :data:`FILETYPE_PEM`,
+        :data:`FILETYPE_ASN1`).
+    :param buffer: The buffer the key is stored in. Must be a Python string
```
hynek: This si what we have `:type buffer:` for fine sir :)

src/OpenSSL/crypto.py
```diff
@@ -2404,6 +2429,37 @@ def _read_passphrase(self, buf, size, rwflag, userdata):
...
+        object, either unicode or bytestring.
+    :return: The :class:`PKey` object.
```
hynek: just make this `:rtype: PKey`

CHANGELOG.rst
```diff
@@ -54,6 +54,8 @@ Changes:
   This will default us to the setting that actually works.
   To revert this you can call ``OpenSSL.crypto._lib.ASN1_STRING_set_default_mask_asc(b"default")``.
   [`#234 <https://github.com/pyca/pyopenssl/pull/234>`_]
+- Added :func:`OpenSSL.crypto.dump_publickey` to dump :class:`OpenSSL.crypto.PKey` objects that represent public keys. [`#382 <https://github.com/pyca/pyopenssl/pull/382>`_]
```
hynek: I wouldn’t bother you if there wasn’t anything else, but since there were more changes: could you put the `[#num]` in separate lines please?

hynek:
1. please move the changelog entries to the top (within the changes section!) :D  you can also make it one entry for all I care.
2. you forgot to add the `autofunction` directives to crypto.rst :)  (we don't do automodule)

src/OpenSSL/crypto.py
```diff
@@ -1623,6 +1623,32 @@ def dump_certificate(type, cert):
...
 
 
+def dump_publickey(type, pkey):
+    """
+    Dump a public key to a buffer.
+
+    :param type: The file type (one of :data:`FILETYPE_PEM` or
+        :data:`FILETYPE_ASN1`).
+    :param pkey: The PKey to dump.
+    :type pkey: :class:`PKey`
```
hynek: How about `:param PKey pkey: The public key to dump`?

`:class:` is definitely not necessary if you write type markup (and it's only the one type).

Lukasa:
And updated again. =D

PR #384, opened 2015-10-27
Harmonize copyright notices

hynek:
- According to
https://en.wikipedia.org/wiki/Copyright_notice#Form_of_notice_for_visually_perceptible_copies
, it's enough to mention the year of first release.  If someone has
a connections to VanL, I'd gladly listen. :)
- The copyright notices that were floating around have been all moved into
NOTICE.
- Since Martin's name contains an ö, I've also added an encoding which
shouldn't be a problem in 2015.

***

It would be nice to know whether @exarkun feels violated in his copyrights by these changes in any way (IANAL but it should be all OK).

### DIFF
--- a/NOTICE
+++ b/NOTICE
@@ -1,0 +1,16 @@
+pyOpenSSL
+Copyright 2001 Martin Sjögren
+
+Initial work has been made while Martin worked at AB Strakt
+<http://www.strakt.com/>.
+
+This product includes software developed by Jean-Paul Calderone
+<http://as.ynchrono.us> who maintained pyOpenSSL from 2008 until 2015.
+
+Portions of this software have been taken from Twisted
+<https://twistedmatrix.com/>.
+
+Some of the examples have been contributed by Michal Wallace
+<http://www.sabren.net/> and Mihai Ibanescu <misa@redhat.com>.
+
+tests/test_rand.py has been originally written by Frederick Dean.
--- a/examples/sni/client.py
+++ b/examples/sni/client.py
@@ -1,5 +1,18 @@
-# Copyright (C) Jean-Paul Calderone
-# See LICENSE for details.
+# -*- coding: utf-8 -*-
+
+# Copyright 2001 Martin Sjögren and pyOpenSSL contributors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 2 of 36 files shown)


Lukasa:
Seems reasonable to me, but I'd like to hold off until @exarkun gives his ok too.

hynek:
Since JP doesn’t seem to care (I pinged him on IRC too with ne response) I think this can be merged unless @alex or @reaperhulk disagree.

lvh:
Seems reasonable to me. I would prefer actual confirmation from @exarkun, but I know he's not always available, and since it seems like this change really only represents the same information in a consistent way, I don't think he would object.

NOTICE
```diff
@@ -0,0 +1,16 @@
+pyOpenSSL
+Copyright 2001 Martin Sjögren
+
+Initial work has been made while Martin worked at AB Strakt
+<http://www.strakt.com/>.
+
+This product includes software developed by Jean-Paul Calderone
+<http://as.ynchrono.us> who maintained pyOpenSSL from 2008 until 2015.
```
exarkun: missing a trailing /

examples/sni/client.py
```diff
@@ -1,5 +1,18 @@
-# Copyright (C) Jean-Paul Calderone
-# See LICENSE for details.
+# -*- coding: utf-8 -*-
+
+# Copyright 2001 Martin Sjögren and pyOpenSSL contributors
```
exarkun: This file wasn't copyright Martin before.  So I wonder why it is now.

exarkun: Same comment for various other files changed in the PR.

src/OpenSSL/SSL.py
```diff
@@ -1,3 +1,19 @@
+# -*- coding: utf-8 -*-
+
+# Copyright 2001 Martin Sjögren and pyOpenSSL contributors
```
exarkun: Since I wrote all of the code in this module from scratch (but not clean-room), I'm not sure Martin's name belongs here.  Same comment for most of the other Python implementation and test modules.

exarkun:
Sorry about the delay in providing feedback.

See my comments inline.

As is, I'd prefer not to see these changes merged.

Thanks.

hynek:
Well I'm not sure what the legally correct way is.  The thought process was that it's originally Martin's project and you contributed to it.

I guess I don't care enough to put any more work into this so it will stay the clusterfuck it is.

Issue #393, opened 2015-12-04
3 tests fail with MD5 in recent python/Fedora 23/OpenSSL


aalba6675:
Environment:
Fedora 23
openssl 1.0.2d
python 2.7.10

Test failures:
Clean git checkout of master 3 test failures with the same error. Is this expected? Location of failures:

X509ReqTests.test_sign
X509ReqTests.test_verify_success
NetscapeSPKITests.test_sign
E       Error: [('asn1 encoding routines', 'ASN1_item_verify', 'unknown message digest algorithm')]

Update: 
if I replace GOOD_DIGEST with SHA1, SHA256 455 passes
if I replace GOOD_DIGEST with SHA384: 3 errors
E       Error: [('rsa routines', 'RSA_sign', 'digest too big for rsa key'), ('asn1 encoding routines', 'ASN1_item_sign_ctx', 'EVP lib')]

Can we change tests/ to using SHA1 or SHA256?

reaperhulk:
I believe MD5 may be disabled in Fedora 23's OpenSSL. The tests probably don't need to be run against an old and insecure algorithm like that, so we can look at updating it.

For SHA384 you're generating a 48 byte digest. Coupled with the necessary PKCS#1 v1.5 padding and the small key sizes in the tests that would explain the error you're seeing.

aalba6675:
I submitted a pull request changing GOOD_DIGEST="SHA1", locally I have 455 tests passing.

Two CI jobs failed, unfortunately I couldn't decipher the error message, didn't seem to have anything to do with the commit.

aalba6675:
Updated branch with coding convention fix.

hynek:
I believe this is fixed.

PR #394, opened 2015-12-04
test_crypto.py: Enable tests to pass when MD5 verification disabled in OpenSSL.

aalba6675:
Change GOOD_DIGEST to SHA1 to allow tests to pass for distros that have disabled MD5 verification in OpenSSL. 455 tests pass on Fedora 23.

### DIFF
--- a/tests/test_crypto.py
+++ b/tests/test_crypto.py
@@ -54,8 +54,6 @@ def normalize_privatekey_pem(pem):
 GOOD_CIPHER = "blowfish"
 BAD_CIPHER = "zippers"
 
-#GOOD_DIGEST = "MD5"
-# MD5 verification disabled in some distros OpenSSL
 GOOD_DIGEST = "SHA1"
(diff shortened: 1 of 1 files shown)


reaperhulk:
This fixes #393 

The OS X test failure here is due to https://github.com/pyca/cryptography/issues/2503, which will be resolved soon. The flake8 failure is your patch. Specifically the lack of a space on the old GOOD_DIGEST line. However, I'd prefer you just remove that line rather than correct the linting as it is unneeded.

aalba6675:
Fix coding convention violation.

reaperhulk:
Thanks for the contribution -- once that upstream bug in cryptography is resolved we'll be able to potentially merge this.

hynek:
We have a working build again, please rebase.

hynek:
Nevermind, we switched that stupid requirement off :)

PR #396, opened 2015-12-09
Update docstrings for get_issuer/get_subject

### LINKED ISSUE #395: Incorrect documentation for X509Req.get_subject
The documentation for X509Req’s function get_subject has the following paragraph:

> This creates a new :py:class:`X509Name`: modifying it does not affect this request.

This appears to be false. I created a fresh X509Req, called get_subject(), assigned to the CN member of the returned object, signed the request, and dumped the request to a file. The `openssl req` command showed the subject field filled in accordingly. This makes sense, given that there otherwise is no other way to actually set the subject in the first place!

Lukasa:
Resolves #395.

This is the *easy* route out of this issue: simply assume the docstrings were in error and update them to reflect the actual behaviour of the code.

This might not be what we want to do though. The particularly confusing thing is that this leads to lots of spooky action at a distance, where multiple objects refer through to the same singleton object. I have terrifying visions of thread safety and weird other nonsense, and it's not clear to me that these are necessarily the best ways to manipulate the subject/issuer fields. *Also*, the `X509` class has `set_issuer`/`set_subject` methods, which further suggests that having the `X509Name` class edit the underlying structure might not have been the original intention.

We can go another way here, and copy the underlying data out when we create these in the `get` methods. This will cause the original docstrings to become correct. We'd also have to add a `set_subject` method to the `X509Req` class.

Thoughts?

### DIFF
--- a/src/OpenSSL/crypto.py
+++ b/src/OpenSSL/crypto.py
@@ -1314,11 +1314,13 @@ class X509(object):
         """
         Return the issuer of this certificate.
 
-        This creates a new :py:class:`X509Name`: modifying it does not affect
-        this certificate.
+        This creates a new :class:`X509Name` that wraps the underlying issuer
+        name field on the certificate. Modifying it will modify the underlying
+        certificate, and will have the effect of modifying any other
+        :class:`X509Name` that refers to this issuer.
 
         :return: The issuer of this certificate.
-        :rtype: :py:class:`X509Name`
+        :rtype: :class:`X509Name`
         """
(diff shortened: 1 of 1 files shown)


hynek:
If we changed the behavior it would be a huge backward incompatible change…

So once our builders work again I’m tending to merge.

hynek:
could you add a "!=2.8.4" pin to py.test like cryptography has please?

Lukasa:
For you, @hynek, of course. :cookie:

hynek:
ugh wrong button

Issue #399, opened 2015-12-16
X509Name.get_components() should decode attribute's value


daa:
Currently `OpenSSL.crypto.X509Name.get_components()` makes string from raw attribute's value without performing any conversion, however `X509Name.__getattr__` converts asn1 string to utf8 representation and then decodes it in Python and returns unicode object. I think that for consistency both method should return the same values. But this inconsistency is not the worst - `_ffi.string()` stops on the first null byte which may be present if original value is encoded in utf-16 and it can lead to incomplete returned value. For example if certificate subject contains O="SomeOrg Орг" encoded in utf-16be, `get_components()` for this attribute will return `('O', '')` because value's byte representation will be `'\00S...'`.

If you don't have any objections, i'll prepare pull request changing this behaviour.

Just noticed that some projects - twisted for example - depend on `get_components()` value element being byte string, so I suggest not decoding to unicode object there but just translation to utf-8 encoded binary string.

daa:
I've created example certificate which shows this issue:
```
>>> import OpenSSL.crypto
>>> with open('example-mbstring-cert.crt', 'rb') as f:
...     s = f.read()
... 
>>> c = OpenSSL.crypto.load_certificate(OpenSSL.crypto.FILETYPE_PEM, s)
>>> c.get_subject()
<X509Name object '/C=RU/CN=some-common-name/L=Any/O=\x00O\x00r\x00g\x00 \x04\x1E\x04@\x043/ST=Some/OU=some-ou'>
>>> c.get_subject().O
u'Org \u041e\u0440\u0433'
>>> c.get_subject().get_components()
[('C', 'RU'), ('CN', 'some-common-name'), ('L', 'Any'), ('O', ''), ('ST', 'Some'), ('OU', 'some-ou')]
```

[example-mbstring-cert.zip](https://github.com/pyca/pyopenssl/files/65236/example-mbstring-cert.zip)

As you can see `O` attribute is properly decoded to unicode but in `get_components()` it's not and is empty.

hynek:
I’m afraid we can’t change the behavior of `OpenSSL.crypto.X509Name.get_components()` since that would break backward-compatibility (as you’ve noted yourself).

But the NUL byte thing ought to be fixed.

daa:
[...]

PR #400, opened 2015-12-21
convert X509Name.get_components() value data to utf-8, issue #399

daa:
Now value part in `get_components()` is converted to utf-8 but stays in binary form, so value is not lost and we know it's encoding. This pull request fix issue #399 when BMPString data is lost in case of current pyOpenSSL or its encoding is unknown in case of pyOpenSSL-0.13.

### DIFF
--- a/src/OpenSSL/crypto.py
+++ b/src/OpenSSL/crypto.py
@@ -156,6 +156,31 @@ def _get_asn1_time(timestamp):
             return string_result
 
 
+def _asn1_string_to_utf8(asn1_string):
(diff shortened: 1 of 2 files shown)


src/OpenSSL/crypto.py
```diff
@@ -156,6 +156,31 @@ def _get_asn1_time(timestamp):
             return string_result
 
 
+def _asn1_string_to_utf8(asn1_string):
```
agladkov: There are bytes. Not UTF-8.
It is better to call function *_to_bytes()

PR #402, opened 2015-12-26
Add pypy3 to Travis

scop:


Lukasa:
LGTM.

alex:
pypy3 is Python 3.2, which we don't support.

On Sun, Dec 27, 2015 at 4:25 AM, Cory Benfield <notifications@github.com>
wrote:

> LGTM.
>
> —
> <https://github.com/pyca/pyopenssl/pull/402#issuecomment-167396499>.
>



-- 

Lukasa:
Oh, yeah, there's that. Whoops!

hynek:
This kind of only makes sense if pypy3 is going to support Python 3.3 (or later) soonish since we’ve dropped Python 3.2.

@alex?

alex:
It's being worked on, but there's no timeline for when a 3.3 will be out, not worth adding a build which isn't passing and which we can't do anything to fix, IMO.

PR #404, opened 2016-01-06
Rename readme dep and sprinkle in skip_installs

hynek:
Library name changed and skip_install will speed up our test times.


Lukasa:
LGTM.

PR #406, opened 2016-01-08
support the new cryptography 1.2 method of getting key type

reaperhulk:


src/OpenSSL/crypto.py
```diff
@@ -248,7 +248,12 @@ def check(self):
         if self._only_public:
             raise TypeError("public key only")
 
-        if _lib.EVP_PKEY_type(self._pkey.type) != _lib.EVP_PKEY_RSA:
+        if hasattr(self._pkey, "type"):
+            key_type = self._pkey.type
+        else:
+            key_type = _lib.Cryptography_EVP_PKEY_id(self._pkey)
+
+        if _lib.EVP_PKEY_type(key_type) != _lib.EVP_PKEY_RSA:
```
alex: This can just be `self.type()`?

reaperhulk: Yup, wasn't sure if we should refactor as part of this though. I can do that though.

alex: Yes, I think it's a clear improvement.

alex:
Coverage miss is because we don't have CI for an older version of cryptography. I think this is ok though.

hynek:
We must add a tox/travis target for the last cryptography that worked the old way so we don’t break it in the future.

reaperhulk:
I can put a try catch back in here so that we don't have to raise the cryptography minimum version (which will prevent the pypy version requirement bump due to cffi 1.0+), but someone else will need to add a CI target that uses an older cryptography to get coverage.

hynek:
proceed please :)

hynek:
fixed in #418

PR #408, opened 2016-01-16
Don't use SSLv2 or SSLv3 in examples/simple

dveeden:


alex:
looks like there are some unrelated failures here, I'll clean those up and then we can get this merged

reaperhulk:
:+1: on the patch itself

alex:
(Closing and reopening to trick travis into rerunning)

Issue #388, opened 2015-11-11
set_session_id does nothing


glyph:
The method body of `Context.set_session_id` is empty. That seems like it is probably not what callers expect.

hynek:
pyOpenSSL does this:

```c
static char ssl_Context_set_session_id_doc[] = "\n\
Set the session identifier.  This is needed if you want to do session\n\
resumption.\n\
\n\
:param buf: A Python object that can be safely converted to a string\n\
:returns: None\n\
";
static PyObject *
ssl_Context_set_session_id(ssl_ContextObj *self, PyObject *args)
{
    unsigned char *buf;
    unsigned int len;

    if (!PyArg_ParseTuple(args, "s#:set_session_id", &buf, &len))
        return NULL;

    if (!SSL_CTX_set_session_id_context(self->ctx, buf, len))
    {
        exception_from_error_queue(ssl_Error);
        return NULL;
    }
    else
    {
        Py_INCREF(Py_None);
        return Py_None;
    }
}
```

Sadly `SSL_CTX_set_session_id_context` isn’t bound in cryptography.

Issue #387, opened 2015-11-11
load_client_ca does nothing


glyph:
The method body of `Context.load_client_ca` is empty. That seems like it is probably not what callers expect.

Lukasa:
Uh...yes. That seems...wrong?

hynek:
pyOpenSSL 0.13 does this:

```c
static char ssl_Context_load_client_ca_doc[] = "\n\
Load the trusted certificates that will be sent to the client (basically\n\
telling the client \"These are the guys I trust\").  Does not actually\n\
imply any of the certificates are trusted; that must be configured\n\
separately.\n\
\n\
:param cafile: The name of the certificates file\n\
:return: None\n\
";
static PyObject *
ssl_Context_load_client_ca(ssl_ContextObj *self, PyObject *args)
{
    char *cafile;

    if (!PyArg_ParseTuple(args, "s:load_client_ca", &cafile))
        return NULL;

    SSL_CTX_set_client_CA_list(self->ctx, SSL_load_client_CA_file(cafile));

    Py_INCREF(Py_None);
    return Py_None;
}
```

Unfortunately,  `SSL_load_client_CA_file` isn’t bound in `cryptography`.  Maybe there’s another way?  Pinning to higher cryptography options is sadly not an option due to the CFFI/PyPy/Travis situation.

alex:
I'll go ahead and send a PR to cryptography anyways, so that it's at least
there in the future.

On Sun, Jan 31, 2016 at 9:59 AM, Hynek Schlawack <notifications@github.com>
wrote:

> pyOpenSSL 0.13 does this:
>
> static char ssl_Context_load_client_ca_doc[] = "\n\Load the trusted certificates that will be sent to the client (basically\n\telling the client \"These are the guys I trust\").  Does not actually\n\imply any of the certificates are trusted; that must be configured\n\separately.\n\\n\:param cafile: The name of the certificates file\n\:return: None\n\";static PyObject *ssl_Context_load_client_ca(ssl_ContextObj *self, PyObject *args)
> {
>     char *cafile;
>
>     if (!PyArg_ParseTuple(args, "s:load_client_ca", &cafile))
>         return NULL;
>
>     SSL_CTX_set_client_CA_list(self->ctx, SSL_load_client_CA_file(cafile));
>
>     Py_INCREF(Py_None);
>     return Py_None;
> }
>
> Unfortunately, SSL_load_client_CA_file isn’t bound in cryptography. Maybe
[...]

PR #410, opened 2016-01-17
Raise NotImplementedError for Unimplemented Functions

mhils:
As the title says - just silently doing nothing is really unexpected.

refs #387, refs #388

### DIFF
--- a/src/OpenSSL/SSL.py
+++ b/src/OpenSSL/SSL.py
@@ -696,6 +696,7 @@ class Context(object):
         :param cafile: The name of the certificates file
         :return: None
         """
+        raise NotImplementedError()
 
     def set_session_id(self, buf):
         """
@@ -705,6 +706,7 @@ class Context(object):
         :param buf: A Python object that can be safely converted to a string
         :returns: None
         """
+        raise NotImplementedError()
 
     def set_session_cache_mode(self, mode):
         """
@@ -1406,6 +1408,7 @@ class Connection(object):
 
         :return: True if the renegotiation can be started, false otherwise
         """
+        raise NotImplementedError()
 
     def do_handshake(self):
         """
@@ -1424,6 +1427,7 @@ class Connection(object):
 
         :return: Whether there's a renegotiation in progress
         """
+        raise NotImplementedError()
 
     def total_renegotiations(self):
         """
--- a/tests/test_ssl.py
+++ b/tests/test_ssl.py
@@ -1293,8 +1293,15 @@ class ContextTests(TestCase, _LoopbackMixin):
             Error, context.use_certificate_chain_file, self.mktemp()
         )
 
-    # XXX load_client_ca
-    # XXX set_session_id
+    def test_load_client_ca(self):
+        context = Context(TLSv1_METHOD)
+        with pytest.raises(NotImplementedError):
+            context.load_client_ca(None)
+
+    def test_set_session_id(self):
+        context = Context(TLSv1_METHOD)
+        with pytest.raises(NotImplementedError):
+            context.set_session_id("session")
 
     def test_get_verify_mode_wrong_args(self):
         """
@@ -3159,6 +3166,16 @@ class ConnectionRenegotiateTests(TestCase, _LoopbackMixin):
         """
         connection = Connection(Context(TLSv1_METHOD), None)
         self.assertEquals(connection.total_renegotiations(), 0)
+
+    def test_renegotiate(self):
+        connection = Connection(Context(TLSv1_METHOD), None)
+        with pytest.raises(NotImplementedError):
+            connection.renegotiate()
+
+    def test_renegotiate_pending(self):
+        connection = Connection(Context(TLSv1_METHOD), None)
+        with pytest.raises(NotImplementedError):
+            connection.renegotiate_pending()
 
 #     def test_renegotiate(self):
 #         """
--- a/tests/test_ssl.py
+++ b/tests/test_ssl.py
@@ -2540,7 +2540,6 @@ class ConnectionTests(TestCase, _LoopbackMixin):
         ctx = Context(TLSv1_METHOD)
         ctx.use_privatekey(key)
         ctx.use_certificate(cert)
-        ctx.set_session_id("unity-test")
 
         def makeServer(socket):
             server = Connection(ctx, socket)
@@ -2580,7 +2579,6 @@ class ConnectionTests(TestCase, _LoopbackMixin):
         ctx = Context(TLSv1_METHOD)
         ctx.use_privatekey(key)
         ctx.use_certificate(cert)
-        ctx.set_session_id("unity-test")
 
         def makeServer(socket):
             server = Connection(ctx, socket)


Lukasa:
This seems reasonable, but I'll let @hynek make that call. Otherwise the patch looks good, though you'll want to improve the test coverage to account for these exceptions.

mhils:
I'd personally rather see `raise NotImplementedError` added to `.coveragerc`'s `exclude_lines`.
Thoughts?

hynek:
1. If we do this, we should totally add tests to it that verify those errors are raised.  That’s also an indication of intent.
2. I would like more input on this (@glyph @alex @reaperhulk ): this is clearly breaking backward compat.  But I have a hunch that it's the right thing to do…opinions?

glyph:
If code is currently calls these APIs and expects to get the security properties they advertise, it is a security flaw to pretend that it worked.  Granted, none of these functions seem like an API which adds a restriction, but on general principle, I think they should be made to raise errors and do a release ASAP.  It's compatibility-breaking, but in that sense it is a compatibility-breaking security update.

alex:
I agree, alernatively we could fix them ASAP.

On Mon, Jan 18, 2016 at 1:57 AM, Glyph <notifications@github.com> wrote:

> If code is currently calls these APIs and expects to get the security
> properties they advertise, it is a security flaw to pretend that it worked.
> Granted, none of these functions seem like an API which adds a restriction,
> but on general principle, I think they should be made to raise errors and
> do a release ASAP. It's compatibility-breaking, but in that sense it is a
> compatibility-breaking security update.
>
> —
> <https://github.com/pyca/pyopenssl/pull/410#issuecomment-172444861>.
>



-- 

reaperhulk:
I'd suggest fixing everything but renegotiate, but raising notimplemented on all of them is definitely an improvement over silently doing nothing.

mhils:
I updated the PR with tests for all four functions.

hynek:
So um it seems to break *a lot of tests* in Twisted: https://travis-ci.org/pyca/pyopenssl/jobs/103128063 you really cool with that @glyph ? :)

glyph:
> So um it seems to break a lot of tests in Twisted: https://travis-ci.org/pyca/pyopenssl/jobs/103128063 you really cool with that @glyph ? :)

First of all, thanks to whoever set up CI to track that sort of thing.  It's very useful to have this information.

Second, it appears that *all* of the errors are going through the same call to `set_session_id`.

I'm not crazy about breaking all these tests, but I'm not entirely sure that the call to `set_session_id` even makes sense.  Good use of `md5`, of course.  But also, does it even make sense to call `set_session_id` and never call `SSL_CTX_set_sess_get_cb`?  The client can't resume a session, so it's a bit nonsense to enable sessions in the first place, as I understand it.

So, I don't know.  It appears harmless, so maybe it's not worth breaking, but this code in Twisted probably *does* need to be rewritten.  The question is, if `set_session_id` starts working, is this code a problem?  Somebody who knows TLS a little better than me might have to answer.

hynek:
*lightspot goes to @reaperhulk & @tiran *

hynek:
(and as for who set it up: remember pyOpenSSL 0.15.0? :|)

I’d really like at least a smoke test for urllib3.

tiran:
It's unclear to me what ```SSL.Context.set_session_id()``` is suppose to do. I guess it's meant to be used for https://www.openssl.org/docs/manmaster/ssl/SSL_CTX_set_session_id_context.html . However then the name isn't right. The context sid is not related to or to be confused with the session id in during the TLS/SSL handshake. It rather sets an identifier for the session context to ensure that sessions aren't shared between unrelated contexts. It's mostly useful for import/export of serialized sessions on the server side. It doesn't hurt to set it on the client side, though.

The session id context doesn't need to be secure or unpredictable. Apache uses the vhost name. CPython simply sets it to "Python".

tiran:
By the way the settings are only relevant for server-side sessions. For client side an application has to set the session an application has to set an explicit session, e.g. with ```SSL_set_session(dst_ssl, SSL_get_session(src_ssl))```. Both ```SSL*``` should refer to the same ```SSL_CTX*```.

https://www.openssl.org/docs/manmaster/ssl/SSL_set_session.html
> SSL_SESSION objects keep internal link information about the session cache list, when being inserted into one SSL_CTX object's session cache. One SSL_SESSION object, regardless of its reference count, must therefore only be used with one SSL_CTX object (and the SSL objects created from this SSL_CTX object).

hynek:
FYI, this is what “old” pyOpenSSL did:

```c

static char ssl_Context_set_session_id_doc[] = "\n\
Set the session identifier.  This is needed if you want to do session\n\
resumption.\n\
\n\
:param buf: A Python object that can be safely converted to a string\n\
:returns: None\n\
";
static PyObject *
ssl_Context_set_session_id(ssl_ContextObj *self, PyObject *args)
{
    unsigned char *buf;
    unsigned int len;

    if (!PyArg_ParseTuple(args, "s#:set_session_id", &buf, &len))
        return NULL;

    if (!SSL_CTX_set_session_id_context(self->ctx, buf, len))
    {
        exception_from_error_queue(ssl_Error);
        return NULL;
    }
    else
    {
        Py_INCREF(Py_None);
        return Py_None;
    }
}
```

If the name is wrong, we should implement and deprecate it at once I guess?

hynek:
After some research it appears to me that the reason why they’re not implemented is that cryptography is missing bindings.

Unless we get a back port to older cryptography versions, we can’t implement them for now.

Would raising a warning be an adequate compromise?

hynek:
@reaperhulk any particular reason why you’d implement everything but renegotiate?  Turns out those two methods are the only that have the necessary bindings. :)

glyph:
@hynek Yes, I think a warning would be the best option available immediately.

reaperhulk:
Mostly because renegotiation is not a good feature in TLS and it's not common to actually need it. :)

glyph:
> Mostly because renegotiation is not a good feature in TLS and it's not common to actually need it. :)

The main people asking for it that I've seen are security researchers looking for new exploits in it :)

hynek:
@mhils so…would you mind transforming those Exceptions into RuntimeWarnings?

I think `Base class for warnings about dubious runtime behavior.` hits the nail on the head…

hynek:
Or please don't do anything until #419 is decided on. 😞

hynek:
I’ve more or less decided to not care about Travis/PyPy anymore and redirect all the hate to them if necessary.

Unfortunately there’s no release with the bindings and 1.3 is weeks away.

@reaperhulk offered to release a cryptography 1.2 point release with the necessary bindings.  I guess that would be a viable solution.

We should also figure out how to test on PyPy ourselves since Travis isn’t likely to show any progress anytime soon I’ve heard.

glyph:
> I’ve more or less decided to not care about Travis/PyPy anymore and redirect all the hate to them if necessary.

💯

glyph:
> We should also figure out how to test on PyPy ourselves since Travis isn’t likely to show any progress anytime soon I’ve heard.

Mimic's build configuration sets a reasonable example here.

hynek:
There’s multiple examples but someone has to actually do it. :)  And integrate with our already nontrivial setup. :(

hynek:
Waiting for pyca/cryptography#2715 to decide what route to go.

hynek:
We’re implementing them!  #422 Sorry @mhils !

PR #412, opened 2016-01-19
2016 snook up on us

hynek:


setup.cfg
```diff
@@ -1,7 +1,7 @@
 [pytest]
 minversion = 2.8.2
 strict = true
-norecursedirs = build dist .* *.egg doc _trial_temp
+testpaths = tests
```
Lukasa: Uh, where did this come from?

hynek: I was just boyscouting a bit.  `testpaths` came with py.test 2.8.0 and simply tells where to look for tests instead of black listing everything else.

Lukasa: Ok cool.

PR #415, opened 2016-01-21
compare datetimes directly for the expiry check

### LINKED ISSUE #414: Double-check #378
Questions about the adequacy of the fix in #378 has been raised.  We should double-check.

reaperhulk:
fixes #414 

This removes the unsafe (and pointless) conversion of the datetime object to a unix timestamp using mktime (as noted by @zed in #378).

### DIFF
--- a/src/OpenSSL/crypto.py
+++ b/src/OpenSSL/crypto.py
@@ -1,6 +1,5 @@
 import datetime
 
-from time import mktime
 from base64 import b16encode
 from functools import partial
 from operator import __eq__, __ne__, __lt__, __le__, __gt__, __ge__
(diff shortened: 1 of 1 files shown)


hynek:
@zed do you agree with this solution?

zed:
@hynek I agree with comparing datetimes directly (I've suggested it). Can't comment on the wider context e.g., whether `not_before` should be checked too or whether UTCTime (YYMMDDHHMMSSZ) format should be supported here in addition to GeneralizedTime (YYYYMMDDHHMMSSZ). https://tools.ietf.org/html/rfc5280#section-4.1.2.5

PR #417, opened 2016-01-31
Greatly simplify test_state_string

### LINKED ISSUE #383: tests.test_ssl.ConnectionTests.test_state_string has race conditions
`tests.test_ssl.ConnectionTests.test_state_string` fails fairly often in CI.  Seems like things aren’t 100% synchronized?

This needs to be fixed before a new release.  ping @kjav

hynek:
It used to more or less a functional test.  We don't need to check whether the function works for all possible states.

This will hopefully fix #383 for good.


Lukasa:
LGTM! :sparkles: :cake: :sparkles:

PR #422, opened 2016-02-13
Implement missing methods

hynek:
First shot at #388.

- It would be nice if @tiran (or anyone else understand the matter) could suggest a better docstring.
- If anyone knows what could be checked in the test case I’m all ears.  There doesn’t seem to be a get_session_id?
- We also need to find a way to simulate errors. :-/

N.B. only the `-cryptographyMaster` envs are supposed to pass.  Once this and the other missing methods are signed off, @reaperhulk will release a cryptography 1.2 point release with the necessary bindings.

- [x] set cryptography requirement to `>=1.3` in setup.py once 1.3 is available.

### DIFF
--- a/src/OpenSSL/SSL.py
+++ b/src/OpenSSL/SSL.py
@@ -689,23 +689,39 @@ class Context(object):
 
     def load_client_ca(self, cafile):
         """
-        Load the trusted certificates that will be sent to the client
-        (basically telling the client "These are the guys I trust").  Does not
-        actually imply any of the certificates are trusted; that must be
+        Load the trusted certificates that will be sent to the client.  Does
+        not actually imply any of the certificates are trusted; that must be
         configured separately.
 
-        :param cafile: The name of the certificates file
+        :param bytes cafile: The path to a certificates file in PEM format.
         :return: None
         """
+        ca_list = _lib.SSL_load_client_CA_file(
+            _text_to_bytes_and_warn("cafile", cafile)
+        )
+        _openssl_assert(ca_list != _ffi.NULL)
+        # SSL_CTX_set_client_CA_list doesn't return anything.
+        _lib.SSL_CTX_set_client_CA_list(self._context, ca_list)
 
     def set_session_id(self, buf):
         """
-        Set the session identifier.  This is needed if you want to do session
-        resumption.
+        Set the session id to *buf* within which a session can be reused for
+        this Context object.  This is needed when doing session resumption,
+        because there is no way for a stored session to know which Context
+        object it is associated with.
 
-        :param buf: A Python object that can be safely converted to a string
+        :param bytes buf: The session id.
+
         :returns: None
         """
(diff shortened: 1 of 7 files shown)


src/OpenSSL/SSL.py
```diff
@@ -699,12 +699,22 @@ def load_client_ca(self, cafile):
...
-        :param buf: A Python object that can be safely converted to a string
         :returns: None
         """
+        buf = _text_to_bytes_and_warn("buf", buf)
+        if _lib.SSL_CTX_set_session_id_context(
+                self._context,
+                buf,
+                len(buf),
+        ) != 1:
+            _raise_current_error()
```
reaperhulk: You can get around the lack of coverage for this case by making a new function:

```python
def pyopenssl_assert(ok):
    if not ok:
        _raise_current_error()
```

then you can change the above to

```python
        res = _lib.SSL_CTX_set_session_id_context(
                self._context,
                buf,
                len(buf),
        )
        _pyopenssl_assert(res != 1)
```

Coverage on `_pyopenssl_assert` is, of course, trivial then. This is the approach we use in cryptography when we have response codes we expect to always be a specific value unless something catastrophic has occurred.

reaperhulk: That said, in this case you could trigger the error by passing a session ID context value greater than `SSL_MAX_SSL_SESSION_ID_LENGTH`, which appears to be 32 (bytes?)

hynek: Good idea, done!  I’ve taken a, much higher number in the hopes that this number won't change significantly in the next OpenSSL release…

hynek:
(FYI as soon as this is signed off [but no sooner], I’ll tackle the next one.  It would be nice tho if someone else could start working on them too.)

hynek:
FYI I’ll fix #387 and #305 here too.  It’s cumbersome to work on them separately.

hynek:
Relevant builders are green.

Please tear it apart so we can have a 16.0.0 soon!

alex:
@hynek this needs a rebase

hynek:
done, all relevant builds are green.

tests/test_ssl.py
```diff
@@ -3129,24 +3240,26 @@ def test_total_renegotiations(self):
         connection = Connection(Context(TLSv1_METHOD), None)
         self.assertEquals(connection.total_renegotiations(), 0)
 
-#     def test_renegotiate(self):
-#         """
-#         """
-#         server, client = self._loopback()
+    def test_renegotiate(self):
```
reaperhulk: Should this test call `renegotiate_pending` during and after to confirm it properly returns True/False at the expected time?

hynek: I’ve added some asserts and the builders are still green.

glyph:
This is amazing and all you guys are awesome.

PR #433, opened 2016-03-10
treat x509 extension objects as opaque pointers

reaperhulk:
These functions have been bound a long time in cryptography and `X509_EXTENSION` needs to become opaque for OpenSSL 1.1.0 (see pyca/cryptography#2798)


alex:
@hynek this PR looks ok to me, not sure what to say about the coverage?

Lukasa:
LGTM. Looks like the coverage failure is because we don't have tests where the extension object is *not* NID_crl_reason. That was *already* true, so this diff doesn't regress coverage, but if it's possible to have such a test it'd probably make hynek happier.

hynek:
It's not a regression but we should try really hard to complete coverage for lines we touch.

In this case for all I know `X509_EXTENSION_free` doesn’t exist.

alex:
So basically, @reaperhulk please add a test for these methods with a extension that's not CRL reason.

hynek:
If it's not too much trouble, it would be nice if you could write that tests.  Those branches are non-trivial and I don't even know how to manually test them.

tests/test_crypto.py
```diff
@@ -2933,9 +2972,34 @@ def test_b64_encode(self):
...
+        The get_reason method on the Revoked class checks to see if the
+        extension is NID_crl_reason and should skip it otherwise. This test
+        loads a CRL with extensions it should ignore.
+        """
+        crl = load_crl(FILETYPE_PEM, crlDataUnsupportedExtension)
+        revoked = crl.get_revoked()
+        reason = revoked[1].get_reason()
+        assert reason == b'Unspecified'
+
+    def test_ignores_unsupported_revoked_cert_extension_set_new_reason(self):
```
hynek: *cough* docstring :) *cough*

PR #434, opened 2016-03-10
switch to DSA_generate_parameters_ex

### LINKED ISSUE #432: Stop using DSA_generate_parameters
DSA_generate_parameters was deprecated in 0.9.8. The new hotness is DSA_generate_parameters_ex.

(These issues are part of my work to get cryptography functioning with OpenSSL 1.1.0)

reaperhulk:
fixes #432

### DIFF
--- a/src/OpenSSL/crypto.py
+++ b/src/OpenSSL/crypto.py
@@ -221,6 +221,8 @@ class PKey(object):
             if dsa == _ffi.NULL:
                 # TODO: This is untested.
                 _raise_current_error()
+
+            dsa = _ffi.gc(dsa, _lib.DSA_free)
             res = _lib.DSA_generate_parameters_ex(
                 dsa, bits, _ffi.NULL, 0, _ffi.NULL, _ffi.NULL, _ffi.NULL
             )
@@ -230,7 +232,7 @@ class PKey(object):
             if not _lib.DSA_generate_key(dsa):
                 # TODO: This is untested.
                 _raise_current_error()
-            if not _lib.EVP_PKEY_assign_DSA(self._pkey, dsa):
+            if not _lib.EVP_PKEY_set1_DSA(self._pkey, dsa):
                 # TODO: This is untested.
                 _raise_current_error()
         else:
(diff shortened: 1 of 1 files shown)


src/OpenSSL/crypto.py
```diff
@@ -217,11 +217,16 @@ def generate_key(self, type, bits):
...
-                bits, _ffi.NULL, 0, _ffi.NULL, _ffi.NULL, _ffi.NULL, _ffi.NULL)
+            dsa = _lib.DSA_new()
             if dsa == _ffi.NULL:
                 # TODO: This is untested.
                 _raise_current_error()
+            res = _lib.DSA_generate_parameters_ex(
+                dsa, bits, _ffi.NULL, 0, _ffi.NULL, _ffi.NULL, _ffi.NULL
+            )
+            if not res == 1:
+                # TODO: This is untested.
```
alex: You need to free `dsa` in this path.

hynek: Wasn’t there a way to add objects to GC?  I would prefer that with assert_openssl coming soon.

reaperhulk: I can register the DSA object to free with the gc (ffi.gc) but `assign_DSA` assigns memory ownership to the EVP_PKEY object, so that will ultimately free the memory in the success case. So either we explicitly free it in this path or I register it for gc at creation and then switch to `set1_DSA` which will increment refcount and make it so calling DSA_free the first time will decrement refcount by 1 (when the proxy object falls out of scope), and then the EVP_PKEY_free call will work later as it'll decrement refcount to 0.

hynek: I tend to prefer the `ffi.gc` version I think?  Seems cleaner to me, although it requires a comment explaining what’s going on there which by itself is a red flag.

I’ll leave the decision to you, just make sure nothing leaks. :)

reaperhulk:
Okay, this now uses the `gc` approach. Once #422 lands then a lot of the branches can go away, but for now they're no worse than what's already there.

hynek:
LGTM But I'm leaving this to @alex to merge.

PR #435, opened 2016-03-10
X509_REVOKED_dup is a thing cryptography can do for you

### LINKED ISSUE #431: Stop using M_ASN1_TIME_dup
This macro is going away in OpenSSL 1.1.0.

All it really does is `ASN1_STRING_dup((const ASN1_STRING *)a)` (where `a` is an `ASN1_TIME`).

Filing this so we remember to do something about it. On the cryptography side I will make it so the symbol is still available to avoid breaking older pyopenssl.

reaperhulk:
This is one way to fix #431. Unfortunately it causes a failure in `test_add_revoked_keyword`. This is because that test attempts to add an invalid (empty) revoked certificate structure, which can't be serialized to ASN.1 (which is how the OpenSSL `dup` functions work). I've altered the test to set a serial and revocation date, which makes it a valid structure and will pass. This should be fine since the current behavior won't actually work if you try to create a CRL with an empty revoked (it will just fail later in the process).

### DIFF
--- a/tests/test_crypto.py
+++ b/tests/test_crypto.py
@@ -3186,7 +3186,7 @@ class CRLTests(TestCase):
         """
         crl = CRL()
         revoked = Revoked()
-        revoked.set_serial("01")
+        revoked.set_serial(b"01")
         revoked.set_rev_date(b"20160310020145Z")
         crl.add_revoked(revoked=revoked)
         self.assertTrue(isinstance(crl.get_revoked()[0], Revoked))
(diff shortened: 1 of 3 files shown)


tests/test_crypto.py
```diff
@@ -3186,6 +3186,8 @@ def test_add_revoked_keyword(self):
         """
         crl = CRL()
         revoked = Revoked()
+        revoked.set_serial("01")
```
alex: Needs a `b` prefix

Lukasa:
LGTM.

hynek:
Uh codecov seems confused at least on my phone.  It's showing me the patch coverage for the dsa/gc changes.  

Since you have zero branches there, I'm gonna ignore it.

hynek:
I’ve double checked, the lines that has been changed are being exercised.

Issue #436, opened 2016-03-11
set_cipher_list tests break on modern OpenSSL


hynek:
Because we try to set it to `EXP-RC4-MD5` that leads to a `Error: [('SSL routines', 'SSL_CTX_set_cipher_list', 'no cipher match')]`

It’s trivial to fix but I want to delay it until #422 is merged that introduces py.test based test classes for `test_ssl`.

alex:
This is fixed.

PR #437, opened 2016-03-11
Fix set_cipher_list on modern OpenSSL

hynek:
So I’ve changed my mind on #436 and did it right away while forward-porting the useful changes from #422 so they don’t linger around and we have to do without them.  The warnings change is btw for deprecation warning checks in py.test.

### DIFF
--- a/src/OpenSSL/SSL.py
+++ b/src/OpenSSL/SSL.py
@@ -5,7 +5,6 @@ from itertools import count, chain
 from weakref import WeakValueDictionary
 from errno import errorcode
 
-from six import text_type as _text_type
 from six import binary_type as _binary_type
 from six import integer_types as integer_types
 from six import int2byte, indexbytes
@@ -15,6 +14,7 @@ from OpenSSL._util import (
     lib as _lib,
     exception_from_error_queue as _exception_from_error_queue,
     native as _native,
+    make_assert as _make_assert,
     text_to_bytes_and_warn as _text_to_bytes_and_warn,
     path_string as _path_string,
     UNSPECIFIED as _UNSPECIFIED,
@@ -148,6 +148,7 @@ class Error(Exception):
 
 
 _raise_current_error = partial(_exception_from_error_queue, Error)
+_openssl_assert = _make_assert(Error)
 
 
 class WantReadError(Error):
@@ -441,7 +442,7 @@ class Session(object):
 
 class Context(object):
     """
-    :py:obj:`OpenSSL.SSL.Context` instances define the parameters for setting
+    :class:`OpenSSL.SSL.Context` instances define the parameters for setting
     up new SSL connections.
     """
     _methods = {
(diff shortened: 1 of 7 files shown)


src/OpenSSL/SSL.py
```diff
@@ -808,20 +809,22 @@ def set_tmp_ecdh(self, curve):
...
+        :param bytes cipher_list: An OpenSSL cipher string.
         :return: None
         """
-        if isinstance(cipher_list, _text_type):
-            cipher_list = cipher_list.encode("ascii")
+        cipher_list = _text_to_bytes_and_warn("cipher_list", cipher_list)
 
         if not isinstance(cipher_list, bytes):
-            raise TypeError("cipher_list must be bytes or unicode")
+            raise TypeError("cipher_list must be a bytes string.")
```
Lukasa: nit: s/bytes/byte/

Take it or leave it.

tests/test_ssl.py
```diff
@@ -347,6 +347,44 @@ def test_SSLeay_version(self):
...
+
+
+class TestContext(object):
+    @pytest.mark.parametrize("cipher_string", [
+        b"hello world:AES128-SHA",
+        u"hello world:AES128-SHA",
+    ])
+    def test_set_cipher_list(self, context, cipher_string):
+        """
+        :meth:`Context.set_cipher_list` accepts both :py:obj:`bytes` naming the
```
Lukasa: This says "both bytes", which implies that the next bit would be "and unicode strings`. Did I misunderstand this?

Lukasa:
I have two small notes, both to do with docstrings. Otherwise, this LGTM.

hynek:
I’ve fixed the nits.

Since this the assert is a @reaperhulk / @alex concept, I’d like the feedback of either too.

src/OpenSSL/SSL.py
```diff
@@ -808,20 +809,22 @@ def set_tmp_ecdh(self, curve):
...
-        :param cipher_list: A cipher list, see ciphers(1)
+        See the OpenSSL manual for more information (e.g.
+        :manpage:`ciphers(1)`).
+
+        :param bytes cipher_list: An OpenSSL cipher string.
         :return: None
         """
-        if isinstance(cipher_list, _text_type):
-            cipher_list = cipher_list.encode("ascii")
+        cipher_list = _text_to_bytes_and_warn("cipher_list", cipher_list)
```
reaperhulk: The usage is correct here but I can't help but feel like text_to_bytes_and_warn should raise a TypeError if it's not text or bytes so that we don't need the isinstance check that follows this. That should be in a separate PR though I guess since there are presumably numerous callsites for this function.

src/OpenSSL/SSL.py
```diff
@@ -808,20 +809,22 @@ def set_tmp_ecdh(self, curve):
...
 
         if not isinstance(cipher_list, bytes):
-            raise TypeError("cipher_list must be bytes or unicode")
+            raise TypeError("cipher_list must be a byte string.")
 
-        result = _lib.SSL_CTX_set_cipher_list(self._context, cipher_list)
-        if not result:
-            _raise_current_error()
+        _openssl_assert(
+            _lib.SSL_CTX_set_cipher_list(self._context, cipher_list)
```
reaperhulk: I'd suggest making this `_lib.SSL_CTX_set_cipher_list(self._context, cipher_list) == 1`. That way the value that goes into _openssl_assert is a boolean. In many cases the expected value from an openssl call is a 0 for success so establishing a convention where you compare it to the expected value is a good idea. :)

src/OpenSSL/SSL.py
hynek: Well usually we don’t check the types of our arguments since a TypeError would bubble up from CFFI anyway.  I guess pyOpenSSL is full of things that only make sense through the C-API lens.

src/OpenSSL/SSL.py
hynek: Agreed…do you think it would be better to make the assert function more explicit and do a `ok is not True`?  Looks a bit fragile to me tho.

reaperhulk: `ok is not True` is worth a shot IMO. We're not doing it in cryptography but maybe we should.

reaperhulk:
LGTM :)

PR #440, opened 2016-03-13
Polish up examples (somewhat)

### LINKED ISSUE #438: Examples should be at least linked in the docs
I had trouble with using docs at https://pyopenssl.readthedocs.org/en/latest/ without seeing any code examples.

Only later on I discovered https://github.com/pyca/pyopenssl/tree/master/examples .

There should be at least a link to this in the docs, in my opinion.

hynek:
- Mention them in the docs (arguably a bit hamfistedly).
- Make the README an RST.
- Make them pass flake8 and add flake8 to tox.ini

They should all be rewritten and made Python 3-friendly but that's out
of scope here.

Fixes #438.


reaperhulk:
LGTM

PR #443, opened 2016-03-13
De-Sphinx CHANGELOG & add rel info to description

hynek:
These changes give us a very nice release overview for `long_description` and the changelog is nicely readable from GitHub.


reaperhulk:
LGTM

PR #446, opened 2016-03-14
Add smoke tests for urllib3.

Lukasa:
@hynek is worried about breaking urllib3 with changes, so he wants some smoke tests for urllib3. This adds them to the Travis environment.

Because urllib3 doesn't install its test suites when pip installed, we need to do a git clone of the repository. This means that this test failing can potentially severely dirty the user's working copy, so it's deliberately excluded from the Travis envlist.

### DIFF
--- a/.travis/install_urllib3.sh
+++ b/.travis/install_urllib3.sh
@@ -3,6 +3,6 @@
 set -e
 set -x
 
-git clone https://github.com/shazow/urllib3.git
+git clone --depth 1 https://github.com/shazow/urllib3.git
 pip install -r ./urllib3/dev-requirements.txt
 pip install ./urllib3[socks]
--- a/.travis/install_urllib3.sh
+++ b/.travis/install_urllib3.sh
@@ -4,4 +4,5 @@ set -e
 set -x
 
 git clone https://github.com/shazow/urllib3.git
+pip install -r ./urllib3/dev-requirements.txt
 pip install ./urllib3[socks]
--- a/.travis/install_urllib3.sh
+++ b/.travis/install_urllib3.sh
@@ -1,0 +1,7 @@
+#!/bin/bash
+
+set -e
+set -x
+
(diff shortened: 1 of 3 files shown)


reaperhulk:
I assume you mean tox envlist and not travis since you definitely added it to travis :smiley:

Lukasa:
Ok, so I've got tests running. We do have some breakage. Part of that is that PyOpenSSL is raising new warnings that urllib3 is not expecting: we should fix that on urllib3's end. Some other bits are a bit more surprising, but I just realised that I maybe know where they're coming from.

hynek:
Sounds like it was already worth it. :)

Lukasa:
The failing builds should be fixed when shazow/urllib3#813 is merged to master.

hynek:
Let's whitelist /bin/rm to get rid of this warning: https://travis-ci.org/pyca/pyopenssl/jobs/115940344#L339

Lukasa:
Hmm, that test failure is unrelated, but concerning.

hynek:
Yeah that failure is a really weird flake I have seen only on OS X so far.  I tried to step through it and it indeed never reaches the callback.  @reaperhulk any idea about it?  https://travis-ci.org/pyca/pyopenssl/jobs/115946570  It’s indeed scary but I’m not sure if it’s a pyOpenSSL issue.

.travis/install_urllib3.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+set -e
+set -x
+
+git clone https://github.com/shazow/urllib3.git
```
hynek: Please use `--depth 1` we don’t need the whole history.

hynek:
FTR I’ve opened #447 for the test flake mentioned.

.travis/install_urllib3.sh
Lukasa: ACK

PR #449, opened 2016-03-16
Remove code for pre-cryptography 1.2

alex:


alex:
I don't really understand what codecov is saying here.

hynek:
Codecov became very unreliable lately. :(

Since you removed code lines and didn’t touch tests, I’m gonna ignore it.

PR #451, opened 2016-03-17
switch to OPENSSL_free

reaperhulk:
CRYPTO_free's function signature is changing in 1.1.0 and OPENSSL_free does the same thing for this purpose.


hynek:
Huh, you can choose how to free memory? :)

PR #454, opened 2016-03-19
Prepare 16.0.0

hynek:
It’s time.


Lukasa:
Hooray! Make it so, @hynek.

PR #460, opened 2016-03-29
Raise NotImplementedError when SNI not present.

### LINKED ISSUE #458: SSL_set_tlsext_host_name may not be present
https://github.com/pyca/pyopenssl/blob/master/src/OpenSSL/SSL.py#L1239 assumes the binding is available, but < 1.0.0 that is not true. We're dropping support for 0.9.8 in cryptography shortly, but we should probably fix this for pyOpenSSL 16.1 anyway.

This manifests as an AttributeError when using requests + pyOpenSSL in CentOS 5. /cc @Lukasa

Lukasa:
Resolves #458.

@hynek, as discussed, this currently adds no test coverage for this new code. Do you want to see that added?

### DIFF
--- a/tests/test_ssl.py
+++ b/tests/test_ssl.py
@@ -3874,8 +3874,8 @@ class TestRequires(object):
             results.append(True)
             return True
 
-        assert inner()
-        assert results == [True]
+        assert inner() is True
+        assert [True] == results
 
     def test_unavailable(self):
         """
@@ -3894,7 +3894,7 @@ class TestRequires(object):
             inner()
 
         assert "Error text" in str(e.value)
-        assert not results
+        assert results == []
 
 
 if __name__ == '__main__':
--- a/tests/test_ssl.py
+++ b/tests/test_ssl.py
@@ -3856,10 +3856,10 @@ class InfoConstantTests(TestCase):
             self.assertTrue(isinstance(const, int))
 
 
-class TestRequires(TestCase):
+class TestRequires(object):
     """
     Tests for the decorator factory used to conditionally raise
-    NotImplementedError when older OpenSSL's are used.
+    NotImplementedError when older OpenSSLs are used.
     """
     def test_available(self):
         """
@@ -3874,8 +3874,8 @@ class TestRequires(TestCase):
             results.append(True)
             return True
 
-        self.assertTrue(inner())
-        self.assertEqual(results, [True])
+        assert inner()
+        assert results == [True]
 
     def test_unavailable(self):
         """
@@ -3890,8 +3890,11 @@ class TestRequires(TestCase):
             results.append(True)
             return True
 
-        self.assertRaises(NotImplementedError, inner)
-        self.assertFalse(results)
+        with pytest.raises(NotImplementedError) as e:
+            inner()
+
+        assert "Error text" in str(e.value)
+        assert not results
 
 
 if __name__ == '__main__':
--- a/tests/test_ssl.py
+++ b/tests/test_ssl.py
@@ -55,6 +55,7 @@ from OpenSSL.SSL import (
     Error, SysCallError, WantReadError, WantWriteError, ZeroReturnError)
 from OpenSSL.SSL import (
     Context, ContextType, Session, Connection, ConnectionType, SSLeay_version)
+from OpenSSL.SSL import _make_requires
 
 from OpenSSL._util import lib as _lib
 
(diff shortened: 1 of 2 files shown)


hynek:
Does this actually fix urllib3 or does it just break in a…more obvious way?

Lukasa:
It breaks it in a more obvious way. =)

I think dealing with this problem is not PyOpenSSL's job: there's really nothing it can do but fail appropriately. But I think we should object properly: `AttributeError` is a weird thing to raise here.

hynek:
OK…so…I really don’t care enough.  Sorry. :)

If urllib3 has to catch an error anyways, they can just as well catch this AttributeError and introspect.  And while it it’s perfect, the error kind of makes sense: a method is missing. :)

cryptography is dropping 0.9.8 with its next release, it doesn’t make any sense to add all this code to pyOpenSSL, just to rip it out again later this year.  Agreed @reaperhulk ?

***

While looking at it I realized tho is that our `_requires_alpn` adds unnecessary runtime overhead.

IMHO it should be more like:

```python
def _requires_alpn(func):
   if not haz_alpn:
        @wraps(func)
        def explode(*args, **kw):
            raise NotImplementedErrror(yaddayadda)
        return explode

    return func
```

right?

reaperhulk:
We are dropping 0.9.8 in cryptography 1.4, but with a flag to turn it back on. Our assumption is that this will be okay and that in 1.5 we'll completely remove all support.

hynek:
blergh ugh.

so Cory, please make the decorator like I outlined above and add some monkey patched tests after you’ve tested it actually works. *sigh*

you also have my permission to fix _requires_alpn while doing so because it’s totally related. ;)

hynek:
Maybe we should even create a generic solution since such problems will haunt us forever…

```python
_requires_sni = _make_requires(_lib.Cryptography_HAS_TLSEXT_HOSTNAME, "SNI not available")
```

maybe?

(n.b. I wouldn’t make assumptions *why* it’s not available because it might just as well be another asinine packaging problem)

Lukasa:
Yeah, I think that's a reasonable approach.

Lukasa:
@hynek Considering that this new code detects the features at import time, what's your suggested method of testing the SNI stuff (other than being hopeful that the fact that it uses identical code to the NPN/ALPN side means it works)?

hynek:
I’d say test it explicitly.  i.e.

```
class TestRequires(object):
    def test_met():
        pass
    def test_not_met():
        pass
```

Then just apply the function within the test methods.

Lukasa:
What are you talking about codecov? If I hit all of the diff, how can I possibly have reduced coverage?

hynek:
codecov is really unreliable lately :(((

tests/test_ssl.py
```diff
@@ -3855,5 +3856,43 @@ def test_integers(self):
             self.assertTrue(isinstance(const, int))
 
 
+class TestRequires(TestCase):
+    """
+    Tests for the decorator factory used to conditionally raise
+    NotImplementedError when older OpenSSL's are used.
```
hynek: dat apostrophe :)

tests/test_ssl.py
```diff
@@ -3855,5 +3856,43 @@ def test_integers(self):
             self.assertTrue(isinstance(const, int))
 
 
+class TestRequires(TestCase):
```
hynek: you wrote it as Java tests just to upset me, didn’t you? :)

Lukasa:
Regardless, @hynek, this should be R4R now.

Lukasa:
Oh wait, you did review it. NVM, more patches!

tests/test_ssl.py
```diff
@@ -3874,8 +3874,8 @@ def inner():
             results.append(True)
             return True
 
-        self.assertTrue(inner())
-        self.assertEqual(results, [True])
+        assert inner()
+        assert results == [True]
```
hynek: we do expected first…makes them nicely aligned

tests/test_ssl.py
```diff
@@ -3890,8 +3890,11 @@ def inner():
             results.append(True)
             return True
 
-        self.assertRaises(NotImplementedError, inner)
-        self.assertFalse(results)
+        with pytest.raises(NotImplementedError) as e:
+            inner()
+
+        assert "Error text" in str(e.value)
+        assert not results
```
hynek: not is a bit too vague…[] == results is what you mean, right?

tests/test_ssl.py
```diff
@@ -3874,8 +3874,8 @@ def inner():
             results.append(True)
             return True
 
-        self.assertTrue(inner())
-        self.assertEqual(results, [True])
+        assert inner()
```
hynek: `is True` is nicer and more explicit

hynek:
Thanks.  It’s stupid that the coverage check fails because the coverage decreased because we have less LoC…

PR #466, opened 2016-04-28
Fix RTD links

hynek:


Lukasa:
Go home @codecov-io, you're drunk.

PR #474, opened 2016-06-02
We have always been at war with easy to read code.

alex:
aka EVERYBODY GET READY FOR OPENSSL 1.1.0

### DIFF
--- a/src/OpenSSL/SSL.py
+++ b/src/OpenSSL/SSL.py
@@ -1637,11 +1637,14 @@ class Connection(object):
 
         :return: A string representing the state
         """
-        if self._ssl.session == _ffi.NULL:
+        session = _lib.SSL_get_session(self._ssl)
+        if session == _ffi.NULL:
             return None
-        return _ffi.buffer(
-            self._ssl.s3.server_random,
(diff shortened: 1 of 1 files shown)


src/OpenSSL/SSL.py
```diff
@@ -1637,35 +1637,46 @@ def server_random(self):
...
         """
-        if self._ssl.session == _ffi.NULL:
+        session = _lib.SSL_get_session(self._ssl)
+        if session == _ffi.NULL:
             return None
-        return _ffi.buffer(
-            self._ssl.s3.server_random,
-            _lib.SSL3_RANDOM_SIZE)[:]
+        length = _lib.SSL_get_server_random(self._ssl, _ffi.NULL, 0)
+        assert length > 0
```
hynek: what about these asserts?

alex: What about it? AFAIK, they're impossible to hit.

hynek: I have mixed feelings about adding asserts to code since they get optimized out.  Didn’t we just remove all kinds of asserts from cryptography?

PR #476, opened 2016-06-02
X509 sigalg is now opaque

alex:


reaperhulk:
What do we want to do here? This depends on a binding that isn't in a released version so it won't pass CI (although you can see it passes the cryptography master tests)

alex:
@reaperhulk cryptography release? Is there harm in doing a small handful of these at the sprints?

alex:
Probably better would be for me to aggregate my changes a bit more and see if there's other crap we need.

PR #477, opened 2016-06-02
Make most things opaque

alex:


Lukasa:
LGTM, @reaperhulk should review too.

PR #481, opened 2016-06-03
fix a small bug with load_publickey

reaperhulk:


tests/test_crypto.py
```diff
@@ -2427,6 +2427,17 @@ def _runopenssl(pem, *args):
     return output
 
 
+class TestFunctions(object):
+    """
+    py.test-based tests for the free functions in the crypto module.
+
+    If possible, add new tests here.
+    """
+    def test_load_publickey(self):
```
hynek: docstring a la https://jml.io/pages/test-docstrings.html pls :)

tests/test_crypto.py
```diff
@@ -2427,6 +2427,17 @@ def _runopenssl(pem, *args):
...
+        key = load_publickey(FILETYPE_PEM, cleartextPublicKeyPEM)
+        assert key._only_public is True
```
hynek: maybe add some more asserts while at it?

reaperhulk: There are quite a few tests that assert on other things elsewhere. This is really just to verify the only_public is being set properly. I'll update the name of the test and add the docstring now though. (Alternately we could move this into another test, I have no strong opinion)

PR #496, opened 2016-07-03
Fix memory leak in OpenSSL.crypto.dump_privatekey

### LINKED ISSUE #494: Memory leak in OpenSSL.crypto.dump_privatekey with FILETYPE_TEXT
Code to demonstrate:
```py
import OpenSSL.crypto


key = OpenSSL.crypto.PKey()
key.generate_key(OpenSSL.crypto.TYPE_RSA, 64)
data = OpenSSL.crypto.dump_privatekey(OpenSSL.crypto.FILETYPE_ASN1, key)

while True:
    key = OpenSSL.crypto.load_privatekey(OpenSSL.crypto.FILETYPE_ASN1, data)
    OpenSSL.crypto.dump_privatekey(OpenSSL.crypto.FILETYPE_TEXT, key)
```

The problem is here: https://github.com/pyca/pyopenssl/blob/master/src/OpenSSL/crypto.py#L1735-L1738

hynek:
Fixes #494

### DIFF
--- a/src/OpenSSL/crypto.py
+++ b/src/OpenSSL/crypto.py
@@ -1734,9 +1734,11 @@ def dump_privatekey(type, pkey, cipher=None, passphrase=None):
     elif type == FILETYPE_ASN1:
         result_code = _lib.i2d_PrivateKey_bio(bio, pkey._pkey)
     elif type == FILETYPE_TEXT:
-        rsa = _lib.EVP_PKEY_get1_RSA(pkey._pkey)
+        rsa = _ffi.gc(
+            _lib.EVP_PKEY_get1_RSA(pkey._pkey),
+            _lib.RSA_free
+        )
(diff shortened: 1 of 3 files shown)


src/OpenSSL/crypto.py
```diff
@@ -1735,14 +1736,13 @@ def dump_privatekey(type, pkey, cipher=None, passphrase=None):
     elif type == FILETYPE_TEXT:
         rsa = _lib.EVP_PKEY_get1_RSA(pkey._pkey)
         result_code = _lib.RSA_print(bio, rsa, 0)
-        # TODO RSA_free(rsa)?
+        _ffi.gc(rsa, _lib.RSA_free)
```
alex: This should be `rsa = _ffi.gc(rsa, _lib.RSA_free)` and it should be on the line right below `rsa = _lib.EVP_PKEY...`

PR #498, opened 2016-07-03
Replace several untested paths with _openssl_assert

alex:


src/OpenSSL/crypto.py
```diff
@@ -741,9 +739,7 @@ def __str__(self):
 
         bio = _new_mem_buf()
         print_result = _lib.X509V3_EXT_print(bio, self._extension, 0, 0)
-        if not print_result:
-            # TODO: This is untested.
-            _raise_current_error()
+        _openssl_assert(print_result)
```
hynek: what does X509V3_EXT_print actually return?  _openssl_assert does an `is` comparison so this check seems shaky.

alex: They were all broken in that regard, fixed.

tests/test_crypto.py
```diff
@@ -2827,6 +2827,13 @@ def test_load_pkcs7_data_invalid(self):
         """
         self.assertRaises(Error, load_pkcs7_data, FILETYPE_PEM, b"foo")
 
+    def test_load_pkcs7_type_invalid(self):
+        """
+        If the type paased to :obj:`load_pkcs7_data`, :obj:`ValueError` is
```
hynek: `paased`

tests/test_crypto.py
```diff
@@ -2827,6 +2827,13 @@ def test_load_pkcs7_data_invalid(self):
...
+        raised.
+        """
+        self.assertRaises(ValueError, load_pkcs7_data, object(), b"foo")
```
hynek: `with pytest.raises` pls

PR #503, opened 2016-07-04
Write a few more tests

alex:


tests/test_crypto.py
```diff
@@ -971,6 +971,8 @@ def test_attributes(self):
         self.assertEqual(name.commonName, "quux")
         self.assertEqual(name.CN, "quux")
 
+        assert name.OU is None
+
```
Lukasa: Boy, this looks weird alongside the unittest style assertions in the rest of this test. I'm not sure what @hynek's guidance is here so I won't merge this until he sees it.

Lukasa:
Ok, this LGTM except for a note about test style that I'd need @hynek's opinion on. Waiting for him before merging. =)

tests/test_crypto.py
hynek: it would be nice if you could transform the rest too.  and add some empty lines to separate the blocks while doing so. :)

but i don’t want to see new code with old asserts :)

hynek:
I would *prefer* if the rest gets py.test-ified but if @alex says nope, you can merge it anyway.

alex:
@hynek done.

PR #508, opened 2016-07-31
coverage 4.2 doesn't need the clean step anymore

hynek:


tox.ini
```diff
@@ -73,11 +73,6 @@ basepython = python2.7
 commands =
      sphinx-build -W -b html doc doc/_build/html
 
-[testenv:coverage-clean]
-deps = coverage
-skip_install = true
-commands = coverage erase
-
 [testenv:coverage-report]
 deps = coverage
```
alex: Should this have `>=4.2` as well?

PR #517, opened 2016-08-24
Avoid pytest 3 warning

hynek:
Also kind of a way to force a fresh Travis run before starting the release process.


alex:
@hynek `tox.ini` should be updated with the min-version if you want to do that.

PR #539, opened 2016-09-11
empty commit to test codecov

alex:


alex:
Codecov appears to be fixed now

PR #557, opened 2016-10-16
Ensure the LICENSE file is part of our wheel

hynek:
Also add some comments the other options.

Reference for `force-manifest` claim: https://github.com/python/cpython/blob/c06449043d29c0a5b21e949e38df8ea614e1577b/Lib/distutils/command/sdist.py#L65-L67

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -4,12 +4,20 @@ strict = true
 testpaths = tests
 
 [sdist]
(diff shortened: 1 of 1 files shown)


JensTimmerman:
I'm sorry, but I'm not a reviewer for this repository.

Issue #340, opened 2015-09-05
Switch test style from unittest to py.test


alex:
This will make them easier to read, give better error messages, and let us delete a bunch of code.

hynek:
👍

Maybe we should document it in the contribution guide that people should generally write py.test-style tests for new code?

alexwlchan:
I am doing some work on this in my fork.

hynek:
Done! 🎉

PR #560, opened 2016-10-20
Migrate a large number of tests from unittest to pytest style

alexwlchan:
This mega-patch converts all the tests (as best I can) to be pytest-style, not unittest style.

The magic of regexes, eh? 😄 

Nowhere near ready for review and/or merging, I just want to start getting CI running against it. I suspect it probably needs breaking up into smaller, individual patches, before being actually reviewed.

Addresses #340.

### DIFF
--- a/tests/test_crypto.py
+++ b/tests/test_crypto.py
@@ -1165,7 +1165,7 @@ class X509NameTests(TestCase):
         b = self._x509name(CN="foo")
         self.assertEqual(a.hash(), b.hash())
         a.CN = "bar"
-        self.assertNotEqual(a.hash(), b.hash())
+        assert a.hash() != b.hash()
 
     def test_der(self):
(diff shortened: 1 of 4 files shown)


alexwlchan:
I’m going to bed now, but at this point I’ve removed the vast majority of unittest-style asserts (the two holdouts are the 500-odd occurrences of `assertEqual()` and `assertRaises()`). I *think* all the tests are still passing, but I’ll find out in the morning if I screwed up.

I had some weird interactions between `@pytest.mark.parametrize` and the `utils.TestCase` class – do all the tests need to subclass from this and/or `unittest.TestCase`, or can they subclass from `object`? (Or some other small class that has a `mktemp()` method.)

alexwlchan:
I’ll open a fresh PR.

PR #561, opened 2016-10-21
Umbrella PR for migrating tests from unittest to pytest style

alexwlchan:
This patch converts most of the tests to use pytest-style, not unittest style, as per #340.

Each individual commit expunges precisely one type of assert helper from the test suite, and (locally at least) is passing tests and listing correctly. Each one could be merged individually, and I can raise them as separate PRs if that would be easier to handle.

(With apologies for thinking out loud last night all over #560)

### DIFF
--- a/tests/test_crypto.py
+++ b/tests/test_crypto.py
@@ -1165,7 +1165,7 @@ class X509NameTests(TestCase):
         b = self._x509name(CN="foo")
         self.assertEqual(a.hash(), b.hash())
         a.CN = "bar"
-        self.assertNotEqual(a.hash(), b.hash())
+        assert a.hash() != b.hash()
 
     def test_der(self):
         """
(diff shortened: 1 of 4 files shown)


hynek:
Holy 💩!  So you’re getting an A for effort, but we gotta cut this down a little bit. :D

You’ve noticed yourself, that there’s a problem if a test class subclasses `TestCase` vs. pytest features.

I’d suggest you start with simple classes that can be safely moved to subclass object and work ourselves up to the bigger challenges.  Be free to open one PR per self-contained class.  They are free. :)

alexwlchan:
Heh, thought that might be the case. I wasn’t expecting to merge this in one go, I just wanted to get CI running against it. I’ll start breaking it up.

/me proceeds to open ALL THE PRS 😄

hynek:
Please don’t have more than 2 open at a time pls tho. :)

hynek:
Done! 🎉

PR #562, opened 2016-10-23
Convert test_util to use pytest-style tests

alexwlchan:
Fix up the assert helpers, subclass from `object` rather than
`TestCase`, and then I had to rename the class -- pytest doesn't
pick it up with the original name.

Addresses #340.

---

Part of breaking up #561.

### DIFF
--- a/tests/test_util.py
+++ b/tests/test_util.py
@@ -1,9 +1,9 @@
 from OpenSSL._util import exception_from_error_queue, lib
 
-from .util import TestCase
+import pytest
 
 
-class ErrorTests(TestCase):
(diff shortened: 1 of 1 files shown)


alexwlchan:
Looks like the Xcode build just stalled – on every other job, the tests are passing (and they’re passing locally). I suspect if somebody re-runs the Travis job, it will all go green.

Lukasa:
Restarted build

alexwlchan:
Yay, green CI! 🍏

hynek:
Thanks!

PR #563, opened 2016-10-23
Convert test_rand to use pytest-style tests

alexwlchan:
Fix up the assert helpers, subclass form `object` rather than test case,
and use parametrization where appropriate.

One helper method on the original `TestCase` was the ability to create
temporary directories that were cleaned up at the end of the test --
now we use a pytest fixture instead: http://doc.pytest.org/en/latest/tmpdir.html

Addresses #340.

---

Part of breaking up #561.

### DIFF
--- a/src/OpenSSL/rand.py
+++ b/src/OpenSSL/rand.py
@@ -107,7 +107,7 @@ def status():
     """
     Check whether the PRNG has been seeded with enough data.
 
-    :return: :obj:`1` if the PRNG is seeded enough, :obj:`0` otherwise.
+    :return: 1 if the PRNG is seeded enough, 0 otherwise.
     """
     return _lib.RAND_status()
 
--- a/src/OpenSSL/rand.py
+++ b/src/OpenSSL/rand.py
@@ -107,7 +107,7 @@ def status():
     """
     Check whether the PRNG has been seeded with enough data.
 
-    :return: :obj:`True` if the PRNG is seeded enough, :obj:`False` otherwise.
+    :return: :obj:`1` if the PRNG is seeded enough, :obj:`0` otherwise.
     """
     return _lib.RAND_status()
 
--- a/tests/test_rand.py
+++ b/tests/test_rand.py
@@ -106,13 +106,13 @@ class TestRand(object):
 
     def test_status(self):
         """
-        `OpenSSL.rand.status` returns `True` if the PRNG has
-        sufficient entropy, `False` otherwise.
+        `OpenSSL.rand.status` returns `1` if the PRNG has
+        sufficient entropy, `0` otherwise.
         """
         # It's hard to know what it is actually going to return.  Different
         # OpenSSL random engines decide differently whether they have enough
         # entropy or not.
-        assert rand.status() in (True, False)
+        assert rand.status() in (0, 1)
 
     @pytest.mark.parametrize('args', [
         (b"foo", 255),
--- a/tests/test_rand.py
+++ b/tests/test_rand.py
@@ -13,25 +13,31 @@ import pytest
 
 from OpenSSL import rand
 
-from .util import NON_ASCII, TestCase
+from .util import NON_ASCII
 
 
-class RandTests(TestCase):
-    def test_bytes_wrong_args(self):
+class TestRand(object):
+
+    @pytest.mark.parametrize('args', [
+        (),
+        (None),
+        (3, None)
+    ])
+    def test_bytes_wrong_args(self, args):
         """
         :py:obj:`OpenSSL.rand.bytes` raises :py:obj:`TypeError` if called with
         the wrong number of arguments or with a non-:py:obj:`int` argument.
         """
-        self.assertRaises(TypeError, rand.bytes)
-        self.assertRaises(TypeError, rand.bytes, None)
-        self.assertRaises(TypeError, rand.bytes, 3, None)
+        with pytest.raises(TypeError):
+            rand.bytes(*args)
 
-    def test_insufficientMemory(self):
+    def test_insufficient_memory(self):
         """
         :py:obj:`OpenSSL.rand.bytes` raises :py:obj:`MemoryError` if more bytes
         are requested than will fit in memory.
         """
-        self.assertRaises(MemoryError, rand.bytes, sys.maxsize)
+        with pytest.raises(MemoryError):
+            rand.bytes(sys.maxsize)
 
     def test_bytes(self):
         """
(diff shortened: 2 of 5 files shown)


tests/test_rand.py
```diff
@@ -13,25 +13,31 @@
 
 from OpenSSL import rand
 
-from .util import NON_ASCII, TestCase
+from .util import NON_ASCII, tmpfile
```
hynek: Um why do they fail?  Shouldn’t pytest be able to pick up fixtures globally?

alexwlchan: Not sure -- I'd expect pytest to pick up failures globally as well, but if you look [at the first Travis run](https://travis-ci.org/pyca/pyopenssl/jobs/169862020), it doesn't seem to have noticed. 

```
________________ ERROR at setup of TestRand.test_unicode_paths _________________
file /Users/travis/build/pyca/pyopenssl/tests/test_rand.py, line 219
      def test_unicode_paths(self, tmpfile):
E       fixture 'tmpfile' not found
>       available fixtures: cache, capfd, capsys, doctest_namespace, monkeypatch, pytestconfig, record_xml_property, recwarn, tmpdir, tmpdir_factory
>       use 'pytest --fixtures [testpath]' for help on them.
```

I did have `from .util import NON_ASCII` in that version of the code, but it didn't seem to pick them up. With this line in, the tests pass, but then flake8 gets upset.

(I haven't tried running `pytest --fixtures [testpath]` yet, probably should give that another go.)

hynek: This reminds me that fixtures should generally go into `contest.py`.  Is it possible, that pytest has different discovery mechanisms?

alexwlchan:
And another CI failure, again in a test that I haven't touched. 😕

Looks like a timing window -- is datetime mocked out in these tests?

hynek:
There is a known race condition that’s sadly depending on calling external programs.  It takes more than monkey patching datetime to fix it.

What bothers me more is that [py27 with xcode8 and OpenSSL 1.0.2](https://travis-ci.org/pyca/pyopenssl/jobs/170100896) is repeatedly hanging.  I’ve already restarted it once.

Maybe try running pytest with `-v`?  ISTM it’s hanging in the same test.

alexwlchan:
And naturally, the problem disappears with verbose reporting.

I don't think this was the issue, but I did spot this at the bottom of the [py27 with Xcode 8 and OpenSSL 1.0.2](https://travis-ci.org/pyca/pyopenssl/jobs/170117744) log:

```
/Users/travis/build.sh: line 109: shell_session_update: command not found
```

hynek:
Ugh, *of course* it passed this time! :|

tests/test_rand.py
```diff
@@ -96,63 +112,77 @@ def test_status(self):
         # It's hard to know what it is actually going to return.  Different
         # OpenSSL random engines decide differently whether they have enough
         # entropy or not.
-        self.assertTrue(rand.status() in (1, 2))
+        assert rand.status() in (True, False)
```
hynek: Um do both versions pass? :|

alexwlchan:
> I don't think this was the issue, but I did spot this at the bottom of the [py27 with Xcode 8 and OpenSSL&nbsp;1.0.2](https://travis-ci.org/pyca/pyopenssl/jobs/170117744) log:

> ```/Users/travis/build.sh: line 109: shell_session_update: command not found```

Bit weird, that script doesn't exist in the pyOpenSSL codebase (or indeed, any mention of it).

@hynek Can you trigger another test run?

tests/test_rand.py
```diff
@@ -96,63 +112,77 @@ def test_status(self):
...
-                     (None,),
-                     ("foo", None),
-                     (None, 3),
-                     ("foo", 3, None)]:
-            with pytest.raises(TypeError):
-                rand.egd(*args)
+        with pytest.raises(TypeError):
+            rand.egd(*args)
 
     def test_cleanup_wrong_args(self):
```
hynek: Uh this whole test can go.  All those args tests are from times when pyOpenSSL was written in C but many of them don’t make a lot of sense anymore.  This one particularly.   We don’t have to test whether a function that takes no arguments raises a TypeError.  Let’s leave that to Python core. ;)

hynek:
I’ve restarted that unfortunate job.

Another wish:  could you please rid the test file of those `:py:obj:` prefixes?  Don’t don’t make any sense in general (should `:meth:`, `:func:` etc)  but in test files they don’t make any sense in particular since those comments never get rendered anyway so let’s save some (untestable) noise.

hynek:
The shell_session_update message seems to be benign btw: https://github.com/travis-ci/travis-ci/issues/6307

tests/test_rand.py
alexwlchan: I assume it was passing before, seems to be passing now.

The [pyOpenSSL docs say](http://www.pyopenssl.org/en/stable/api/rand.html#OpenSSL.rand.status) that it returns `True`/`False`:

> Check whether the PRNG has been seeded with enough data.
> 
> Returns: `True` if the PRNG is seeded enough, `False` otherwise.

although looking [at the pyOpenSSL code](https://github.com/alexwlchan/pyopenssl/blob/master/src/OpenSSL/rand.py#L106), it seems like that error is being emitted somewhere from OpenSSL itself.

The [OpenSSL manpage](https://www.openssl.org/docs/manmaster/crypto/RAND_add.html) says:

> RAND_status() returns 1 if the PRNG has been seeded with enough data, 0 otherwise.

So I think this is several bugs in one:

* The old test was checking the wrong values
* The documentation is wrong -- `rand.status()` actually returns integers, not booleans:

    ```pycon
    >>> from OpenSSL import rand
    >>> rand.status()
    1
    ```

    It should be cast to a boolean before returning.

* We don't have a test when it doesn't have enough entropy -- all the systems have plenty of entropy, so this never returns `0`. Is there a way to drain its entropy pool for testing purposes?

hynek:
(and let’s keep the `-v` for a while; our output isn’t as pathologic as from the Twisted test suite :D)

tests/test_rand.py
alexwlchan: Take out all the args tests, or just this one?

hynek: I’d say take out those whose result depends solely on the argument count.

src/OpenSSL/rand.py
```diff
@@ -107,7 +107,7 @@ def status():
     """
     Check whether the PRNG has been seeded with enough data.
 
-    :return: :obj:`True` if the PRNG is seeded enough, :obj:`False` otherwise.
```
hynek: you forgot the `:obj:`s here :)

alexwlchan:
Docstrings tidied.

hynek:
Thanks!

PR #564, opened 2016-10-23
Convert X509ExtTests to use pytest-style tests

alexwlchan:
Tricky part here was the `setUp()` and `tearDown()` methods, but
these can be replaced by a pytest fixture, and in practice, they
weren't being used by many test cases anyway.

Addresses #340.

---

### DIFF
--- a/tests/test_crypto.py
+++ b/tests/test_crypto.py
@@ -46,7 +46,7 @@ from OpenSSL.crypto import (
 from OpenSSL._util import native, lib
 
 from .util import (
-    EqualityTestsMixin, TestCase, WARNING_TYPE_EXPECTED
+    EqualityTestsMixin, is_consistent_type, TestCase, WARNING_TYPE_EXPECTED
 )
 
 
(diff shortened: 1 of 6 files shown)


tests/test_crypto.py
```diff
@@ -588,55 +578,50 @@ def test_construction(self):
         :py:class:`X509ExtensionType` instance.
         """
         basic = X509Extension(b'basicConstraints', True, b'CA:true')
-        self.assertTrue(
-            isinstance(basic, X509ExtensionType),
-            "%r is of type %r, should be %r" % (
-                basic, type(basic), X509ExtensionType))
+        assert isinstance(basic, X509ExtensionType)
```
alexwlchan: Note: I dropped the custom error message because pytest is smart enough to include information about the actual type of `basic` without it; for example:

```
__________________________________ TestX509Ext.test_construction ___________________________________

self = <tests.test_crypto.TestX509Ext object at 0x105e0d5c0>

    def test_construction(self):
        """
            :py:class:`X509Extension` accepts an extension type name, a critical
            flag, and an extension value and returns an
            :py:class:`X509ExtensionType` instance.
            """
        basic = X509Extension(b'basicConstraints', True, b'CA:true')
>       assert isinstance(basic, ValueError)
E       assert False
E        +  where False = isinstance(<OpenSSL.crypto.X509Extension object at 0x105e0d780>, ValueError)

tests/test_crypto.py:590: AssertionError
```

alexwlchan:
And this failure is in an area of code that I, as far as I can tell, had nothing to do with.

CI really doesn’t like me today, huh?

exarkun:
Can I opt out of mention bot?  I don't want to hear about pyOpenSSL being converted to py.test. :( :( :(

hynek:
Technically [you are](https://github.com/pyca/pyopenssl/blob/master/.mention-bot), but sadly mention-bot doesn’t seem to be good software (unlike pytest!).  I wanted to kill it anyways.

tests/test_crypto.py
```diff
@@ -524,60 +524,49 @@ def normalize_privatekey_pem(pem):
 """
 
 
-class X509ExtTests(TestCase):
+@pytest.fixture
+def x509_data():
```
hynek: Please add a docstring explaining the fixture.

alexwlchan:
Rebased against master and review markups made.

hynek:
Merry xmas, you’ve pulled the “new flake8 came out and is breaking the build of your PR and now you have to fix it”-card :D

tests/test_crypto.py
```diff
@@ -524,235 +524,223 @@ def normalize_privatekey_pem(pem):
...
+            issuer=x509)
+        x509.add_extensions([ext2])
+        x509.sign(pkey, 'sha1')
+        text = dump_certificate(FILETYPE_TEXT, x509)
+        assert b'X509v3 Authority Key Identifier:' in text
+        assert b'DirName:/CN=Yoda root CA' in text
 
     def test_missing_issuer(self):
         """
         If an extension requires an issue and the :py:data:`issuer` parameter
```
hynek: you missed this `:py:data`

tests/util.py
```diff
@@ -305,10 +305,26 @@ def assertConsistentType(self, theType, name, *constructionArgs):
...
-        self.assertEqual(theType.__name__, name)
-        self.assertTrue(isinstance(theType, type))
-        instance = theType(*constructionArgs)
-        self.assertIdentical(type(instance), theType)
+        assert is_consistent_type(theType, name, *constructionArgs)
+
+
+def is_consistent_type(theType, name, *constructionArgs):
+    """
+    Perform various assertions about `theType` to ensure that it is a
```
hynek: `*theType*` please since it’s an argument.

hynek:
you missed one flake8; should be ready for merge afterwards.

PR #565, opened 2016-10-24
Convert ServerNameCallbackTests to use pytest-style tests

alexwlchan:
As well as pytest-ifying up the tests, remove a few redundant tests
and tidy up docstrings as per feedback in #563.

Addresses #340.
---


hynek:
Can you get rid of the mixin?

alexwlchan:
Mixin expunged from this class; I'll be working through the other test classes so I'll unpick the mixin's influence from them as I get to them.

hynek:
thanks!

PR #567, opened 2016-10-24
Add the ability to set a custom verification time on X509Store

tsileo:
This PR add a `set_time` methods on `OpenSSL.crypto.X509Store`.

This new methods:
- create a `X509_VERIFY_PARAM`
- "set" the given time (a `datetime.datetime` instance)
- and finally "set" the param on the `X509Store`

One things I'm worried about is if the `X509_VERIFY_PARAM` should be "gc"ed manually? or it will be cleaned when the `X509Store` will be "gc"ed?

I chose to implement the method on `X509Store` because it's more convenient to set the time only once on the store, and then verify multiple certificates (instead of having to set it in each `X509StoreContext`).

Thanks!

### DIFF
--- a/src/OpenSSL/crypto.py
+++ b/src/OpenSSL/crypto.py
@@ -1524,7 +1524,7 @@ class X509Store(object):
         .. versionadded: 16.3.0
 
         :param datetime vfy_time: The verification time to set on this store.
-        :return: ``None`` if the verification time were successfully set.
+        :return: ``None`` if the verification time was successfully set.
         """
         param = _lib.X509_VERIFY_PARAM_new()
         param = _ffi.gc(param, _lib.X509_VERIFY_PARAM_free)
--- a/src/OpenSSL/crypto.py
+++ b/src/OpenSSL/crypto.py
@@ -1514,12 +1514,12 @@ class X509Store(object):
 
         :param int flags: The verification flags to set on this store.
             See :class:`X509StoreFlags` for available constants.
-        :return: ``None`` if the verification flags were successfully set.
         """
         _openssl_assert(_lib.X509_STORE_set_flags(self._store, flags) != 0)
 
     def set_time(self, vfy_time):
-        """Set verification time to this store.
+        """
+        Set verification time to this store.
 
         .. versionadded: 16.3.0
 
--- a/src/OpenSSL/crypto.py
+++ b/src/OpenSSL/crypto.py
@@ -1520,9 +1520,15 @@ class X509Store(object):
 
     def set_time(self, vfy_time):
         """
-        Set verification time to this store.
+        Set the time against which the certificates are verified.
 
-        .. versionadded: 16.3.0
+        Normally the current time is used.
+
+        .. note::
+
+          For example, you can determine if a certificate was valid at a given time.
+
+        .. versionadded:: 16.3.0
 
         :param datetime vfy_time: The verification time to set on this store.
         :return: ``None`` if the verification time was successfully set.
--- a/src/OpenSSL/crypto.py
+++ b/src/OpenSSL/crypto.py
@@ -1514,6 +1514,7 @@ class X509Store(object):
 
         :param int flags: The verification flags to set on this store.
             See :class:`X509StoreFlags` for available constants.
+        :return: ``None`` if the verification flags were successfully set.
         """
         _openssl_assert(_lib.X509_STORE_set_flags(self._store, flags) != 0)
 
--- a/src/OpenSSL/crypto.py
+++ b/src/OpenSSL/crypto.py
@@ -1518,6 +1518,18 @@ class X509Store(object):
         """
         _openssl_assert(_lib.X509_STORE_set_flags(self._store, flags) != 0)
 
+    def set_time(self, vfy_time):
+        """Set verification time to this store.
+
+        .. versionadded: 16.3.0
+
+        :param datetime vfy_time: The verification time to set on this store.
+        :return: ``None`` if the verification time were successfully set.
+        """
+        vfy_param = _lib.X509_VERIFY_PARAM_new()
+        _lib.X509_VERIFY_PARAM_set_time(vfy_param, int(vfy_time.strftime('%s')))
+        _openssl_assert(_lib.X509_STORE_set1_param(self._store, vfy_param) != 0)
+
 
 X509StoreType = X509Store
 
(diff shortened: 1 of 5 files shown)


src/OpenSSL/crypto.py
```diff
@@ -1518,6 +1518,18 @@ def set_flags(self, flags):
...
 
+    def set_time(self, vfy_time):
+        """Set verification time to this store.
+
+        .. versionadded: 16.3.0
+
+        :param datetime vfy_time: The verification time to set on this store.
+        :return: ``None`` if the verification time were successfully set.
+        """
+        param = _lib.X509_VERIFY_PARAM_new()
```
reaperhulk: You'll need to add a `param = _ffi.gc(param, X509_VERIFY_PARAM_free)` to this. `X509_STORE_set1_param` increfs it so without a corresponding free to decref after creation it will leak.

tsileo: Thanks for your feedback!

When I try:
```python
from OpenSSL._util import lib
lib.X509_VERIFY_PARAM_free
```
I got this exception on my system:  `AttributeError: 'module' object has no attribute 'X509_VERIFY_PARAM_free'`

It looks like `X509_VERIFY_PARAM_free` is not exported, I should do a PR on cryptography to export it? or something is wrong on my side?

Thanks!

reaperhulk: Nope, you're quite right it isn't currently exported in cryptography. Please do submit a PR there!

tsileo:
I've made the change, I'll push once the cryptography PR has been merged (to make the CI works).

reaperhulk:
We won't be able to merge this until after we do a cryptography release with the new binding (and this PR will then need to be updated to reflect the new minimum version required), but now we can see if it works against cryptography master :)

reaperhulk:
I am unfamiliar with the state of docs for X509Store, does this function need to be documented outside of the docstring @hynek?

src/OpenSSL/crypto.py
```diff
@@ -1518,6 +1518,20 @@ def set_flags(self, flags):
         """
         _openssl_assert(_lib.X509_STORE_set_flags(self._store, flags) != 0)
 
+    def set_time(self, vfy_time):
+        """Set verification time to this store.
+
+        .. versionadded: 16.3.0
+
+        :param datetime vfy_time: The verification time to set on this store.
+        :return: ``None`` if the verification time were successfully set.
```
reaperhulk: s/were/was

hynek:
On phone but it's gonna need an automethod line somewhere.

tsileo:
@reaperhulk no problem I get it. I'll follow the "Sixteenth Release" milestone of cryptography.

Let me know what changes are needed for the doc, I checked `doc/api/crypto.rst` and with `autoclass:: X509Store` and `:members:` I though it would be OK with doc, unless it need to be displayed somewhere else.

tsileo:
The build fails because it's using `cryptography==1.5.3`, it passes locally with cryptography master.

I will wait till the new cryptography release.

reaperhulk:
@tsileo cryptography 1.6 is now out :)

tsileo:
Thanks for the heads-up!

How should I update the doc to reflect the new minimum version required fro cryptography? I can't find any example.

And I think it should be OK for the doc due to the `autoclass::X509Store` but let me know.

Any other change I need to make?

reaperhulk:
Probably something like "bumped minimum cryptography version to 1.6", but hynek may have a preference. Technically this PR is fine @hynek so if you've got any stylistic nits speak now :)

hynek:
Before we get into stylistic questions, the new minimum version has to be applied to setup.py/tox.ini.

tox.ini
```diff
@@ -9,7 +9,7 @@ deps =
     coverage>=4.2
     pytest>=3.0.1
     cryptographyMaster: git+https://github.com/pyca/cryptography.git
-    cryptographyMinimum: cryptography<1.4
+    cryptographyMinimum: cryptography==1.6
```
hynek: This got to be `<1.7` so we always test against the latest 1.6 release.

src/OpenSSL/crypto.py
```diff
@@ -1518,6 +1518,20 @@ def set_flags(self, flags):
         """
         _openssl_assert(_lib.X509_STORE_set_flags(self._store, flags) != 0)
 
+    def set_time(self, vfy_time):
+        """Set verification time to this store.
```
hynek: `"""` on own lines please.

src/OpenSSL/crypto.py
```diff
@@ -1518,6 +1518,20 @@ def set_flags(self, flags):
...
+
+        .. versionadded: 16.3.0
+
+        :param datetime vfy_time: The verification time to set on this store.
+        :return: ``None`` if the verification time was successfully set.
```
hynek: I don’t think the qualification is necessary.  I *always* returns None.  It just may raise an error.

tests/test_crypto.py
```diff
@@ -3740,6 +3740,27 @@ def test_modification_pre_verify(self):
         store_ctx.set_store(store_good)
         self.assertEqual(store_ctx.verify_certificate(), None)
 
+    def test_verify_with_time(self):
+        """
+        :func:`verify_certificate` raises error when the verification time is
```
hynek: I know the code base is inconsistent in this regard, but please don't use type prefixes like `:func:` in test doc strings.  They are never rendered and there’s no way to verify them at all.

tsileo:
Done! Let me know if there's something else I need to do :)

hynek:
As for documenting minimum cryptography version I’m not sure either.  Until now, we’ve never really documented it in the CHANGELOG…

tsileo:
@hynek I just pushed the style tweaks and updated the minimum cryptography requirements.

src/OpenSSL/crypto.py
```diff
@@ -1514,10 +1514,24 @@ def set_flags(self, flags):
 
         :param int flags: The verification flags to set on this store.
             See :class:`X509StoreFlags` for available constants.
-        :return: ``None`` if the verification flags were successfully set.
```
hynek: heh you removed the wrong one but I see the qualification is everywhere in this class so just re-add this one and let the other be for now

tsileo: Oops, I was checking if it was present  in other methods. I just re-added it.

hynek:
flake8 is failing

tsileo:
Sorry about that, I fixed the line too long (my config is set at 120 :p)

hynek:
thanks!

PR #569, opened 2016-11-05
Convert NextProtoNegotiationTests to use pytest-style tests

alexwlchan:

### DIFF
--- a/tests/test_ssl.py
+++ b/tests/test_ssl.py
@@ -1688,193 +1688,170 @@ class TestNextProtoNegotiation(object):
     """
     Test for Next Protocol Negotiation in PyOpenSSL.
     """
-    if _lib.Cryptography_HAS_NEXTPROTONEG:
-        def test_npn_success(self):
-            """
-            Tests that clients and servers that agree on the negotiated next
-            protocol can correct establish a connection, and that the agreed
-            protocol is reported by the connections.
-            """
-            advertise_args = []
-            select_args = []
+    def test_npn_success(self):
+        """
+        Tests that clients and servers that agree on the negotiated next
+        protocol can correct establish a connection, and that the agreed
+        protocol is reported by the connections.
+        """
+        advertise_args = []
+        select_args = []
 
-            def advertise(conn):
-                advertise_args.append((conn,))
-                return [b'http/1.1', b'spdy/2']
+        def advertise(conn):
+            advertise_args.append((conn,))
+            return [b'http/1.1', b'spdy/2']
 
-            def select(conn, options):
-                select_args.append((conn, options))
-                return b'spdy/2'
+        def select(conn, options):
+            select_args.append((conn, options))
+            return b'spdy/2'
 
-            server_context = Context(TLSv1_METHOD)
-            server_context.set_npn_advertise_callback(advertise)
+        server_context = Context(TLSv1_METHOD)
+        server_context.set_npn_advertise_callback(advertise)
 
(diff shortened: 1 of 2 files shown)


hynek:
So the twisted test suite fails because of https://twistedmatrix.com/trac/ticket/8876

`03:54:14 <hawkowl> hynek use python -m twisted.trial, sorry`

Feel free to just fix it here too.

alexwlchan:
Ah yes, I remember this from hyper. Should be fixed.

hynek:
Codecov makes it look like fail_,rthkds might be empty? 🤔

alexwlchan:
That’s a bit sad. 😕

But I think Codecov actually thinks we never come down this test at all – if it was just that `fail_methods` were empty, we’d still hit the `conn = Connection(context)` on line 1873, which it seems to think is missed. That whole test is under an `else:` clause – is something causing Codecov not to get data from this branch?

hynek:
Can you `XXX`-verify we do?

(= put a XXX there and look if it breaks the test)

alexwlchan:
So indeed, we never seem to be exploring that branch in CI. The only failure goes nowhere nearthe different exception

But looking at the coverage in master, that doesn't seem to be a regression in this particular patch: https://codecov.io/gh/pyca/pyopenssl/src/master/tests/test_ssl.py#L1860

I have slightly increased the line count of the test suite (`self.assertRaises(...)` is 1 compared to `with pytest.raises(Error): ...` which is 2. Could that make a difference?

hynek:
What I’m confused about is the *test* that is supposedly not ran (and was before – according to codecov).  If the test code is dead for whatever reason, we need to fix it somehow.  For all we know it’s all syntax errors that don’t get executed. :)

alexwlchan:
Hmm. Maybe I’m misreading Codecov’s interface, but it looks to me like this test isn’t being run on master:

https://codecov.io/gh/pyca/pyopenssl/src/master/tests/test_ssl.py#L1858

I’m not an OpenSSL expert, but it looks like this test only runs if your OpenSSL has NPN support. Presumably all the OpenSSL builds on Travis include it as a desirable feature. Is there any way to disable NPN even if your OpenSSL supports it?

hynek:
Ugh yeah…hm…we dropped 0.9.7… @Lukasa do we need to support NPN at all or do all OpenSSL versions we support support ALPN now? (that was a lot of supports)

Lukasa:
NPN was added in OpenSSL 1.0.1, but ALPN was added in OpenSSL 1.0.2, so OpenSSL 1.0.1 is the category of OpenSSLs that supports NPN but not ALPN. Unfortunately that's also code for "the most widely deployed version of OpenSSL".

alexwlchan:
I’m confused. Where does ALPN fit into this? The failure is in a branch of code that’s supposed to run if we have no NPN support – ALPN is a different test class.

If all the OpenSSLs we support include NPN, then this test case not running is to be expected – because it only runs if you don’t have NPN. In which case the way to fix the coverage issue is to delete this branch.

Does that sound right?

hynek:
I’m a bit confused because I thought that OpenSSL 1.1 dropped NPN but it seems I might be wrong. *sigh*  Weren’t there some plans like that Cory?

Lukasa:
My recollection is that *Chrome* dropped NPN support, not OpenSSL. I doubt OpenSSL will drop NPN support for a long time to come because there's really no need to.

hynek:
OK kill it off.

PR #570, opened 2016-11-10
Convert ALPN and SessionTests to use pytest-style tests

alexwlchan:
I did the ALPN tests, and then `SessionTests` was so short I just did that as well.

Throwing away `test_construction_wrong_args` is based on this comment: https://github.com/pyca/pyopenssl/pull/563#discussion_r84685308

Addresses #340.

---


hynek:
Aren’t we rollin’!

PR #574, opened 2016-11-19
Convert a couple of classes in `test_crypto.py` to use pytest-style.

alexwlchan:

### DIFF
--- a/tests/test_crypto.py
+++ b/tests/test_crypto.py
@@ -662,7 +662,7 @@ class TestX509Ext(object):
             b'subjectKeyIdentifier', False, b'hash', subject=x509)
         # assert False, 'point 2'
         x509.add_extensions([ext3])
-        assert False, 'point 3'
+        # assert False, 'point 3'
         x509.sign(pkey, 'sha1')
         assert False, 'point 4'
         text = dump_certificate(FILETYPE_TEXT, x509)
--- a/tests/test_crypto.py
+++ b/tests/test_crypto.py
@@ -3115,9 +3115,12 @@ class TestRevoked(object):
         with pytest.raises(ValueError):
             revoked.set_serial(b'pqrst')
 
-        for bad_serial in [100, 1, None, ""]:
+        with pytest.raises(TypeError):
+            revoked.set_serial(100)
+
+        for bad_serial in [1, None, ""]:
             with pytest.raises(TypeError):
-                revoked.set_serial(bad_serial)
+                revoked.get_serial(bad_serial)
 
     def test_date(self):
         """
--- a/tests/test_crypto.py
+++ b/tests/test_crypto.py
@@ -2963,12 +2963,12 @@ class LoadCertificateTests(TestCase):
 
 class TestPKCS7(object):
     """
-    Tests for `PKCS7Type`.
+    Tests for `PKCS7`.
     """
 
     def test_type(self):
         """
-        `PKCS7Type` is a type object.
+        `PKCS7` is a type object.
         """
         assert isinstance(PKCS7Type, type)
         assert PKCS7Type.__name__ == 'PKCS7'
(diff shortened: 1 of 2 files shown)


hynek:
🔥🔥🔥 CI 🔥🔥🔥

I've also noticed it used the Type alias class all around which is rather deprecated.  Would you mind replace it with the real thing?  Feel free to add a Test they're the same. 

(Sorry for Hand-wavy instructions, im on my phone and thank you very much for keeping to put all the work into this thankless task! ❤️)

alexwlchan:
@hynek I think I’ve done what you mean. I’ve also turned on a test that `PKCS7` and `PKCS7Type` are identical, which was present in the original code but commented out because it didn’t work – except, now it does. 😄

hynek:
the fuck is wrong with python 2 builds? 😭

tests/test_crypto.py
```diff
@@ -3136,111 +3085,84 @@ def test_ignores_unsupported_revoked_cert_extension_set_new_reason(self):
...
+            revoked.set_serial(b'pqrst')
 
-        self.assertRaises(ValueError, revoked.set_serial, b'pqrst')
-        self.assertRaises(TypeError, revoked.set_serial, 100)
-        self.assertRaises(TypeError, revoked.get_serial, 1)
-        self.assertRaises(TypeError, revoked.get_serial, None)
-        self.assertRaises(TypeError, revoked.get_serial, "")
+        for bad_serial in [100, 1, None, ""]:
+            with pytest.raises(TypeError):
+                revoked.set_serial(bad_serial)
```
alexwlchan: Aha, I see the problem. `set_serial` != `get_serial`, and this was subtle enough that I did it wrong.

I bet this is what’s breaking the Python 2 builds.

hynek: byte/unicode funsies just keep giving!

alexwlchan:
Booooo CI timeouts. 🔥

hynek:
I did restart some already.  The hell is going on. :(

hynek:
fwiw it seems like it always hangs in tests/test_crypto.py::TestX509Ext::test_subject ?

alexwlchan:
😕 

I'll give it a spin locally when I get home.

alexwlchan:
Hmm. Nothing obviously jumps out in that test as being slow (and I haven’t changed any of the code in it). Running the 2.7 tests locally all work fine (and fast). Maybe an intermittent issue with the Travis builds?

😕

hynek:
Still hanging…maybe try to comment it out?  It might be some weird side-effect from another test.  Or  add a bunch of `print` statements.

alexwlchan:
Okay, so it looks like there definitely is a bug in that test. 😕 

@hynek Just FYI, I’m taking a holiday next week, so all programming work is about to grind to a halt. I’ll look at this and #576 when I get back, 5th December-ish.

hynek:
Enjoy your holiday!

hynek:
(as for debugging this: i’d just say add print statements after each line so we have a clue what hangs ¯\_(ツ)_/¯)

alexwlchan:
So the line it’s hanging on is this one:

```python
assert b'X509v3 Subject Key Identifier:' in text
```

which I suppose makes some sort of sense, as that is the only line in the test that I’ve changed. But `test_unused_subject` has a near identical line which works fine. Err… wut?

hynek:
maybe try something like `print(type(text), text)`? and add `-k test_subject -s` to tox.ini?

alexwlchan:
(Apologies for the commit spam, but I am unable to reproduce this locally, so I’m having to rely on Travis for debugging.)

alexwlchan:
I’m really quite stumped by this. I think we must be hitting some very obscure timing window – almost any change to `test_subject()` unblocks it, as does deleting prior tests. And I can’t reproduce it locally.

I’m going to ask the Travis folks if I can get hold of an exact copy of their container image for a better chance of a local repro.

alexwlchan:
@hynek For the purposes of getting this merged (and because I don’t want to go too much further down this rabbit hole), an alternative fix would be to add:

```python
assert isinstance(text, str)
```

to `test_subject()`. For whatever reason, that tips the timing in such a way that the test no longer hangs. It’s not a technically incorrect assert to write, just not a very useful one.

Would you be okay if I included that in the patch, just for the sake of fixing Travis?

hynek:
Yeah along with a comment.  I wonder if we could involve someone from pytest to help us out here.  Maybe he assert rewriting might be at fault too?  I don't think the timing is predictable enough to hang reproducibly.

alexwlchan:
@hynek I’ve sunk enough time on this rabbit hole already, I’d rather not drag in somebody from pytest and/or Travis for such an obscure window condition without evidence of a more serious underlying problem.

I’m going to close this unmerged and keep the branch around, as the commits and build links might be vaguely useful for somebody if we ever run into this again. (Also the comment history is utterly trashed.) I’ll open a new PR with a tidied patch and the commented fix.

Issue #485, opened 2016-06-04
Document last 0.9.8 version tuple


hynek:
cryptography 1.4 dropped OpenSSL 0.9.8 🎉.  While nobody should ever use it, we should document the working combination to save us some issues/complaining.

hynek:
I keep having to look it up myself.  There needs to be some easy to find place that’s unequivocal.

PR #576, opened 2016-11-24
Convert the elliptic curve tests to be pytest-style

alexwlchan:

### DIFF
--- a/tests/test_crypto.py
+++ b/tests/test_crypto.py
@@ -3864,52 +3864,52 @@ class SignVerifyTests(TestCase):
         sign(priv_key, content, "sha1")
 
 
-class EllipticCurveTests(TestCase):
+class TestEllipticCurve(object):
     """
-    Tests for :py:class:`_EllipticCurve`, :py:obj:`get_elliptic_curve`, and
-    :py:obj:`get_elliptic_curves`.
+    Tests for `_EllipticCurve`, `get_elliptic_curve`, and
+    `get_elliptic_curves`.
     """
 
     def test_set(self):
         """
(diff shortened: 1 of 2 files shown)


alexwlchan:
So looking at Codecov, it looks like the bug is that we call `get_elliptic_curves()` and always get something back. In particular, the `test_some_curves()` test never executes the code path where we don't get some curves from OpenSSL.

This is [the same on master](https://codecov.io/gh/pyca/pyopenssl/src/e15e60a587efcddd61e391a8e5917291c98d554e/tests/test_crypto.py#L3889).

Has the supported OpenSSL version advanced to the point where every version has elliptic curve support?

hynek:
A clear case of #485.

We don’t support 0.9.8 anymore but there’s no docs on when the EC crap has been added.

1.0.0!?  1.0.1?

Sadly, cryptography supports “OpenSSL 1.0.0-fips” but we don’t test against it so we don’t know if it even works.  Technically we should add Travis builders for all supported OpenSSLs. :'(

@reaperhulk do you know more?

reaperhulk:
EC support is present even in 0.9.8. Depends what feature you specifically care about. Also, we drop 1.0.0 in cryptography 1.7, which is coming out next week. Probably Monday.

hynek:
OK, @alexwlchan feel free to kill off no-EC-present code.  Next pyOpenSSL will be 1.0.1+ only.

alexwlchan:
> feel free to kill off no-EC-present code

Done.

hynek:
seems like we got rid of an import dep (flake8)

hynek:
Thanks once again!

PR #579, opened 2016-11-28
We should consider actually testing Twisted.

Lukasa:
Currently we're not actually running any Twisted tests: see [this build](https://travis-ci.org/pyca/pyopenssl/jobs/179282980). This patch actually runs them.


hawkowl:
lol the py35 tests failed because of a timestamp

hynek:
yeah that’s my favorite race condition.  one day i’ll fix it and save 93758903427589043h of restarting jobs all the time.

PR #580, opened 2016-11-29
Add support for OCSP stapling.

Lukasa:
This PR adds a collection of functions that add support for server and client side OCSP stapling.

In the spirit of PyOpenSSL's "thin wrapper" policy, these functions specifically do not provide any *validation* of stapled OCSP assertions: that's officially Someone Else's Problem. However, they do provide a mechanism whereby servers may staple OCSP assertions if they desire, and clients may retrieve those stapled OCSP assertions.

**This PR cannot be merged yet, it will fail tests against most cryptography versions**. We will need to wait until a new cryptography release is cut (specifically 1.7) and then make that our minimum supported cryptography version in order to properly enable this functionality.

### DIFF
--- a/src/OpenSSL/SSL.py
+++ b/src/OpenSSL/SSL.py
@@ -368,9 +368,24 @@ class _ALPNSelectHelper(_CallbackExceptionHelper):
         )
 
 
-class _OCSPCallbackHelper(_CallbackExceptionHelper):
+class _OCSPServerCallbackHelper(_CallbackExceptionHelper):
     """
-    Wrap a callback such that it can be used as an OCSP callback.
+    Wrap a callback such that it can be used as an OCSP callback for the server
+    side.
+
+    Annoyingly, OpenSSL defines one OCSP callback but uses it in two different
+    ways. For servers, that callback is expected to retrieve some OCSP data and
+    hand it to OpenSSL, and may return only SSL_TLSEXT_ERR_OK,
+    SSL_TLSEXT_ERR_FATAL, and SSL_TLSEXT_ERR_NOACK. For clients, that callback
+    is expected to check the OCSP data, and returns a negative value on error,
+    0 if the response is not acceptable, or positive if it is. These are
+    mutually exclusive return code behaviours, and they mean that we need two
+    helpers so that we always return an appropriate error code if the user's
+    code throws an exception.
+
+    Given that we have to have two helpers anyway, these helpers are a bit more
+    helpery than most: specifically, they hide a few more of the OpenSSL
+    functions so that the user has an easier time writing these callbacks.
     """
 
     def __init__(self, callback):
@@ -545,7 +560,7 @@ class Context(object):
         self._npn_select_callback = None
         self._alpn_select_helper = None
         self._alpn_select_callback = None
-        self._ocsp_helper = None
+        self._ocsp_server_helper = None
         self._ocsp_callback = None
         self._ocsp_data = None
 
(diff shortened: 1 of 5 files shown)


src/OpenSSL/SSL.py
```diff
@@ -368,6 +368,137 @@ def wrapper(ssl, out, outlen, in_, inlen, arg):
...
+                    data = _ffi.from_handle(cdata)
+                else:
+                    data = None
+
+                # Get the OCSP data.
+                ocsp_ptr = _ffi.new("unsigned char **")
+                ocsp_len = _lib.SSL_get_tlsext_status_ocsp_resp(ssl, ocsp_ptr)
+                if ocsp_len < 0:
+                    # No OCSP data.
+                    ocsp_data = b''
```
Lukasa: It's not clear to me whether this is the best behaviour. In particular, I wonder whether we should set `ocsp_data = None` here: this means we actively received no OCSP data, and so is technically a bit distinct from the case where the server sent zero-length OCSP data (which I think is technically possible, though isn't possible with the way I've constructed the server callback helper).

Lukasa:
I have pushed the requirement for cryptography 1.7 and merged in the current master.

hynek:
jftr i have no idea how ocsp works so y’all on your own :)

tiran:
@hynek magic! :)

I'm vaguely familiar with OCSP and can assist with review next week. Feel free to ping me if I don't respond by Jan 18th.

Lukasa:
Remember that this PR does no validation of OCSP of any kind: just exposes the raw binary data of the stapled OCSP responses.

src/OpenSSL/SSL.py
reaperhulk: Can you think of a scenario where the distinction between "no OCSP data" vs "zero length OCSP data" is meaningful?

tests/test_ssl.py
```diff
@@ -3842,3 +3842,247 @@ def inner():
...
+        ctx.use_certificate(load_certificate(FILETYPE_PEM, server_cert_pem))
+        ctx.set_ocsp_server_callback(callback, data)
+        server = Connection(ctx)
+        server.set_accept_state()
+        return server
+
+    def test_callbacks_arent_called_by_default(self):
+        """
+        If both the client and the server have registered OCSP callbacks, but
+        the client does not sent the OCSP request, neither callback gets
```
reaperhulk: does not send

src/OpenSSL/SSL.py
Lukasa: So, I double-checked the RFC: the server cannot send zero-length OCSP data. The minimum allowed length for OCSP data is 1 byte. So I think we're safe: if the server sent zero-length OCSP data then we should just act like they didn't send it at all.

Lukasa:
Ok @reaperhulk, I've responded to your feedback.

hynek:
could someone pls add a link to this pr in the changelog? thx!

Lukasa:
@hynek See #590.

abhieee23:
I have  pyopenssl v16.2.0 but it does not have the code for OCSP. The dependency, Cryptography is at version 1.8.1.
Have the changes been merged?

hynek:
It has been merged but not released yet.

abhieee23:
As far as I've seen the code and understood, It has support for stapling the OCSP data during the handshake, but what I am confused about is, does it have support for an outbound connection from server to OCSP server that gets the latest OCSP data? ie if there is any mechanism which involves server and OCSP server?

abhieee23:
As far as I've seen the code and understood, It has support for stapling the OCSP data during the handshake, but what I am confused about is, does it have support for an outbound connection from server to OCSP server that gets the latest OCSP data? ie if there is any mechanism which involves server and OCSP server?

Lukasa:
No. As originally stated in the PR, the code provides *only* support for providing stapled OCSP data. No support is provided for obtaining that OCSP data.

jszhang1985:
def zjs_ocsp_callback(conn,response,tmp):
    print 'response=%s' % response
    return True
......
ctx.set_ocsp_client_callback(zjs_ocsp_callback)
....
cnx = SSL.Connection(ctx, sock)
cnx.request_ocsp()
......


The simplified code is above this, sniffer the package, the script will send client hello with status extenssion, and get status reply from ssl server.
Now, I want to print the OCSP stapling response, but the script will print mess code,  no matter decode it with asicc or utf-8.
can anybody guard me a way to print this in human being ?  thanks

Issue #456, opened 2016-03-19
crypto.load_privatekey raises an crypto.Error without any details


adiroiban:
In 0.13 crypto.load_privatekey is raising a crypo.Error which provides some details about the error.

In master the error is empty.

Here are the updated tests 

https://github.com/chevah/pyopenssl/commit/4115b5f04ead708bc34d25324c60f6994e011f39

I found those details very useful for auditing/logging problems in production.

Thanks!

hynek:
Is this something that changed in 0.14 or 16.0?

adiroiban:
This defect was introduced in 0.14.

I ran my test on 0.14, 0.15 and 16.0.0 and it fails

The problem is that when `_raise_current_error` is called from load_privatekey  `lib.ERR_get_error()` returns 0 ... might be an issue in python cryptography

Thanks!

topnotcher:
`lib.ERR_get_error()` returns `0` because all the errors have already been read from the queue by `helper.raise_if_problem()`.

crypto.py:
```python
    if type == FILETYPE_PEM:
        evp_pkey = _lib.PEM_read_bio_PrivateKey(
            bio, _ffi.NULL, helper.callback, helper.callback_args)
        helper.raise_if_problem()  # This eats all the errors.
...
    if evp_pkey == _ffi.NULL:
        _raise_current_error()   # This raises an empty exception
```

When `helper.raise_if_problem()` does not raise an exception (i.e. there was no exception caught while calling `helper._read_passphrase()`) and `evp_pkey` is `NULL`, `exception_from_error_queue()` is called a second time (the first time being in `helper.raise_if_problem()`), and `lib.ERR_get_error()` returns `0` because all errors have already been retreived.

Issue #119, opened 2014-06-06
Empty error message when trying to load an invalid or incomplete server.pem


Julian:
Creating an empty `server.pem` followed by attempting to point pyOpenSSL at it (via `twistd -n web -p ssl:port=4443:privateKey=server.pem --path .`) produces:

```
/usr/local/Cellar/python/2.7.6/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/twisted/spread/jelly.py:93: DeprecationWarning: the sets module is deprecated
  import sets as _sets
/usr/local/Cellar/python/2.7.6/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/twisted/internet/endpoints.py:30: DeprecationWarning: twisted.internet.interfaces.IStreamClientEndpointStringParser was deprecated in Twisted 14.0.0: This interface has been superseded by IStreamClientEndpointStringParserWithReactor.
  from twisted.internet.interfaces import (
Traceback (most recent call last):
  File "/usr/local/bin/twistd", line 14, in <module>
    run()
  File "/usr/local/Cellar/python/2.7.6/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/twisted/scripts/twistd.py", line 27, in run
    app.run(runApp, ServerOptions)
  File "/usr/local/Cellar/python/2.7.6/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/twisted/application/app.py", line 642, in run
    runApp(config)
  File "/usr/local/Cellar/python/2.7.6/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/twisted/scripts/twistd.py", line 23, in runApp
    _SomeApplicationRunner(config).run()
  File "/usr/local/Cellar/python/2.7.6/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/twisted/application/app.py", line 376, in run
    self.application = self.createOrGetApplication()
  File "/usr/local/Cellar/python/2.7.6/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/twisted/application/app.py", line 436, in createOrGetApplication
    ser = plg.makeService(self.config.subOptions)
  File "/usr/local/Cellar/python/2.7.6/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/twisted/web/tap.py", line 230, in makeService
    strports.service(config['port'], site).setServiceParent(s)
[...]

PR #581, opened 2016-11-30
Prevent _PassphraseHelper.raise_if_problem() from eating exceptions

topnotcher:
When `_PassphraseHelper.raise_if_problem()` does not raise an exception, it still flushes the OpenSSL error queue, possibly causing subsequent code to `raise crypto.Error([])` as in bugs #456, #119.

This changeset modifies `_PassphraseHelper.raise_if_problem()` to flush the error queue only when it raises an exception, thus leaving the OpenSSL errors untouched when it does not raise an exception.


reaperhulk:
I have no idea why that test is failing...maybe rebase/merge forward just to make it refresh the environment?

topnotcher:
Yeah... looks like a timing issue. I rebased on current master.

PR #586, opened 2016-12-20
Convert another couple of test classes to pytest-style

alexwlchan:
Part of breaking up #561.


hynek:
Thanks once again!  Getting there, aren’t we?

PR #589, opened 2017-01-23
Convert TestContext to be pytest-style

alexwlchan:
Part of breaking up #561.

This class is a bit of a monster, so it gets its own pull request.


hynek:
thanks!

PR #593, opened 2017-01-25
Convert the remaining tests in test_crypto.py to be pytest-style

alexwlchan:
Part of breaking up #561.


hynek:
Thanks once again!  Are we done yet? ;)

PR #594, opened 2017-01-25
Convert the rest of TestConnection to be pytest-style

alexwlchan:


hynek:
any reason to not make the _LoopbackMixin fixtures?

PR #599, opened 2017-01-30
Rip out the last vestages of unittest from the test suite

alexwlchan:
This patch changes the remaining tests in `test_ssl.py` to be pytest style, and throws away the old unittest-based test helpers. When this is merged, I believe #340 will finally be done.

Part of breaking up #561.


hynek:
According to codecov, we can kill off LoopbackMixin for good?

alexwlchan:
> According to codecov, we can kill off LoopbackMixin for good?

Oops, forgot to push that. Fixed.

PR #601, opened 2017-01-30
Fix up some pytest-related stuff

hynek:
No need to mention is specifically anymore.

### DIFF
--- a/tests/test_ssl.py
+++ b/tests/test_ssl.py
@@ -2040,8 +2040,8 @@ class TestConnection(object):
 
     def test_peek(self):
         """
-        `Connection.recv` peeks into the connection if `socket.MSG_PEEK`
-        is passed.
+        `Connection.recv` peeks into the connection if `socket.MSG_PEEK` is
+        passed.
         """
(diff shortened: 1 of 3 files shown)


tests/test_ssl.py
```diff
@@ -2040,8 +2040,8 @@ def test_pending(self):
 
     def test_peek(self):
         """
-        `Connection.recv` peeks into the connection if `socket.MSG_PEEK`
-        is passed.
+        `Connection.recv` peeks into the connection if `socket.MSG_PEEK` is
```
markrwilliams: Formatting change that snuck in?

tests/test_ssl.py
```diff
@@ -2051,8 +2051,8 @@ def test_peek(self):
 
     def test_connect_wrong_args(self):
         """
-        `Connection.connect` raises `TypeError` if called with
-        a non-address argument.
+        `Connection.connect` raises `TypeError` if called with a non-address
```
markrwilliams: Formatting change that snuck in?

PR #602, opened 2017-01-30
Assorted cheap coverage wins

### LINKED ISSUE #592: Patch up easy TypeError-based coverage gaps
If you look through the coverage report, there are a bunch of paths of the following form which aren’t tested:

```python
def foo(arg):
    if not isinstance(arg, desired_type):
        raise TypeError('arg is not of desired_type!')
    ...
```

There’s a small but easy coverage win to add tests for all of these paths.

If nobody else tackles this first, I’ll do it once I’ve finished #340.

alexwlchan:
And just when @hynek thought they could take the night off…

This patch contains:

* A few extra tests that patch some basic holes in the coverage, mostly unchecked `TypeError` for bad arguments.
* A few extra pytest-related tidies.
* We have bits of the test suite we explicitly expect not to run – rather than flag them as missing coverage, use `# pragma: no cover` to ignore them in the Coveralls report.

Resolves #592.

### DIFF
--- a/tests/test_crypto.py
+++ b/tests/test_crypto.py
@@ -1189,8 +1189,7 @@ class _PKeyInteractionTestsMixin:
 
     def signable(self):
         """
-        Return something with a `set_pubkey`, `set_pubkey`,
-        and `sign` method.
+        Return something with a `set_pubkey`, `set_pubkey`, and `sign` method.
         """
         raise NotImplementedError()
 
@@ -1924,6 +1923,16 @@ class TestX509Store(object):
         store = X509Store()
         store.add_cert(cert)
 
+    @pytest.mark.parametrize('cert', [None, 1.0, 'cert', object()])
+    def test_add_cert_wrong_args(self, cert):
+        """
+        `X509Store.add_cert` raises `TypeError` if passed a non-X509 object
+        as its first argument.
+        """
+        store = X509Store()
+        with pytest.raises(TypeError):
+            store.add_cert(cert)
+
     def test_add_cert_rejects_duplicate(self):
         """
         `X509Store.add_cert` raises `OpenSSL.crypto.Error` if an attempt is
@@ -2553,6 +2562,16 @@ class TestFunction(object):
             load_privatekey(
                 FILETYPE_PEM, encryptedPrivateKeyPEM, lambda *args: 3)
 
+    def test_dump_privatekey_wrong_args(self):
+        """
+        `dump_privatekey` raises `TypeError` if called with a `cipher`
+        argument but no `passphrase` argument.
+        """
+        key = PKey()
+        key.generate_key(TYPE_RSA, 512)
+        with pytest.raises(TypeError):
+            dump_privatekey(FILETYPE_PEM, key, cipher=GOOD_CIPHER)
+
     def test_dump_privatekey_unknown_cipher(self):
         """
         `dump_privatekey` raises `ValueError` if called with an unrecognized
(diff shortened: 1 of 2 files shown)


alexwlchan:
After this, the biggest coverage hole is a bunch of Windows-specific code, but I don’t see how to get to that until Travis supports Windows runs.

hynek:
We could add appveyor I guess?

**£

I don't know about removing memdbg though.  It seems like nobody knows how to actually use it.  Maybe @exarkun can help us out with that?  We need to either document it or remove it.

exarkun:
`memdbg` is used just by importing it.  It's not for end users (I don't recall it being part of any package in the olden days; now it appears to be part of the test suite).  It's for maintainers trying to debug memory allocation problems - mostly leaks and double frees.  It provides information about the C and Python call stacks at the point where parts of the heap owned by OpenSSL were allocated.

It does this partly by turning some incorrect memory management operations into Python-level errors (ie, an exception for a double-free at the point of the second free).  It also provides the `heap` global that maps allocated memory to the full stack trace at the point of the allocation.

alexwlchan:
> `memdbg` is used just by importing it.

Hmm. Looking over the old code, it seems like `memdbg` got imported and incorporated as part of the `TestCase` base class that was used for tests in the unittest days; the work to switch entirely to pytest removed any uses of it.

hynek:
Can someone make it actually work?  It always fails on macOS.

alexwlchan:
The question of `memdbg` aside (I’m unlikely to have time/ability to get it working any time soon), is the rest of the patch okay? I can back out the `memdbg` deletion – would the remaining changes be mergeable?

(I’m not trying to dodge the issue, just uncouple it from what I was expecting to be an easy patch.)

reaperhulk:
Leave out the pragma: no cover as well and it should be an easy review.

alexwlchan:
> Leave out the pragma: no cover as well and it should be an easy review.

Okay, pushed a version that just has the new tests and pytest fixes.

hynek:
Hm maybe add `# pragma: nocover` (no need to add any configuration for that) to the failing test branches.

hynek:
There's a few more instances and also verify callbacks that aren't called apparently?  If it's on purpose, maybe they should be a proper function that has a pytest.fail + nocover?

hynek:
Getting there https://codecov.io/gh/pyca/pyopenssl/compare/b748099426db3963c7b06d2e0aec29e12e016c07...a3df545b5558eb2f5efdd04561f795772054c8c2/src/tests/test_ssl.py

I really would like to see 100% patch coverage before merging...

alexwlchan:
Maybe I’m just familiar with the Codecov.io output – but what part of this patch isn’t covered? 😕

hynek:
look for yellow and red.  linking to lines doesn’t seem to work. :(

but e.g. line 1353, 1379, 1499…

hynek:
Hey, any chance we can finish this up?

alexwlchan:
Sorry, I let this and #605 slide. I’ve been quite busy lately, but I’ll try to look at them again this weekend.

hynek:
NP, leave being sorry to me.

alexwlchan:
I remain somewhat flummoxed by the coverage result; I’ve done a bit of tidying and pushed a rebased version. Then I’ll take another look at the codecov results.

hynek:
It seems that most of the uncovered or partially covered lines are empty lambdas.  I’m not gonna pester you with those, I’m just a bit…flummoxed :)…with the following:

 https://codecov.io/gh/pyca/pyopenssl/pull/602/src/tests/test_ssl.py → L1528  It seems like this branch used to be green but isn’t anymore.  Did we change something, that made callback become None?

I found one bug: https://codecov.io/gh/pyca/pyopenssl/pull/602/src/tests/test_ssl.py → L2208 is fortunately wrong in a simple way.  The assertions about the raised exception have to happen *outside* the block.

After that let’s call it a day/patch.

P.S. The inability to link to lines for some reason is upsetting. :|

alexwlchan:
L1528 – no idea, doesn’t look like code I’ve touched.

L2208 – not sure why that’s gone from green-to-red with this patch, but I’ll fix it anyway.

hynek:
Thanks Alex, I’ve just YOLO’d it in.  I’m sorry this was this painful.

PR #603, opened 2017-02-01
Limit SSL_write bufsize to avoid OverflowErrors

mhils:
We got [this](https://github.com/mitmproxy/mitmproxy/issues/1961) bug report where @ohhorob triggered an OverflowError when trying to `.sendall()` more than 2,2GB of data. ¯\\\_(ツ)_/¯
```
  File "/usr/local/lib/python3.6/site-packages/mitmproxy/net/tcp.py", line 188, in write
    return self.o.sendall(v)
  File "/usr/local/lib/python3.6/site-packages/OpenSSL/SSL.py", line 1285, in sendall
    result = _lib.SSL_write(self._ssl, data + total_sent, left_to_send)
OverflowError: integer 3114393428 does not fit '32-bit int'
```
`SSL_write` only accepts an int32 as length argument, so we can't send everything at once. Here's a fix that limits the buffer size to 2**32-1 bytes to fit into an int32. Not sure how this could be tested without sending 2,2GB over the wire (that'd be an expensive test) or mocking very heavily, so this comes without tests for now. I'm happy to take suggestions on what to do about testing 😃

### DIFF
--- a/src/OpenSSL/SSL.py
+++ b/src/OpenSSL/SSL.py
@@ -1456,7 +1456,8 @@ class Connection(object):
         if not isinstance(buf, bytes):
             raise TypeError("data must be a memoryview, buffer or byte string")
 
-        result = _lib.SSL_write(self._ssl, buf, len(buf))
+        buf_len = min(len(buf), 2147483647)  # cannot send more than 2**31-1 bytes at once.
+        result = _lib.SSL_write(self._ssl, buf, buf_len)
         self._raise_ssl_error(self._ssl, result)
         return result
     write = send
@@ -1486,7 +1487,7 @@ class Connection(object):
         data = _ffi.new("char[]", buf)
 
         while left_to_send:
-            result = _lib.SSL_write(self._ssl, data + total_sent, left_to_send)
+            result = _lib.SSL_write(self._ssl, data + total_sent, min(left_to_send, 2147483647))
             self._raise_ssl_error(self._ssl, result)
             total_sent += result
             left_to_send -= result
--- a/src/OpenSSL/SSL.py
+++ b/src/OpenSSL/SSL.py
@@ -1455,9 +1455,10 @@ class Connection(object):
             buf = str(buf)
         if not isinstance(buf, bytes):
             raise TypeError("data must be a memoryview, buffer or byte string")
+        if len(buf) > 2147483647:
+            raise ValueError("Cannot send more than 2**31-1 bytes at once.")
 
-        buf_len = min(len(buf), 2147483647)  # cannot send more than 2**31-1 bytes at once.
-        result = _lib.SSL_write(self._ssl, buf, buf_len)
+        result = _lib.SSL_write(self._ssl, buf, len(buf))
         self._raise_ssl_error(self._ssl, result)
         return result
     write = send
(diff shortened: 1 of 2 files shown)


hynek:
Is it an int32 on 64 bit architectures too?

src/OpenSSL/SSL.py
```diff
@@ -1456,7 +1456,8 @@ def send(self, buf, flags=0):
         if not isinstance(buf, bytes):
             raise TypeError("data must be a memoryview, buffer or byte string")
 
-        result = _lib.SSL_write(self._ssl, buf, len(buf))
+        buf_len = min(len(buf), 2147483647)  # cannot send more than 2**31-1 bytes at once.
```
reaperhulk: This will silently truncate. I think the previous behavior of blowing up if you try to send too much here is appropriate.

src/OpenSSL/SSL.py
```diff
@@ -1486,7 +1487,7 @@ def sendall(self, buf, flags=0):
         data = _ffi.new("char[]", buf)
 
         while left_to_send:
-            result = _lib.SSL_write(self._ssl, data + total_sent, left_to_send)
+            result = _lib.SSL_write(self._ssl, data + total_sent, min(left_to_send, 2147483647))
```
reaperhulk: Looks fine, although adding a comment about the limit would be helpful.

src/OpenSSL/SSL.py
hynek: Yeah silent truncation is a no-go.  We could raise a more helpful ValueError though.

mhils: Thanks for catching this - I did not consider that SSL_MODE_ENABLE_PARTIAL_WRITE is off by default. 🙈 Will make it a ValueError.

mhils:
Sorry for the slow turnaround. I fixed the truncation issue and added a test for that behavior. :-)

src/OpenSSL/SSL.py
```diff
@@ -1486,7 +1488,13 @@ def sendall(self, buf, flags=0):
         data = _ffi.new("char[]", buf)
 
         while left_to_send:
-            result = _lib.SSL_write(self._ssl, data + total_sent, left_to_send)
```
hynek: I believe this change can be rolled back, since we bail out if left_to_send is to big, no?

tests/test_ssl.py
```diff
@@ -2662,6 +2663,12 @@ def test_result(self):
             assert isinstance(cipher, str)
 
 
+class VeryLarge(bytes):
+    """Mock object so that we don't have to allocate 2**31 bytes"""
```
hynek:
```python
"""
docstring
"""
```

please :)

tests/test_ssl.py
```diff
@@ -2725,6 +2732,16 @@ def test_short_buffer(self):
         assert count == 2
         assert client.recv(2) == b'xy'
 
+    @pytest.mark.skipif(
+        sys.maxsize < 2**31,
+        reason="sys.maxsize < 2**31 - test requires 64 bit"
+    )
+    def test_buf_too_large(self):
```
hynek: Test docstring please: https://jml.io/pages/test-docstrings.html

src/OpenSSL/SSL.py
mhils: We constrain left_to_send with `min(left_to_send, 2147483647)` so it won't bail out. With this PR, pyOpenSSL will call `send()` multiple times (with smaller buffers) if a buffer >= 2**31 is passed to `sendall()`. Does that make sense?

hynek: oh i’m sorry, i missed it’s a different method!

mhils:
\o/

Thank you guys for the _insanely_ fast turnaround. 😊

PR #612, opened 2017-04-05
Fix invalid cast from ASN1_TIME to ASN1_GENERALIZEDTIME

moriyoshi:
`X509_get_notBefore()` and `X509_get_notAfter()` return a `ASN1_TIME` structure which is not compatible with `ASN1_GENERALIZEDTIME`, but the original code recklessly casts it to `ASN1_GENERALIZEDTIME` and pass it to `ASN1_GENERALIZEDTIME_set_string()`.

LibreSSL is rigorous enough to reject such.

https://github.com/pyca/cryptography/pull/3491 needs to be applied to cryptography before the application of this patch.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -96,7 +96,7 @@ if __name__ == "__main__":
         package_dir={"": "src"},
         install_requires=[
             # Fix cryptographyMinimum in tox.ini when changing this!
-            "cryptography>=1.7",
+            "cryptography>=1.9",
             "six>=1.5.2"
         ],
     )
--- a/src/OpenSSL/crypto.py
+++ b/src/OpenSSL/crypto.py
@@ -103,7 +103,7 @@ def _set_asn1_time(boundary, when):
     """
     The the time value of an ASN1 time object.
 
-    @param boundary: An ASN1_GENERALIZEDTIME pointer (or an object safely
+    @param boundary: An ASN1_TIME pointer (or an object safely
         castable to that type) which will have its value set.
     @param when: A string representation of the desired time value.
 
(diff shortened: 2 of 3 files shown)


moriyoshi:
**UPDATE:** https://github.com/pyca/cryptography/pull/3491 was merged.

reaperhulk:
Those two functions do return an `ASN1_TIME`, but unfortunately the pyOpenSSL API here for `set_notAfter` and `set_notBefore` is documented as taking the form of a byte string generalized time. (e.g. not just `YYYYMMDDHHMMSSZ` but also `YYYYMMDDhhmmss+hhmm` or `YYYYMMDDhhmmss-hhmm`). No one should ever have been passing any form other than the first though, and after checking our primary downstreams it looks like that's not a thing anyone is doing, so let's go ahead and make the change.

moriyoshi:
> the pyOpenSSL API here for set_notAfter and set_notBefore is documented as taking the form of a byte string generalized time. (e.g. not just YYYYMMDDHHMMSSZ but also YYYYMMDDhhmmss+hhmm or YYYYMMDDhhmmss-hhmm).

It appears `ASN1_TIME_set_string()` can handle the generalized time in addition to the UTC time. 

https://github.com/openssl/openssl/blob/master/crypto/asn1/a_time.c#L120-L124

BTW, `ASN1_TIME_set_string_X509()` is supposed to be better here rather than `ASN1_TIME_set_string()`, which is a newly introduced, more strict variant of the latter:

https://github.com/openssl/openssl/commit/04e62715db684d83bffac53793ff4cfac51e047a

reaperhulk:
Interesting. Thanks for investigating that. We should revert that backwards incompatibility warning then.

PR #584, opened 2016-12-11
Document supported OpenSSL versions

hynek:
Fixes #485 

(presumes existence of cryptography 1.7)

PR #615, opened 2017-04-19
Prepare for 17.0.0

hynek:
Fix a bunch URLs and account for the non-existance of 16.3.

I’d like @alex and/or @reaperhulk to double-check the INSTALL.rst version matrix.


hynek:
oh I found https://github.com/pyca/pyopenssl/pull/584#discussion_r91856852 – let’s do this!

PR #616, opened 2017-04-20
Remove code for checking if EC support is present

alexwlchan:
@hynek We killed off the no-EC-present code in the tests in #576, because all supported OpenSSL versions have EC support. This covers the one part of PyOpenSSL itself where we were checking for EC support.


hynek:
Ha, you found the last one! :)

PR #622, opened 2017-04-25
Add support for sending bytearrays.

### LINKED ISSUE #621: bytearray is a bytes-like object
```python
In [1]: from OpenSSL import SSL

In [2]: ctx = SSL.Context(SSL.SSLv23_METHOD)

In [3]: conn = SSL.Connection(ctx, None)

In [4]: conn.set_accept_state()

In [5]: conn.send(bytearray(10))
TypeError: data must be a memoryview, buffer or byte string
```

But `bytearray`s are nice and friendly and deserve your love too!

Actually I'd expect that it should accept anything that implements the buffer interface, using [`ffi.from_buffer`](https://cffi.readthedocs.io/en/latest/ref.html#ffi-buffer-ffi-from-buffer).

Lukasa:
Resolves #621.

This patch adds support for sending bytearrays by using `ffi.from_buffer` for both bytearrays and byte strings.

In general this patch is a good idea, however, it has some caveats. Most notable is that bytearray and byte string support for `from_buffer` only landed in cffi 1.8, but our lowest supported cryptography version only requires cffi 1.4. That may make this patch unacceptable. The *minimum* we could get away with would be cffi 1.7, and only then if we add separate code paths for byte strings and byte arrays.

On the other hand, if we could go up to requiring cffi 1.10 then we could remove the conditionals for handling memoryview and buffer objects, as `ffi.from_buffer` also supports them directly.

Anyway, here's a PoC, we can discuss whether it's acceptable or not.

### DIFF
--- a/src/OpenSSL/SSL.py
+++ b/src/OpenSSL/SSL.py
@@ -5,6 +5,8 @@ from itertools import count, chain
 from weakref import WeakValueDictionary
 from errno import errorcode
 
+import cffi
+
 from six import binary_type as _binary_type
 from six import integer_types as integer_types
 from six import int2byte, indexbytes
@@ -159,6 +161,46 @@ class ZeroReturnError(Error):
 
 class SysCallError(Error):
     pass
+
+
+def _make_buffer_checker(cffi_version_tuple):
+    """
+    Returns a callable that converts a sendable buffer object into something
+    that can be sent by OpenSSL using CFFI. Has different supported types
+    depending on the CFFI version.
+
+    :param cffi_version_tuple: The version of CFFI in use, from
+        cffi.__version_info.
+    """
+    if cffi_version_tuple >= (1, 8):
+        allowed_types = (bytes, bytearray)
+        error_message = (
+            "data must be a memoryview, buffer, bytearray or byte string"
+        )
+        make_buffer = _ffi.from_buffer
+    else:
+        allowed_types = (bytes,)
+        error_message = "data must be a memoryview, buffer, or byte string"
+        make_buffer = partial(_ffi.new, "char[]")
+
+    def buffer_factory(buf):
+        """
+        Given a sendable data buffer passed into send or sendall, returns a
+        CFFI object representing it.
+
+        :param buf: The sendable data buffer the user wishes to send.
+        """
+        if isinstance(buf, _memoryview):
+            buf = buf.tobytes()
+        if isinstance(buf, _buffer):
+            buf = str(buf)
+
+        if not isinstance(buf, allowed_types):
+            raise TypeError(error_message)
+
+        return make_buffer(buf)
+
+    return buffer_factory
 
 
 class _CallbackExceptionHelper(object):
@@ -1290,6 +1332,7 @@ class Connection(object):
         self._ssl = _ffi.gc(ssl, _lib.SSL_free)
         self._context = context
         self._app_data = None
+        self._buffer_factory = _make_buffer_checker(cffi.__version_info__)
 
         # References to strings used for Next Protocol Negotiation. OpenSSL's
         # header files suggest that these might get copied at some point, but
@@ -1448,23 +1491,14 @@ class Connection(object):
         """
         # Backward compatibility
         buf = _text_to_bytes_and_warn("buf", buf)
+        size = len(buf)
 
-        if isinstance(buf, _memoryview):
-            buf = buf.tobytes()
-        if isinstance(buf, _buffer):
-            buf = str(buf)
-
-        if not isinstance(buf, (bytes, bytearray)):
-            raise TypeError(
-                "data must be a memoryview, buffer, bytearray or byte string"
-            )
-
-        if len(buf) > 2147483647:
+        if size > 2147483647:
             raise ValueError("Cannot send more than 2**31-1 bytes at once.")
 
-        buf = _ffi.from_buffer(buf)
+        buf = self._buffer_factory(buf)
 
-        result = _lib.SSL_write(self._ssl, buf, len(buf))
+        result = _lib.SSL_write(self._ssl, buf, size)
         self._raise_ssl_error(self._ssl, result)
         return result
     write = send
@@ -1482,18 +1516,9 @@ class Connection(object):
         """
         buf = _text_to_bytes_and_warn("buf", buf)
 
-        if isinstance(buf, _memoryview):
-            buf = buf.tobytes()
-        if isinstance(buf, _buffer):
-            buf = str(buf)
-        if not isinstance(buf, (bytes, bytearray)):
-            raise TypeError(
-                "buf must be a memoryview, buffer, bytearray or byte string"
-            )
-
         left_to_send = len(buf)
         total_sent = 0
-        data = _ffi.from_buffer(buf)
+        data = self._buffer_factory(buf)
 
         while left_to_send:
             # SSL_write's num arg is an int,
--- a/src/OpenSSL/SSL.py
+++ b/src/OpenSSL/SSL.py
@@ -1453,10 +1453,16 @@ class Connection(object):
             buf = buf.tobytes()
         if isinstance(buf, _buffer):
             buf = str(buf)
-        if not isinstance(buf, bytes):
-            raise TypeError("data must be a memoryview, buffer or byte string")
+
+        if not isinstance(buf, (bytes, bytearray)):
+            raise TypeError(
+                "data must be a memoryview, buffer, bytearray or byte string"
+            )
+
         if len(buf) > 2147483647:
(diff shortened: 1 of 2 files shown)


hynek:
Hm.  I would assume most people run the latest cffi anyways.  Since we already have a conditional, how about we externalize it into a function `_is_acceptable_for_from_buffer()` whose implementation depends on the cffi version?

N.B. I’m not telling, I’m asking. :)

Lukasa:
I'm certainly open to that approach. It'd be interesting for @alex or @reaperhulk to weigh in with some relative usage data of different cffi versions, or at least with gut feelings, to work out how valuable that code will be, especially as it will suck to test. :wink:

hynek:
I think we could make it suck a lot less with something like this:

```python
_is_acceptable_for_from_buffer = _make_buffer_checker(cffi.__version__)
```

Lukasa:
Yeah, that's true, that could absolutely work.

reaperhulk:
We already have some code in cryptography dependent on cffi >= 1.7, but we don't require it so we do this:

https://github.com/pyca/cryptography/blob/master/src/cryptography/hazmat/primitives/ciphers/base.py#L148-L160

We get coverage on that branch by testing an old pypy with coverage enabled. https://github.com/pyca/cryptography/blob/master/.travis.yml#L27

The only real consequence to upping minimum cffi version is that you leave behind older PyPys (since they can't upgrade it). cryptography decided against doing that, but this project can make its own call.

reaperhulk:
Here are the number of downloads by implementation for the last month for pyopenssl:

```
1	2044707	CPython
2	108733	null	 
3	2238	PyPy	 
4	26	Jython	 
5	1	IronPython	 
```

reaperhulk:
And here's the top 10 versions within that PyPy umbrella:

```
1	816	5.4.1	 
2	536	5.6.0	 
3	134	5.3.1	 
4	122	4.0.1	 
5	117	5.7.0	 
6	110	5.7.1	 
7	62	2.4.0	 
8	61	5.0.1	 
9	50	5.4.0	 
10	50	2.5.0
```

A (very) quick check shows that PyPy2 5.2 shipped with cffi 1.8 (5.1 shipped with 1.6).

hynek:
I would just go the hybrid way I’ve sketched out.  It seems like something we can test easily and we’ve learned the hard way, that dependencies don’t get auto-updated. :|

src/OpenSSL/SSL.py
```diff
@@ -161,6 +164,53 @@ class SysCallError(Error):
...
+    Returns a callable that converts a sendable buffer object into something
+    that can be sent by OpenSSL using CFFI. Has different supported types
+    depending on the CFFI version.
+
+    :param cffi_version_string: The version of CFFI in use, from
+        ffi.__version__.
+    """
+    cffi_version = tuple(
+        int(component) for component in
+        _CFFI_VERSION_RE.match(cffi_version_string).group(1, 2)
```
hynek: why not just `cffi_version_string.split(".")`?  A RE seems overkill here?

reaperhulk: Why not just `from packaging.version import parse`? Having pyOpenSSL depend on packaging (which is a dependency of both setuptools and cryptography) doesn't seem unreasonable.

Lukasa: @hynek The main reason is that you can conceivably have a version string like `1.8dev4`, which will not behave nicely under `.split('.')`. It's possible I'm being too defensive here, but I was trying to avoid having everything explode in situations I didn't entirely forsee.

@reaperhulk That seems reasonable to me, if it's ok with @hynek I'll happily do that instead.

hynek: Go ahead!

tests/test_ssl.py
```diff
@@ -3880,3 +3896,56 @@ def client_callback(*args):  # pragma: nocover
...
+    @pytest.mark.parametrize('data', supported_by_newer_cffi)
+    def test_ffi_buffer_support_late(self, cffi_version, data):
+        """
+        When using CFFI versions 1.7 or later, bytearrays, bytes, and
+        memoryviews are supported.
+        """
+        buffer_factory = _make_buffer_checker(cffi_version)
+        cdata_buffer = buffer_factory(data)
+        # We need an explicit length here because cffi on recent PyPy doesn't
+        # 't correctly construct buffers when it knows the length of the cdata
```
hynek: 't looks like an oops :)

hynek:
@njsmith would this make you happy?

tests/test_ssl.py
Lukasa: Resolved.

src/OpenSSL/SSL.py
```diff
@@ -161,6 +164,47 @@ class SysCallError(Error):
...
+def _make_buffer_checker(cffi_version_string):
+    """
+    Returns a callable that converts a sendable buffer object into something
+    that can be sent by OpenSSL using CFFI. Has different supported types
+    depending on the CFFI version.
+
+    :param cffi_version_string: The version of CFFI in use, from
+        ffi.__version__.
+    """
+    cffi_version = parse(cffi_version_string)
```
njsmith: A quick check shows that cffi has had a pre-parsed version tuple at `cffi.__version_info__` for some time (I checked 1.0.0 and it had it). That might be nicer than pulling in a dependency on `packaging`?

src/OpenSSL/SSL.py
```diff
@@ -161,6 +164,47 @@ class SysCallError(Error):
...
+        """
+        Given a sendable data buffer passed into send or sendall, returns a
+        CFFI object representing it.
+
+        :param buf: The sendable data buffer the user wishes to send.
+        """
+        if isinstance(buf, _memoryview):
+            buf = buf.tobytes()
+        if isinstance(buf, _buffer):
+            buf = str(buf)
```
njsmith: 1) It may not matter much in practice, but I think on recent CFFI these lines are unnecessary and are inefficient in that they impose an extra copy? 
2) I might add a comment on the `str` line noting that this branch can only be taken on py2 (because py3 has no buffer objects)

src/OpenSSL/SSL.py
```diff
@@ -161,6 +164,47 @@ class SysCallError(Error):
...
+    """
+    Returns a callable that converts a sendable buffer object into something
+    that can be sent by OpenSSL using CFFI. Has different supported types
+    depending on the CFFI version.
+
+    :param cffi_version_string: The version of CFFI in use, from
+        ffi.__version__.
+    """
+    cffi_version = parse(cffi_version_string)
+    if cffi_version >= Version("1.7"):
```
njsmith: The [docs](https://cffi.readthedocs.io/en/latest/ref.html#ffi-buffer-ffi-from-buffer) say "bytearray objects were supported in version 1.7 onwards ... and byte strings were supported in version 1.8 onwards." So the minimum version should probably be 1.8, not 1.7?

njsmith:
Some comments above to do with as you wish. In general I am not too picky though, because I'm just using PyOpenSSL in my test suite, so I don't care about speed or dependencies or any of that :-). I was just surprised when I had to add a call to `bytes(...)` so I figured I'd be a good citizen and report it upstream...

src/OpenSSL/SSL.py
Lukasa: 1. They are, but the code refactor required to fix it and keep the tests in place was not something I wanted to undertake on this PR.
2. I don't think it matters really?

Lukasa:
Aside from the two points I addressed the rest are things I want to take action on, I'll do that sometime soon.

src/OpenSSL/SSL.py
njsmith: 2. Yeah, not really; I only mention it because I had a moment of "wtf you can't call `str` on a `buffer`!" that took a few minutes to resolve :-)

hynek:
oh looks like this only needs docs + changelog?  completely lost track of it!

dholth:
I was about to suggest this same bugfix! I guess you don't like just letting cffi decide whether the buffer passed is acceptable or not? In `SSL send()` e.g. `try: buf = _ffi.from_buffer(buf) except: pyopenssl error about buffers`

It almost certainly calls from_buffer internally as part of any wrapped function that takes char*?

reaperhulk:
@dholth I think at this point if you want to submit a new PR based on this we can look at that (@Lukasa is working on other projects and this is no longer an area of focus). When this PR was originally filed cryptography didn't require cffi >=1.8, but it does now so `from_buffer` will Just Work™.

dholth:
Obsoleted by #852

mhils:
This is fixed indeed, thanks @dholth! 🍰

dholth:
Hooray!

PR #629, opened 2017-05-31
Use the normal OpenSSL function instead of the crypto wrapper

alex:


reaperhulk:
Does anybody remember why we were using this? Was it conditional in 0.9.8 or something?

alex:
I do not.

alexwlchan:
> Does anybody remember why we were using this? Was it conditional in 0.9.8 or something?

Looks like a 0.9.8-ism. A comment on the [original implementation](https://github.com/alex/cryptography/blob/ff90806ea1c62212f77a4d040fe491d05fdf3462/src/_cffi_src/openssl/evp.py) of `Cryptography_EVP_PKEY_id` is suggestive:

```c
/* EVP_PKEY_id is not available on RHEL5 0.9.8e so we'll define our own */
int Cryptography_EVP_PKEY_id(const EVP_PKEY *key) {
    return key->type;
}
```

As does [the commit that simplified the implementation](https://github.com/pyca/cryptography/commit/41b4bce14260325348191044571c9808a77229aa).

PR #636, opened 2017-06-10
Allow elliptic curve keys in from_cryptography_key().

theno:
With this change an (Certificate Transparency) SCT can be verified
against the public key of a CT-Log. CT-Logs usually are signed with an
elliptic curve digest.  The argument `cert` of the function
`verify(cert, signature, data, digest)` is just a wrapper for its `pkey`
attribute.  This attribute now can contain an ec-pubkey.  Here is an example
of this usage:
https://github.com/theno/ctutlz/blob/60cd6b9aeee7a7c6f0f90b0262a306a292808985/ctutlz/sct/verification.py#L29

Elliptic curve support in PKey objects still needs to be implemented.

### DIFF
--- a/src/OpenSSL/crypto.py
+++ b/src/OpenSSL/crypto.py
@@ -10,7 +10,7 @@ from six import (
     text_type as _text_type,
     PY3 as _PY3)
 
-from cryptography.hazmat.primitives.asymmetric import dsa, rsa
+from cryptography.hazmat.primitives.asymmetric import ec, dsa, rsa
 
 from OpenSSL._util import (
     ffi as _ffi,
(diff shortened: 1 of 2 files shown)


Lukasa:
Thanks for this! However, I think you're approaching this backwards. Without support for EC keys in PKey, I don't think we should allow them to be loaded: I'd much rather see EC PKey support added first. 

I'd even more rather see SCR verification land in cryptography, as this seems right up @Alex's street.

theno:
For convenience I marked the failing test to be skipped running pytest to make unit tests passing.

This even makes the [arguments](https://github.com/pyca/pyopenssl/pull/636#issuecomment-307547980) of Lukasa more obvious.

ronf:
I was just about to submit a pull request very similar to this. I have a need for this functionality in my AsyncSSH package. I'm in the middle of adding support for X.509 certificates in SSH and I have things working for RSA certificates. However, I ran into this issue when attempting to use EC certificates.

As it turns out, EC certificates work just fine when you feed them to OpenSSL.crypto.load_certificate(). However, to support generating certificates, I need to be able to convert a cryptography EC key to a PKey, and it looks like these explicit checks are the only thing stopping that from working.

While it's true that PyOpenSSL doesn't support creating new EC PKeys directly yet, I don't need that in my application and supporting conversion from already constructed PyCA keys would be a good first step.

I might be able to work around this by generating a certificate using the native PyCA X.509 support and then using OpenSSL.crypto.load_certificate() to get a PyOpenSSL certificate object out of that. However, I'd prefer not to have to do that.

Long term, I'd love to convert everything over to PyCA and not depend on PyOpenSSL at all, but I can't do that until PyCA adds support for X.509 certificate chain validation.

alex:
This was added at some point.

PR #660, opened 2017-07-07
FIxed #266 -- attempt to deflake our tests

### LINKED ISSUE #266: Multiple tests have race conditions

alex:


alex:
Ok, I don't understand `test_gmtime_adj_notBefore` at all. Why does it fail? Why do we have this idiot API at all tbh.

reaperhulk:
That is an atrocious API. So is most of pyOpenSSL, but that's particularly egregious and I can't come up with a use for it. Anyway, LGTM.

PR #661, opened 2017-07-07
move deps to extras

reaperhulk:
This will make it so we don't have to keep updating the cryptography jenkinsfile whenever we change a test dep in pyopenssl.


alex:
update `MANIFEST.in`

PR #665, opened 2017-07-19
fix #664

### LINKED ISSUE #664: Root CA certs lookup in /opt/pyca/cryptography/openssl/certs
Hello,

When using pyOpenSSL/crytpography in a venv (cp36-cp36m-manylinux1_x86_64.whl) ssl_cert_dir defaults to /opt/pyca/cryptography/openssl/certs. After some digging, it seems to be related to a little miss in `src/OpenSSL/SSL.py` (_ffi.string return bytes and a later condition involves strings).

```diff
@@ -731,8 +731,12 @@ class Context(object):
             _lib.X509_get_default_cert_file_env()
         ).decode("ascii")
         if not self._check_env_vars_set(dir_env_var, file_env_var):
-            default_dir = _ffi.string(_lib.X509_get_default_cert_dir())
-            default_file = _ffi.string(_lib.X509_get_default_cert_file())
+            default_dir = _ffi.string(
+                _lib.X509_get_default_cert_dir()
+            ).decode("ascii")
+            default_file = _ffi.string(
+                _lib.X509_get_default_cert_file()
+            ).decode("ascii")
             # Now we check to see if the default_dir and default_file are set
             # to the exact values we use in our manylinux1 builds. If they are
             # then we know to load the fallbacks
```

Regards.

reaperhulk:


alex:
a) Why don't the tests catch this?
b) If these values are going to be unicode, please make `_CRYPTOGRAPHY_MANYLINUX1_CA_DIR` unicode.

reaperhulk:
Tests didn't catch this because to have it take the fallback path we fetched the values from cffi and the monkeypatched our constants (https://github.com/pyca/pyopenssl/blob/master/tests/test_ssl.py#L1114-L1142). So the comparison succeed since they were both bytes.

I think a better fix here is just to declare these as byte strings and not unicode. They're not really paths to us, just sentinel values.

alex:
That's also fine with me.

alex:
trailing whitespace makes the flake8 fail

reaperhulk:
Why can't GitHub's editor strip trailing whitespace like my vim config does.

alex:
Send a feature request!

PR #667, opened 2017-07-19
document set_default_verify_paths caveats

### LINKED ISSUE #642: Document set_default_verify_paths caveats for macOS/Windows users
Users with the statically linked cryptography wheel have some potentially confusing behaviors when they call `set_default_verify_paths`. Let's document that:

* macOS will work, but only if the user has the openssl@1.1 formula installed (in the default location) from homebrew
* windows will not work.
* manylinux1 will work using #633

reaperhulk:
fixes #642

### DIFF
--- a/doc/api/ssl.rst
+++ b/doc/api/ssl.rst
@@ -308,12 +308,14 @@ Context objects have the following methods:
 
     Specify that the platform provided CA certificates are to be used for
     verification purposes. This method has some caveats related to the
-    binary wheels that cryptography (pyOpenSSL's primary dependency ships:
+    binary wheels that cryptography (pyOpenSSL's primary dependency) ships:
 
     * macOS will only load certificates using this method if the user has
       the ``openssl@1.1`` Homebrew formula installed in the default location.
     * Windows will not work.
     * manylinux1 cryptography wheels will work in pyOpenSSL 17.1.0 and above.
+      pyOpenSSL detects the manylinux1 wheel and attempts to load roots via
+      a fallback path.
 
 .. py:method:: Context.load_tmp_dh(dhfile)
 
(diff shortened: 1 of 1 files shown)


tiran:
How did you get manylinux1 binaries working? Does pyopenssl 17.1.0 detect distribution and look up paths?

alex:
Pretty much, yes.

reaperhulk:
@tiran we use a sentinel value so pyOpenSSL can easily detect if it's running under manylinux1. If and only if that occurs does it use the fallback paths to find the root stores.

tiran:
Could you please add another sentence that mentions this behavior?

reaperhulk:
@tiran How's that?

doc/api/ssl.rst
```diff
@@ -307,8 +307,15 @@ Context objects have the following methods:
...
     Specify that the platform provided CA certificates are to be used for
-    verification purposes.  This method may not work properly on OS X.
-
+    verification purposes. This method has some caveats related to the
+    binary wheels that cryptography (pyOpenSSL's primary dependency) ships:
+
+    * macOS will only load certificates using this method if the user has
+      the ``openssl@1.1`` Homebrew formula installed in the default location.
+    * Windows will not work.
+    * manylinux1 cryptography wheels will work in pyOpenSSL 17.1.0 and above.
```
alex: Maybe we should be clearer "will work on most common linux distributions" or something

reaperhulk: Part of me wants to say that prevarication in the absence of an actual example is annoying, but undoubtedly somewhere somehow somebody is going to find a way to build a linux distro that won't work, so okay.

PR #668, opened 2017-07-19
Ask for signature length before allocating a buffer (#609)

jzakrzew:
Instead of using the private key size to estimate the buffer size, actually ask OpenSSL to compute it for us. This fixes a potential heap buffer overflow that may happen when a signature is longer than the private key, as it is with X9.62 ECDSA format.

This seems to fix issue #609

### DIFF
--- a/src/OpenSSL/crypto.py
+++ b/src/OpenSSL/crypto.py
@@ -2801,9 +2801,13 @@ def sign(pkey, data, digest):
     _lib.EVP_SignInit(md_ctx, digest_obj)
     _lib.EVP_SignUpdate(md_ctx, data, len(data))
 
-    pkey_length = (PKey.bits(pkey) + 7) // 8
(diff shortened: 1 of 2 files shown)


reaperhulk:
`test_sign_verify` is crashing in CI with this patch.

reaperhulk:
Once you get the crash resolved we'll also want a line in the CHANGELOG please. Thanks for the contribution so far!

reaperhulk:
The issue here appears to be that EVP_SignFinal didn't support calling it with a null buffer to determine the length of signature until 1.1.0. Doing it on 1.0.2 or below is an instant segfault. Instead we should size the buffer by using `_lib.EVP_PKEY_size(pkey._pkey)`.

reaperhulk:
Since we're going to ship 17.2.0 momentarily I went ahead and did this on #670 (cherry-picking your commits as well). Thank you!

PR #677, opened 2017-07-23
Drop py33

alex:
…upport it

Refs #676

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -13,7 +13,7 @@ Backward-incompatible changes:
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
 - Dropped support for Python 3.3.
-  `#677 <https://github.com/pyca/pyopenssl/pull/677`_
+  `#677 <https://github.com/pyca/pyopenssl/pull/677>`_
 
 
 Deprecations:
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -12,7 +12,8 @@ The third digit is only for regressions.
 Backward-incompatible changes:
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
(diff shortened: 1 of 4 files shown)


alex:
@hynek this PR is _not_ fully dropping support, it's just dropping support when using versions of cryptography that don't support it. If you want me to fully drop support, let me know.

hynek:
So on one hand: it’s rare that we manage to release a new pyOpenSSL without requiring newer bindings eventually.

OTOH, nobody uses Python 3.3 but Twisted seems to support it anyways.  So let’s ask @glyph if our dropping would be a welcome excuse to drop it too or whether it’ll make him sad.

alex:
Dropping it from cryptography was based on a review of the data, and its extremely low usage. IIRC, pyopenssl it is like .1% of downloads.

glyph:
IIRC @hawkowl has proposed dropping 3.3 and it's already on the slate of things to do because our metrics are similarly tiny.  3.4 is the first version of python 3, I'm pretty sure the memory hole will swallow up <= 3.3 in due time.

hynek:
If you please @alex... 🔥

alex:
🔥 🔥 🔥  MY PLEASURE 🔥 🔥 🔥

alex:
@hynek done.

CHANGELOG.rst
```diff
@@ -12,7 +12,8 @@ The third digit is only for regressions.
 Backward-incompatible changes:
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-*none*
+- Dropped support for Python 3.3.
+  `#677 <https://github.com/pyca/pyopenssl/pull/677`_
```
hynek: This lacks the closing `>` which is breaking CI.

hynek:
bye bye 🔥

PR #682, opened 2017-08-07
Don't use "TLSv1" as a default for loopback clients/servers

alex:


hynek:
2 typos in a 2 line diff – A+ Alex game! :)

PR #686, opened 2017-08-26
added method to export keying material from an ssl connection

kelbyludwig:
This adds a new method to export keying material from an SSL connection (RFC 5705).

It currently will not work without merging in https://github.com/pyca/cryptography/pull/3878.

### DIFF
--- a/src/OpenSSL/SSL.py
+++ b/src/OpenSSL/SSL.py
@@ -1941,6 +1941,31 @@ class Connection(object):
         _lib.SSL_SESSION_get_master_key(session, outp, length)
         return _ffi.buffer(outp, length)[:]
 
+    def export_keying_material(self, label, olen, context=None):
+        """
+        Obtain keying material for application use.
+
+        :param label - a disambiguating label string as described in RFC 5705
+        :param olen - the length of the exported key material in bytes
+        :param context - a per-association context value
+        :return the exported key material bytes or None
+        """
+        outp = _no_zero_allocator("unsigned char[]", olen)
+        # RFC 5705: "Labels here have the same definition as in TLS, i.e., an
+        # ASCII string with no terminating NULL."
+        # So we must remove the NULL terminator.
+        label_buf = _ffi.new("char[]", label)[0:len(label)]
+        context_buf, context_len, use_context, success = _ffi.NULL, 0, 0, 0
+        if context is not None:
+            context_buf = _ffi.new("unsigned char[]", context)
+            context_len = len(context)
+            use_context = 1
+        success = _lib.SSL_export_keying_material(self._ssl, outp, olen,
+                                                  label_buf, len(label),
+                                                  context_buf, context_len,
+                                                  use_context)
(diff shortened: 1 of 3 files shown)


kelbyludwig:
It is also worth mentioning that

alexwlchan:
This is blocked not just on the patch in cryptography being merged, but also a new release. The PyOpenSSL CI doesn’t install from cryptography master, so this will keep failing for now.

reaperhulk:
That's not strictly true -- we have cryptography master tests (that are allowed to fail) in travis as well. In fact, they're currently failing on py3 indicating a bytes/str bug in this PR (https://travis-ci.org/pyca/pyopenssl/jobs/269830756).

reaperhulk:
That said, Travis won't go green until after a cryptography release (and this PR will need to bump the cryptography dependency as well at that point).

kelbyludwig:
The Travis checks look like they might only be failing for versions of cryptography that don't have https://github.com/pyca/cryptography/pull/3878 included?

alex:
The set of tests that are failing is the `cryptographyMinimum` set. Take a look at `tox.ini` and see about bumping that.

kelbyludwig:
the py35-urllib3Master failure seems unrelated.

kelbyludwig:
Is there anything else I could be doing to get this ready for review?

src/OpenSSL/SSL.py
```diff
@@ -1941,6 +1941,31 @@ def master_key(self):
...
+        :param label - a disambiguating label string as described in RFC 5705
+        :param olen - the length of the exported key material in bytes
+        :param context - a per-association context value
+        :return the exported key material bytes or None
+        """
+        outp = _no_zero_allocator("unsigned char[]", olen)
+        # RFC 5705: "Labels here have the same definition as in TLS, i.e., an
+        # ASCII string with no terminating NULL."
+        # So we must remove the NULL terminator.
+        label_buf = _ffi.new("char[]", label)[0:len(label)]
```
reaperhulk: Do you need to slice this? cffi will add a terminating NULL but when you pass it to `SSL_export_keying_material` with `len(label)` the terminating NULL simply won't be read.

src/OpenSSL/SSL.py
```diff
@@ -1941,6 +1941,31 @@ def master_key(self):
...
+        context_buf, context_len, use_context, success = _ffi.NULL, 0, 0, 0
+        if context is not None:
+            context_buf = _ffi.new("unsigned char[]", context)
+            context_len = len(context)
+            use_context = 1
+        success = _lib.SSL_export_keying_material(self._ssl, outp, olen,
+                                                  label_buf, len(label),
+                                                  context_buf, context_len,
+                                                  use_context)
+        return _ffi.buffer(outp, olen)[:] if success == 1 else None
```
reaperhulk: We shouldn't return `None` if this fails. This should call `_openssl_assert(success == 1)`, which will raise an exception if exporting the keying material fails.

reaperhulk:
Finished in #725

PR #695, opened 2017-09-15
Patch 1

mayani:
Only raise exception if underlying library has reported any errors.


reaperhulk:
OpenSSL commonly returns errors where it hasn't added anything to the error queue. This change would make it so those scenarios do not raise an exception, which is not what we want because those are still fatal errors.

mayani:
Thanks, I wasn't aware of it.

PR #708, opened 2017-11-20
restore a subset of the rand module

### LINKED ISSUE #706: Security Implications of #675 (Delete rand.py)
In #675, the rand module was removed from the codebase, and users of the module were instructed to use os.urandom instead. However, using os.urandom is not an alternative for the whole rand module, just for some functions like rand.bytes.

There are other functions in the module, such as rand.add, that are changing the PRNG state of OpenSSL. Using such functions is of particular importance for fork safety, as described here: https://wiki.openssl.org/index.php/Random_fork-safety

and here:
https://docs.python.org/2/library/ssl.html#multi-processing

Please consider reverting #675, or at least bringing back the minimum subset of the rand module that will allow pyopenssl users to deal with fork safety.

reaperhulk:
fixes #706 

We should bikeshed on the changelog language probably.

### DIFF
--- a/src/OpenSSL/rand.py
+++ b/src/OpenSSL/rand.py
@@ -1,0 +1,59 @@
+"""
+PRNG management routines, thin wrappers.
+"""
+
+from OpenSSL._util import (
+    ffi as _ffi,
+    lib as _lib,
+    exception_from_error_queue as _exception_from_error_queue,
+    path_string as _path_string)
+
+
+_builtin_bytes = bytes
+
+
+def add(buffer, entropy):
+    """
+    Mix bytes from *string* into the PRNG state.
+
+    The *entropy* argument is (the lower bound of) an estimate of how much
+    randomness is contained in *string*, measured in bytes.
+
+    For more information, see e.g. :rfc:`1750`.
+
+    This function is only relevant if you are forking Python processes and
+    need to reseed the CSPRNG after fork.
+
+    :param buffer: Buffer with random data.
+    :param entropy: The entropy (in bytes) measurement of the buffer.
+
+    :return: :obj:`None`
+    """
+    if not isinstance(buffer, _builtin_bytes):
+        raise TypeError("buffer must be a byte string")
+
+    if not isinstance(entropy, int):
+        raise TypeError("entropy must be an integer")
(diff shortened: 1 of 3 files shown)


alex:
The error stuff appers unused?

src/OpenSSL/rand.py
```diff
@@ -0,0 +1,55 @@
+"""
+PRNG management routines, thin wrappers.
+"""
+
+from OpenSSL._util import lib as _lib
+
+
+_builtin_bytes = bytes
```
alex: This isn't needed, just use `bytes` below.

src/OpenSSL/rand.py
```diff
@@ -0,0 +1,55 @@
...
+def status():
+    """
+    Check whether the PRNG has been seeded with enough data.
+
+    :return: 1 if the PRNG is seeded enough, 0 otherwise.
+    """
+    return _lib.RAND_status()
+
+
+def cleanup():
```
alex: Do we really need this?

tests/test_rand.py
```diff
@@ -0,0 +1,48 @@
...
+
+    def test_status(self):
+        """
+        `OpenSSL.rand.status` returns `1` if the PRNG has sufficient entropy,
+        `0` otherwise.
+        """
+        # It's hard to know what it is actually going to return.  Different
+        # OpenSSL random engines decide differently whether they have enough
+        # entropy or not.
+        assert rand.status() in (0, 1)
```
alex: _Surely_ this should actually always be `1`?

src/OpenSSL/rand.py
reaperhulk: Probably not. As of 1.1.0 this is already a noop

tests/test_rand.py
reaperhulk: Maybe it's possible to trigger a 0 when calling cleanup and not adding enough back...? I don't know. This is just a restoration of the code that was previously there, but this test is not particularly useful. I'd be fine changing it to assert 1. Plus as an added bonus we'll get to find out some insane edge case if it turns out 1 isn't always true in the tests.

PR #709, opened 2017-11-20
Don't use things after they're freed...duh

alex:


CHANGELOG.rst
```diff
@@ -23,8 +23,9 @@ Deprecations:
 Changes:
 ^^^^^^^^
 
-*none*
 
+- Corrected a use-after-free with some uses of the ``X509`` API.
```
reaperhulk: Should we add a sentence here to state that referencing a previously obtained issuer/subject after a subsequent set call will now raise an exception?

PR #710, opened 2017-11-20
fix errors with latest flake8

alex:


alex:
urllib3's tests appear broken, unrelated to us. Should I move them into allow_failures for now?

PR #713, opened 2017-11-21
enable direct CSR creation

joerichter-stash:
By deriving the class X509Req from X509 certificate requests can be generated from simple dictionaries, thus allowing very easy request templates.
Additionally the keywords from the OpenSSL can be used, which is very beneficial when the user wants to transition existing configurations.

Please see the following example for template based request generation:
```python
from cryptography.hazmat.backends import default_backend
from cryptography.hazmat.primitives.asymmetric import rsa, ec
from cryptography.hazmat.primitives.serialization import Encoding, PrivateFormat, NoEncryption
from OpenSSL import crypto
from OpenSSL.crypto import X509Extension

def gen_ecdsa_key(curve_name, length):
    ec_curve = crypto.get_elliptic_curve(curve_name)
    key = ec.generate_private_key(ec_curve, default_backend())
    key_pem = key.private_bytes(encoding=Encoding.PEM,
            format=PrivateFormat.TraditionalOpenSSL,
            encryption_algorithm=NoEncryption())
    return crypto.load_privatekey(crypto.FILETYPE_PEM, key_pem)

class WrapExtension(object):
    """wrapper for OpenSSL.crypto.X509Extension
    Allows to Wrap an extension in a template dict and build it on certificate/
    request generation.
    """
    def __init__(self, name, value, critical=False):
        self.name = name
        self.value = value
        self.critical = critical
        self.subject = None
        self.issuer = None

    def build_extension(self):
        obj = X509Extension(self.name, self.critical, self.value,
                self.subject, self.issuer)
        return obj


def create_cert_req(pkey, digest="sha256", **name):
    req = crypto.X509Req()
    subj = req.get_subject()
    wrapped_extensions = []

    for key, value in name.items():
        if "extensions" == key:
            #delay extension construction till all req-params are set
            wrapped_extensions = value
        else:
            setattr(subj, key, value)
    req.set_pubkey(pkey)

    # build extensions
    extensions = []
    for e in wrapped_extensions:
        e.subject = req
        extensions.append(e.build_extension())

    req.add_extensions(extensions)
    req.sign(pkey, digest)
    return req


CERT_TMPL_1 = { "countryName": "DE",
    "organizationName": "Test CA",
    "organizationalUnitName": "Example AG",
    "commonName": "empty",
    "stateOrProvinceName": "Saxony",
    "localityName": "Dresden",
    "emailAddress": "user@example.com",
    "extensions": [WrapExtension('keyUsage', 'digitalSignature', True),
                   WrapExtension('extendedKeyUsage', 'clientAuth'),
                   WrapExtension('subjectKeyIdentifier', 'hash')]
    }


def request_from_tmpl(cert_tmpl, user):
    """create a certificate request based on a template-dictionary """
    key = gen_ecdsa_key("secp256k1", 256)
    req_param = dict(cert_tmpl) #copy template
    req_param["commonName"] = user
    req_param["emailAddress"] = user + "@example.com"
    req = create_cert_req(key, **req_param)
    return req


request_from_tmpl(CERT_TMPL_1, "user1234")
```

### DIFF
--- a/src/OpenSSL/crypto.py
+++ b/src/OpenSSL/crypto.py
@@ -830,212 +830,6 @@ X509ExtensionType = deprecated(
 )
 
 
-class X509Req(object):
-    """
-    An X.509 certificate signing requests.
-    """
-
-    def __init__(self):
-        req = _lib.X509_REQ_new()
-        self._req = _ffi.gc(req, _lib.X509_REQ_free)
-        # Default to version 0.
-        self.set_version(0)
-
-    def to_cryptography(self):
-        """
-        Export as a ``cryptography`` certificate signing request.
-
-        :rtype: ``cryptography.x509.CertificateSigningRequest``
-
-        .. versionadded:: 17.1.0
-        """
-        from cryptography.hazmat.backends.openssl.x509 import (
-            _CertificateSigningRequest
-        )
-        backend = _get_backend()
-        return _CertificateSigningRequest(backend, self._req)
-
-    @classmethod
-    def from_cryptography(cls, crypto_req):
-        """
-        Construct based on a ``cryptography`` *crypto_req*.
-
-        :param crypto_req: A ``cryptography`` X.509 certificate signing request
-        :type crypto_req: ``cryptography.x509.CertificateSigningRequest``
-
-        :rtype: PKey
-
-        .. versionadded:: 17.1.0
-        """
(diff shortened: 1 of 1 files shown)


alex:
Thank you for your interest in contributing to pyOpenSSL, howerver as @reaperhulk has said, we're not interested in making any large changes.

(Even if we were, it's very incorrect for X509Req to subclass X509)

PR #721, opened 2017-11-28
Pin pytest until we drop 2.6

hynek:
pytest 3.3 drops Python 2.6 and (ironically) Python 3.3.

I would of course much prefer to drop 2.6 and cryptography 2.2 is dropping 2.6 so an alternative PR to this one would be to kill 2.6 with fire if cryptography 2.2 is on a realistic horizon.

We need to do either of those two because otherwise our tests will fail.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -102,7 +102,9 @@ if __name__ == "__main__":
             "test": [
                 "flaky",
                 "pretend",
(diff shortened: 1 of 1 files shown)


reaperhulk:
cryptography 2.2 is probably 4-8 weeks away (we haven't really done much work thus far).

Issue #722, opened 2017-11-28
Problem in gajim IM with pyopenssl 17.4.0 in ArchLinux


Genues:
Hi!
Please help solved problem in gajim IM with pyopenssl 17.4.0 in ArchLinux:
$ gajim
Traceback (most recent call last):
  File "/usr/lib/python2.7/site-packages/nbxmpp/idlequeue.py", line 549, in _process_events
    return IdleQueue._process_events(self, fd, flags)
  File "/usr/lib/python2.7/site-packages/nbxmpp/idlequeue.py", line 409, in _process_events
    obj.pollin()
  File "/usr/lib/python2.7/site-packages/nbxmpp/transports_nb.py", line 446, in pollin
    self._do_receive()
  File "/usr/lib/python2.7/site-packages/nbxmpp/transports_nb.py", line 669, in _do_receive
    self._on_receive(received)
  File "/usr/lib/python2.7/site-packages/nbxmpp/transports_nb.py", line 683, in _on_receive
    self.on_receive(data)
  File "/usr/lib/python2.7/site-packages/nbxmpp/client_nb.py", line 332, in <lambda>
    self.onreceive(lambda _data:self._xmpp_connect_machine(mode, _data))
  File "/usr/lib/python2.7/site-packages/nbxmpp/client_nb.py", line 385, in _xmpp_connect_machine
    self._xmpp_connect_machine(mode='STREAM_STARTED')
  File "/usr/lib/python2.7/site-packages/nbxmpp/client_nb.py", line 388, in _xmpp_connect_machine
    self._on_stream_start()
  File "/usr/lib/python2.7/site-packages/nbxmpp/client_nb.py", line 424, in _on_stream_start
    self._on_connect()
  File "/usr/lib/python2.7/site-packages/nbxmpp/client_nb.py", line 461, in _on_connect
    self.on_connect(self, self.connected)
  File "/usr/lib/python2.7/site-packages/gajim/common/connection.py", line 1428, in _connect_success
    return self.connection_accepted(con, con_type)
  File "/usr/lib/python2.7/site-packages/gajim/common/connection.py", line 1498, in connection_accepted
    hostname) and '100' not in gajim.config.get_per('accounts',
  File "/usr/lib/python2.7/site-packages/gajim/common/check_X509.py", line 171, in check_certificate
    subject = cert.get_subject()
  File "/usr/lib/python2.7/site-packages/OpenSSL/crypto.py", line 1438, in get_subject
[...]

PR #723, opened 2017-11-29
fix a memory leak and a potential UAF and also #722

reaperhulk:
In the callback we were creating an X509 object that didn't own its own memory so if you retained a reference to it after the callback completed you could end up in a UAF situation.

In PKCS12 we freed the stack but freeing the stack doesn't free the underlying objects. This meant that any PKCS12 with cacerts was leaking memory every load.

Finally, we clean up one other little area where we can use `_from_raw_x509_ptr`.

This needs CHANGELOG entries, but it also can't be merged until we backport https://github.com/pyca/cryptography/pull/4028 and release a 2.1.4 (and then upgrade our minimum dep in pyOpenSSL to 2.1.4, d'oh)

### DIFF
--- a/src/OpenSSL/crypto.py
+++ b/src/OpenSSL/crypto.py
@@ -3058,8 +3058,7 @@ def load_pkcs12(buffer, passphrase=None):
         pycert = None
         friendlyname = None
     else:
-        pycert = X509.__new__(X509)
-        pycert._x509 = _ffi.gc(cert[0], _lib.X509_free)
+        pycert = X509._from_raw_x509_ptr(cert[0])
 
         friendlyname_length = _ffi.new("int*")
         friendlyname_buffer = _lib.X509_alias_get0(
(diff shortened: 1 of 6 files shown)


src/OpenSSL/crypto.py
```diff
@@ -3073,8 +3072,8 @@ def load_pkcs12(buffer, passphrase=None):
 
     pycacerts = []
     for i in range(_lib.sk_X509_num(cacerts)):
-        pycacert = X509.__new__(X509)
-        pycacert._x509 = _lib.sk_X509_value(cacerts, i)
+        x509 = _lib.sk_X509_value(cacerts, i)
+        pycacert = X509._from_raw_x509_ptr(x509)
```
reaperhulk: I'd like a double check here. I'm not confident this is right. Does `sk_X509_free` free the underlying `X509 *` in the stack? If so then this is wrong, but that would also mean the current code in master is just a giant UAF.

alex: `sk_x509_free` only frees the stack itself, not any of the things it contains, afaik

reaperhulk: Okay, if that's true (which is what I thought but it's amazing how quickly you start second guessing yourself when fixing memory bugs) then should be right.

alex: So the memory approach here is that each `X509` object is responsible for hte `X509*` and the `cacerts` var is only responsible for freeing the `stack`?

reaperhulk: Correct

PR #725, opened 2017-11-30
Export keying material support

reaperhulk:
Continued from #686. Huge thanks to @kelbyludwig for doing all the work here.

### DIFF
--- a/src/OpenSSL/SSL.py
+++ b/src/OpenSSL/SSL.py
@@ -2041,7 +2041,9 @@ class Connection(object):
         :return the exported key material bytes or None
         """
         outp = _no_zero_allocator("unsigned char[]", olen)
-        context_buf, context_len, use_context, success = _ffi.NULL, 0, 0, 0
+        context_buf = _ffi.NULL
+        context_len = 0
+        use_context = 0
         if context is not None:
             context_buf = context
             context_len = len(context)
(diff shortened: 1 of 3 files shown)


src/OpenSSL/SSL.py
```diff
@@ -2031,6 +2031,28 @@ def master_key(self):
...
+        """
+        Obtain keying material for application use.
+
+        :param label - a disambiguating label string as described in RFC 5705
+        :param olen - the length of the exported key material in bytes
+        :param context - a per-association context value
+        :return the exported key material bytes or None
+        """
+        outp = _no_zero_allocator("unsigned char[]", olen)
+        context_buf, context_len, use_context, success = _ffi.NULL, 0, 0, 0
```
alex: Use normal assignment, not unpacking.

src/OpenSSL/SSL.py
```diff
@@ -2031,6 +2031,28 @@ def master_key(self):
...
+        if context is not None:
+            context_buf = context
+            context_len = len(context)
+            use_context = 1
+        success = _lib.SSL_export_keying_material(self._ssl, outp, olen,
+                                                  label, len(label),
+                                                  context_buf, context_len,
+                                                  use_context)
+        _openssl_assert(success == 1)
```
alex: This function has error conditions on SSLv3 and some DTLS nonsense I didn't read very closely. Do we need to handle them?

PR #731, opened 2018-01-22
Add DTLS support

gesslerpd:
This PR depends on pyca/cryptography#4089 and adds DTLS capability to pyOpenSSL.

I have a very simple example of this working in `examples/simple_dtls` although the API may need to be changed for the server implementation.


gesslerpd:
Any help with improvements to the server example, non-blocking example, and/or API change brainstorming would be appreciated!

gesslerpd:
Currently this PR doesn't support the `SSL.OP_COOKIE_EXCHANGE` context option and HelloVerifyRequests are not sent by the server. Interfaces to the `SSL_CTX_set_cookie_generate_cb` and `SSL_CTX_set_cookie_verify_cb` methods need to be added to the pyOpenSSL API to support this option.

gesslerpd:
@reaperhulk To support DTLS v1.2 (the generic DTLS method) and the OP_NO_DTLS* options, the tests need to run with openssl>=1.0.2, right now they run with `OpenSSL 1.0.1f 6 Jan 2014`. According to http://brewformulas.org/Openssl, the brew install version is `1.0.2n`.

jlaine:
This pull requests needs rebasing now that documentation is automatically extracted from docstrings

gesslerpd:
@chrysn I can make this change later today to keep the PR up to date w/ master. Do you have any experience working with the OpenSSL DTLS API in other contexts outside of this project? I've been researching, but cannot seem to get the DTLS server side to handle multiple simultaneous connections.

chrysn:
> I can make this change later today to keep the PR up to date w/
> master.

That'd be great, thank you!

> Do you have any experience working with the OpenSSL DTLS API in other
> contexts outside of this project? I've been researching, but cannot
> seem to get the DTLS server side to handle multiple simultaneous
> connections.

Right now I'm using a Python wrapper to tinyTLS, which does have some API
for that, but TBH I prefer to have an API that doesn't try to work
with IP addresses in the first place and manage "connections" myself.
That doesn't need to be the API that is advertised as "do this" in the
documentation, but it should be accessible to other libraries.  In the
tinDTLS case, I take apart the streams and then feed them with a
[::1]:1 fake IP because the API demands that there be one.

One reasons is that I prefer to do retransmission across layers: send
the same encrypted package again rather than send a high-level
retransmit with a new DTLS sequence number, for reasons of
<https://tools.ietf.org/html/draft-ietf-core-echo-request-tag-01#section-3.3.1>
where I have to know whether I "lost" cryptographic packets, and thus
they won't count as lost if I get something back on a retransmit.

There can be an API that looks similar to a UDP socket API as well, but
given there is no multicast in DTLS and there's usually no direct state
exchange among "connections", I think that API better look like a TCP
API rather than like a UDP API. In terms of asyncio's transports and
protocols (because that's what I am using): it should probably have a
"create_connection" and "create_server" constructor, where the former
spawns one Transport and the latter several ones, which have a
`.send(data)` function and trigger a `.datagram_received(data)` method.

jlaine:
"Multi connections" isn't going to work, the SSL context is stateful. If you wanted to handle multiple clients, you would need to have some kind of demultiplexing logic (for instance based on the remote (IP, port)) which feeds your data into the appropriate SSL context.

chrysn:
It appears that it has only been developed on Python 2, as I get a TypeError("data must be bytes"); an `h.update(str(conn).encode('utf8'))` instead of the plain `update(str(...))` after the TODO comment in line 37 fixes that.

I did not manage to get the server to respond with DTLS 1.2 Hello Verify Requests to the DTLS 1.2 Client Hello I am sending; is that supposed to be supported so far? (DTLSv1_METHOD is also the only attribute of SSL matching "dtls*method" at the point where it is passed to SSL.Context; I'm using the cryptography module version 2.2.2 as installed by setup.py, with libssl1.1 version 1.1.1-pre3 from the Linux installation).

gesslerpd:
I've gotten a simple DTLS 1.2 client/server echo connection running (on python 2 I need to get that working on 3), although usually the cookie is generated using the client IP/PORT combination and I didn't get this information passed to the callback yet, so right now the code in that example is using the stringified connection to get by (this needs to be fixed). The generic DTLS method and OP_NO* codes are commented out right now for the tests to pass because they run on a lower openssl version even though it works locally.

chrysn:
I've made some progress on the DTLS1.2 side too; I had to move "DTLS_server_method" into the _methods of SSL.py, and had to unset OP_COOKIE_EXCHANGE (for some reasons that comes back DTLS 1.0).

src/OpenSSL/SSL.py
```diff
@@ -1914,6 +2045,10 @@ def connect(self, addr):
         :param addr: A remote address
         :return: What the socket's connect method returns
         """
+        if self.is_datagram_socket:
+            # DTLS BIO set connected
+            _lib.BIO_ctrl(_lib.SSL_get_rbio(self._ssl),
+                          _lib.BIO_CTRL_DGRAM_SET_CONNECTED, 0, addr[0])
```
davidben: Drive-by comment since I was fiddling with the bindings side. This doesn't look right. See the discussion here:
https://github.com/pyca/cryptography/pull/4220#issuecomment-388480321

jarednm:
Is there any update on getting DTLS pulled in?

fbexiga:
Is there any new progress on this?

Gilks:
Checking in on this. I see this PR has not had much activity since 2018. Is this functionality ever being landed? I am looking at trying to implement DTLS into my project but it involves supporting an outdated third party library.

PR #733, opened 2018-02-06
Add Connection.get_certificate method

jlaine:
This makes it possible to retrieve the local certificate (if any)
for a Connection.

An example where this is useful is when negotiating a DTLS-SRTP
connection, the fingerprint of the local certificate needs to be
communicated to the remote party out-of-band via SDP.

This depends on:
https://github.com/pyca/cryptography/pull/4100

### DIFF
--- a/src/OpenSSL/SSL.py
+++ b/src/OpenSSL/SSL.py
@@ -2156,6 +2156,18 @@ class Connection(object):
         """
         return self._socket.shutdown(*args, **kwargs)
 
+    def get_certificate(self):
+        """
+        Retrieve the local certificate (if any)
+
+        :return: The local certificate
+        """
+        cert = _lib.SSL_get_certificate(self._ssl)
+        if cert != _ffi.NULL:
+            _lib.X509_up_ref(cert)
+            return X509._from_raw_x509_ptr(cert)
+        return None
+
     def get_peer_certificate(self):
         """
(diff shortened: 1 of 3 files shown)


src/OpenSSL/SSL.py
```diff
@@ -2063,6 +2063,18 @@ def sock_shutdown(self, *args, **kwargs):
...
 
+    def get_certificate(self):
+        """
+        Retrieve the local certificate (if any)
+
+        :return: The local certificate
+        """
+        cert = _lib.SSL_get_certificate(self._ssl)
+        if cert != _ffi.NULL:
+            _lib.X509_up_ref(cert)
```
jlaine: Unlike SSL.get_peer_certificate, we don't own the X509 pointer. We need to increase the reference count to avoid the X509 freeing the pointer on GC.

jlaine:
As far as I can tell, this pull request is complete : tests, docstring and an entry in api/ssl.rst. All that's needed is for cryptography 2.2 to be released.

reaperhulk:
I will be circling around to figure out a rough release date for 2.2 next week.

reaperhulk:
Okay, this can be updated to set cryptography 2.2 as minimum version

jlaine:
OK, so this requires #742

PR #734, opened 2018-02-06
Add Context.set_tlsext_use_srtp

jlaine:
This allows negotiating SRTP keying material, which is useful when using
DTLS-SRTP, as WebRTC does for example.

This depends on:
https://github.com/pyca/cryptography/pull/4099

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -24,7 +24,8 @@ Deprecations:
 Changes:
 ^^^^^^^^
 
-*none*
+- Added ``Context.set_tlsext_use_srtp`` to enable negotiation of SRTP keying material.
+  `#734 <https://github.com/pyca/pyopenssl/pull/734>`_
 
 
 ----
--- a/src/OpenSSL/SSL.py
+++ b/src/OpenSSL/SSL.py
@@ -1371,6 +1371,21 @@ class Context(object):
         _lib.SSL_CTX_set_tlsext_servername_callback(
             self._context, self._tlsext_servername_callback)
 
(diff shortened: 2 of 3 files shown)


jlaine:
I'm not sure what the best form would be for the "profiles" argument. Two possibilities:

- A colon-delimited bytestring : this mimics the OpenSSL API exactly, and works the same as existing Context.set_cipher_list for example

- A list of bytestrings : this feels somewhat more pythonic, but deviates from the OpenSSL API

Suggestions?

reaperhulk:
I dislike the OpenSSL APIs for this, but pyOpenSSL generally doesn't abstract much away and since we already have an API that exposes colon delimited behavior it probably makes sense for this one to look like that as well.

One day we'll have a nice TLS API. pyOpenSSL won't be it...

jlaine:
OK thanks for the feedback I'll write the corresponding API docs.

Quick question : any reason why we don't use sphinx.autodoc's automethod? It seems a shame to have to duplicate the docstrings everywhere.

hynek:
We do for new code. Nobody wanted to spend their time to move everything to docstrings yet.

jlaine:
Duly noted, I've updated my pull request to use automethod

jlaine:
This requires #742

jlaine:
@reaperhulk anything you'd like me to change with this PR?

reaperhulk:
@jlaine could you rebase this?

reaperhulk:
Actually I can resolve, nevermind :)

PR #742, opened 2018-03-19
Raise minimum cryptography version to 2.2.1, drop python 2.6

jlaine:


setup.py
```diff
@@ -95,16 +94,14 @@ def find_meta(meta):
...
             "six>=1.5.2"
         ],
         extras_require={
             "test": [
                 "flaky",
                 "pretend",
-                # pytest 3.3 doesn't support Python 2.6 anymore.
-                # Remove this pin once we drop Python 2.6 too.
-                "pytest>=3.0.1,<3.3.0",
+                "pytest",
```
alex: Please keep the version minimum in the pin

jlaine: OK, done!

PR #744, opened 2018-03-25
Make sure a NotImplementedError is always raised on Connection.makefile()

aykevl:
With this patch, code which calls (for example) `conn.makefile('rb')` will get a `NotImplementedError` instead of a confusing `TypeError`:

    TypeError: makefile() takes 1 positional argument but 2 were given

### DIFF
--- a/src/OpenSSL/SSL.py
+++ b/src/OpenSSL/SSL.py
@@ -2019,7 +2019,7 @@ class Connection(object):
             result.append(pyname)
         return result
 
-    def makefile(self, mode=None, bufsize=None):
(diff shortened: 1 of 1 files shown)


src/OpenSSL/SSL.py
```diff
@@ -2019,7 +2019,7 @@ def get_client_ca_list(self):
             result.append(pyname)
         return result
 
-    def makefile(self):
+    def makefile(self, mode=None, bufsize=None):
```
alex: Can you just have this accept `*args, **kwargs` since we don't care what args you call it with?

PR #775, opened 2018-07-21
ISSUE774: add authorityInfoAcces, CDP, AltNames getting info from extension

igron99:
add authorityInfoAcces, CDP, AltNames getting info from extension


igron99:
Unusable.
I would try to write more usable interface...

igron99:
I need in iterator to get OIDs from EKUs. I can do it by _lib.

PR #776, opened 2018-07-23
add key types consts

igron99:
Added PKEY constants for DH and EC types.


igron99:
It can be usefull, if u want to know type of curve

Issue #177, opened 2014-12-06
No support for decoding b64 NetscapeSPKI objects


delroth:
The documentation indicates that it's possible (by passing the base64-encoded DER to the NetscapeSPKI constructor), but the code does not support that.

AFAICT this would need to be done in two steps:
1. Add binding for NETSCAPE_SPKI_b64_decode in cryptography.
2. Implement that one argument variant of crypto.NetscapeSPKI.__init__

joemiller:
Not sure if this is related but it might be. It seems like there is a regression from 0.13 to 0.14 that breaks this class.

Loading a SPKAC string returned by a browser `<keygen>` tag:

```
def spkac_to_pkey(spkac_str):
    """takes a SPKAC string sent by a browser using <keygen> and converts it into
    an OpenSSL.crypto.PKey object. Throws OpensSSL.Crypto.Error.
    """
    return crypto.NetscapeSPKI(spkac_str).get_pubkey()
```

- 0.14
```
  File "/data/home/joe/cert-maker/cert_maker.py", line 15, in spkac_to_pkey
    return crypto.NetscapeSPKI(spkac_str).get_pubkey()
TypeError: __init__() takes exactly 1 argument (2 given)
```

Code works OK in 0.13

delroth:
Oh, I didn't know it used to work in 0.13. I thought it was just never implemented in PyOpenSSL.

Regardless, It looks like cryptography had new releases since I opened this issue. I don't have any time right now to increase PyOpenSSL requirements and implement that __init__(self, str) constructor, so feel free to take this over.

futureimperfect:
@delroth @joemiller Any update on this? Is it just the constructor change that's needed now? Thanks!

alex:
We're not going to be expanding the NetscapeSPKI interface.

PR #777, opened 2018-07-28
Support initialization of NetscapeSPKI (#177)

j-d-r:
Add a way to initialize NetscapeSPKI object.

NetscapeSPKI is somehow obsolete, but only the initializer is missing.

- Call to NETSCAPE_SPKI_b64_decode(arg, -1) is taken from https://github.com/openssl/openssl/blob/a75be9fd34b5d66f349186f21cd8d063d2fa87a4/apps/ca.c#L1985

edit: cleanup

### DIFF
--- a/src/OpenSSL/crypto.py
+++ b/src/OpenSSL/crypto.py
@@ -2582,9 +2582,28 @@ class NetscapeSPKI(object):
     A Netscape SPKI object.
     """
 
-    def __init__(self):
-        spki = _lib.NETSCAPE_SPKI_new()
-        self._spki = _ffi.gc(spki, _lib.NETSCAPE_SPKI_free)
+    def __init__(self, spkac_str=None):
+        if spkac_str is not None:
+            self._spki = NetscapeSPKI.b64_decode(spkac_str)._spki
+        else:
+            spki = _lib.NETSCAPE_SPKI_new()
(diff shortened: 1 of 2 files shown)


j-d-r:
add missing unit tests

j-d-r:
rebased to pass checks

reaperhulk:
What are you using NSPKI for?

j-d-r:
I'm using it offline to generate user certificates from a webserserver ( https://github.com/j-d-r/UserCertServer )

It's the only way I found to store certificate in Firefox mobile.

reaperhulk:
@alex are you aware of any other way to generate CSRs on Firefox Mobile? And, if not, could we get that fixed on Firefox? 😁

If we need to support SPKI I'm inclined to say that we should implement support on the `cryptography` side rather than expand the utility of this junk API though.

alex:
Ok, chatted with J.C. for a few on this...

So, SPKAC is what's generated by `<keygen>`. And that's the only way to generate a key on device -- you can just import a `.der` if you want to import a client cert though. Anyways, `<keygen>` is extremely deprecated at this point, Chrome has dropped support, and I think we plan to drop support in Firefox as well. So I'd be opposed to expanding our support for this API for this reason.

alex:
Closing this in line with my last comment.

PR #782, opened 2018-08-09
fix wantWriteError to error even with big socket buffers

catern:
My system apparently has larger socket buffers than this test assumes, so it fails.
(Debian 9, Linux 4.16, Python 3.7) 
So let's increase the size of the buffers such that it works for me.
This was the smallest power of 2 that worked.

### DIFF
--- a/tests/test_ssl.py
+++ b/tests/test_ssl.py
@@ -2632,7 +2632,7 @@ class TestConnection(object):
         # always happen on all platforms (FreeBSD and OS X particular) for the
         # very last bit of available buffer space.
         msg = b"x"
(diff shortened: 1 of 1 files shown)


tests/test_ssl.py
```diff
@@ -14,7 +14,7 @@
 from sys import platform, getfilesystemencoding
 from socket import MSG_PEEK, SHUT_RDWR, error, socket
 from os import makedirs
-from os.path import join
+from os.path import joinwant
```
alex: Looks like an accident?

catern:
Oops, fixed.

PR #787, opened 2018-08-23
X509Store.add_cert no longer raises an error on duplicate cert

reaperhulk:
OpenSSL 1.1.0i changed X509_STORE_add_cert such that it no longer raises an error if a duplicate cert is added. This patch makes it so we behave consistently (and do not error) on all versions.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -11,7 +11,8 @@ The third digit is only for regressions.
 Backward-incompatible changes:
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-*none*
+- ``X509Store.add_cert`` no longer raises an error if you add a duplicate cert.
+  `#787 <https://github.com/pyca/pyopenssl/pull/787>`_
 
 
 Deprecations:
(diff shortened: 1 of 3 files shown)


src/OpenSSL/crypto.py
```diff
@@ -1607,7 +1607,15 @@ def add_cert(self, cert):
...
 
-        _openssl_assert(_lib.X509_STORE_add_cert(self._store, cert._x509) != 0)
+        # As of OpenSSL 1.1.0i adding the same cert to the store more than
+        # once doesn't cause an error. Accordingly, this code now silences
+        # the error for OpenSSL < 1.1.0i as well.
+        if _lib.X509_STORE_add_cert(self._store, cert._x509) != 0:
+            code = _lib.ERR_get_error()
+            err_reason = _lib.ERR_GET_REASON(code)
+            _openssl_assert(
+                err_reason == _lib.X509_R_CERT_ALREADY_IN_HASH_TABLE
```
alex: Are there any other errors that can happen?

CHANGELOG.rst
```diff
@@ -23,7 +23,8 @@ Deprecations:
 Changes:
 ^^^^^^^^
 
-*none*
+- ``X509Store.add_cert`` no longer raises an error if you add a duplicate cert.
```
alex: This should be in the backwards incompat section

PR #804, opened 2019-01-08
Handle NULL bytes in get_components() values

romuald:
Some old software may generate "bogus" CN with each character preceded
by a NULL.

This is already handled in commonName, but wasn't in get_components()


romuald:
(2 failed tests here, but they are the same as the tests failing on the master branch)

reaperhulk:
You'll need to use `b'CN'` so that the test can work on both python 2 and 3. Additionally, this can be simplified to: `value = _ffi.buffer(_lib.ASN1_STRING_data(fval), _lib.ASN1_STRING_length(fval))[:]` rather than unpack call.

romuald:
@reaperhulk done. I'm not familiar with FFI yet thanks :)

reaperhulk:
Thanks for the quick update. Now the maintainers (that includes me, sigh) need to fix the other tests so we can get this merged!

PR #818, opened 2019-02-15
Raise an Error with "no cipher match" even with TLS 1.3

markrwilliams:
This makes Twisted's OpenSSLAcceptableCiphers.fromOpenSSLCipherString
and seamlessly work with TLS 1.3:

https://github.com/twisted/twisted/pull/1100/files/a5df2fb373ac67b0e3032acc9291ae88dfd0b3b1#diff-df501bac724aab523150498f84749b88R1767

### DIFF
--- a/tests/test_ssl.py
+++ b/tests/test_ssl.py
@@ -410,18 +410,13 @@ class TestContext(object):
 
         assert "AES128-SHA" in conn.get_cipher_list()
 
-    @pytest.mark.parametrize("cipher_list,error", [
-        (object(), TypeError),
-        ("imaginary-cipher", Error),
-    ])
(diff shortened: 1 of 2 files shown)


tests/test_ssl.py
```diff
@@ -423,6 +423,24 @@ def test_set_cipher_list_wrong_args(self, context, cipher_list, error):
         with pytest.raises(error):
             context.set_cipher_list(cipher_list)
 
+    def test_set_cipher_list_no_cipher_match(self, context):
```
alex: Instead of a new test, can you just split `test_set_cipher_list_wrong_args` into two, since this really duplicates half of that.

PR #827, opened 2019-04-12
Make test_ssl pass in an IPv6-only environment

davidben:


tests/test_ssl.py
```diff
@@ -101,6 +101,21 @@
 skip_if_py3 = pytest.mark.skipif(PY3, reason="Python 2 only")
 
 
+def socket_any_family():
+    try:
+        return socket(AF_INET)
+    except error:
```
alex: I'm sure I'll regret asking, is there any way for us to narrow the exception down so we don't handle other failures?

tests/test_ssl.py
```diff
@@ -101,6 +101,21 @@
...
+        return socket(AF_INET6)
+
+
+def loopback_address(socket):
+    if socket.family == AF_INET:
+        return "127.0.0.1"
+    if socket.family == AF_INET6:
+        return "::1"
+    raise ValueError("Unknown socket family")
```
alex: To make coverage happy you can write this:

```
if socket.family == AF_INET:
    return "127.0.0.1"
else:
    assert socket.family == AF_INET6
    return "::1"
```

my kingdom for proper enums 🤷‍♂️

tests/test_ssl.py
alex: Yes, this should have an `errno` attribute, let's check that it's the right one.

davidben: I take it Python on Windows will try to map the errno and such?

alex: I think so. I've definitely written a lot of code that does `.errno == ENOENT` and no one has complained 😬

davidben: Done.

tests/test_ssl.py
davidben: Done.

tests/test_ssl.py
```diff
@@ -101,6 +101,23 @@
 skip_if_py3 = pytest.mark.skipif(PY3, reason="Python 2 only")
 
 
+def socket_any_family():
+    try:
+        return socket(AF_INET)
+    except error as e:
+        if e.errno == EAFNOSUPPORT:
+            return socket(AF_INET6)
+        raise e
```
alex:
```suggestion
        raise
```

this will preserve the stack trace better.

tests/test_ssl.py
```diff
@@ -10,9 +10,9 @@
 import uuid
 
 from gc import collect, get_referrers
-from errno import ECONNREFUSED, EINPROGRESS, EWOULDBLOCK, EPIPE, ESHUTDOWN
+from errno import EAFNOSUPPORT, ECONNREFUSED, EINPROGRESS, EWOULDBLOCK, EPIPE, ESHUTDOWN
```
alex: Eeep, line too long, can you wrap this with `()`

davidben: Done.

PR #828, opened 2019-04-12
Make tests pass after 2019

### LINKED ISSUE #735: certificate verify fails in 2020-01-01
For [reproducible builds](https://reproducible-builds.org/) we need to test if software builds still produce the same result when building it later. However, for pyOpenSSL this is currently not possible, because parts of the testsuite fail with
```
Error: [('SSL routines', 'tls_process_server_certificate', 'certificate verify failed')]
```

failing tests are
TestX509StoreContext.test_valid
TestX509StoreContext.test_reuse
TestX509StoreContext.test_modification_pre_verify
TestContext.test_add_extra_chain_cert
TestContext.test_use_certificate_chain_file_bytes
TestContext.test_use_certificate_chain_file_unicode

Also, independently of that, as a Linux distribution, we want to be able to ship updated versions of packages with added patches (even when upstream support expired) for up to 15 years.

possible approaches:
* generate CA and other certs on demand so they will be valid at the time of test
* check the current time and expect those tests to fail
* check the current time and skip those tests

bmwiedemann:
Without this patch, `TestX509StoreContext.test_valid` and 5 other tests
would fail after 2020-01-01

Fixes #735

This PR was done while working on reproducible builds for openSUSE.

Without the later commits, tests were failing in 2037 from pem files embedded in `tests/test_crypto.py`
as can be seen by pasting them into `openssl x509 -text | grep Not.After`

```
intermediate_cert_pem:         Not After : Aug 25 02:04:08 2024 GMT
intermediate_server_cert_pem:  Not After : Aug 25 02:10:48 2024 GMT
root_cert_pem:                 Not After : Jun  6 22:32:59 2037 GMT
server_cert_pem:               Not After : Jun  7 00:10:57 2037 GMT
```


bmwiedemann:
Added commits to make tests pass in 2037.

bmwiedemann:
Why is there always some unrelated travis check failing? How is someone supposed to get anything done in such an environment?

reaperhulk:
pyOpenSSL is a widely used project and as such it has some downstream tests for prominent packages to make sure changes do not break them. urllib3 is one such package. Unfortunately, its test suite is somewhat flaky, but we have judged the value of keeping it in the test suite to be higher than the cost. Repository owners can trigger retries but PR authors can also push empty commits to cause it to rebuild in the case of flakiness like that. I restarted the job and it passed, so there you go.

Of course, your change fails the check-manifest job for legitimate reasons which you should look into.

bmwiedemann:
Thanks for the hint. The manifest problem was not so obvious (I do not do that much python). Pushed a fix and it looks greener now.

bmwiedemann:
ping

bmwiedemann:
2020 is getting closer and I would still like to see this merged. Or is there anything else that needs improvement?

dtzWill:
Well, thanks for this PR/patch! Fixes tests that started failing now that it's 2020 :).

alex:
Just want to note that we're not ignoring this intentionally.... we're just very bad and unmotivated with respect to pyOpenSSL.

alex:
It's 2020, tests are passing, can we declare this complete?

bmwiedemann:
OK, it seems, @alex fixed it in 675534c18ab60517985fdee424837986e58ab260 for #888

PR #815, opened 2019-02-10
Travis CI: Run flake8 on Python, not on legacy Python

cclauss:
__buffer__ was both removed from Python on 1/1/2020.  These functions are both marked with `@skip_if_py3 functions` so this PR marks their undefined names with the linter directive `# noqa: F821`.

PR #831, opened 2019-05-19
Travis CI: Run flake8 on Python instead of legacy Python

cclauss:
Python 3 syntax is stricter than Python 2 syntax.

Also, __sudo: required__ no longer is...  [Travis are now recommending removing the __sudo__ tag](https://blog.travis-ci.com/2018-11-19-required-linux-infrastructure-migration).

"_If you currently specify __sudo:__ in your __.travis.yml__, we recommend removing that configuration_"

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,4 @@
 dist: xenial
-sudo: true
 language: python
 
 cache:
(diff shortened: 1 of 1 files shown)


alex:
Dupe of #815, which you also sent. Please don't send the same PR twice.

PR #838, opened 2019-05-20
Delete examples/sni/

alex:


arekm:
I wonder, what was the reason for dropping examples? (commit logs don't explain anything)

reaperhulk:
Examples that are not tested have a nasty tendency to bitrot and be worse than nothing. We deleted examples because no one had stepped up to write examples that could also be run under the test harness to ensure they continue to function.

PR #843, opened 2019-06-05
Switch to new notBefore/After APIs

neheb:
Related: https://github.com/pyca/cryptography/pull/4914

### DIFF
--- a/src/OpenSSL/crypto.py
+++ b/src/OpenSSL/crypto.py
@@ -1304,7 +1304,7 @@ class X509(object):
         if not isinstance(amount, int):
             raise TypeError("amount must be an integer")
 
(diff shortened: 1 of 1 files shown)


neheb:
Rebased. This passes.

reaperhulk:
What version of cryptography did we add these APIs? Do we need to update the minimum version?

neheb:
Beats me. I was just looking through my open PRs and fixing them up.

reaperhulk:
https://github.com/pyca/cryptography/commit/f5735cf25acd08222368a1db615bbf61d36b8007 Looks like 2.8 so we’ll need to update the minimum version to 2.8 for this.

neheb:
I assume that means the defines can be eliminated?

PR #852, opened 2019-07-08
use _ffi.from_buffer() to support bytearray

dholth:
If you try to send a dict it produces this error message: `TypeError: a bytes-like object is required, not 'dict'`

It doesn't accept `bytearray` without `from_buffer` (cffi doesn't automatically convert that type for `char*` arguments)

### DIFF
--- a/src/OpenSSL/SSL.py
+++ b/src/OpenSSL/SSL.py
@@ -15,6 +15,7 @@ from OpenSSL._util import (
     UNSPECIFIED as _UNSPECIFIED,
     exception_from_error_queue as _exception_from_error_queue,
     ffi as _ffi,
+    from_buffer as _from_buffer,
     lib as _lib,
     make_assert as _make_assert,
     native as _native,
@@ -1725,14 +1726,18 @@ class Connection(object):
         # Backward compatibility
         buf = _text_to_bytes_and_warn("buf", buf)
 
-        buf = _ffi.from_buffer(buf)
+        with _from_buffer(buf) as buf:
 
-        if len(buf) > 2147483647:
-            raise ValueError("Cannot send more than 2**31-1 bytes at once.")
+            # THE TEST MOCK HAS ZERO BYTES HERE
 
-        result = _lib.SSL_write(self._ssl, buf, len(buf))
-        self._raise_ssl_error(self._ssl, result)
+            if len(buf) > 2147483647:
+                raise ValueError("Cannot send more than 2**31-1 bytes at once.")
+
+            result = _lib.SSL_write(self._ssl, buf, len(buf))
+            self._raise_ssl_error(self._ssl, result)
+            
         return result
+
     write = send
 
     def sendall(self, buf, flags=0):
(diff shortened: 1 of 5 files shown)


reaperhulk:
Let’s put a test for a few invalid types here so we’ll know if cffi ever changes its exceptions on this.

dholth:
Like this one? https://github.com/pyca/pyopenssl/pull/852/files#diff-72cbea3491325683d8c77fbeea9dd31aR2799

dholth:
Same fix will need to apply to sendall(). Not sure about bio, since it is the other side of the encryption.

njsmith:
Technically speaking, you should call `ffi.release` on the cffi buffer after you're done with it. (Or, I think it has a context manager version too?)

The reason is: the way Python's buffer protocol works, as long as someone holds a reference to the raw data in the buffer, the original object is pinned. So for example, if you call `from_buffer` on a bytearray, and then try to resize the bytearray, you'll get an error, because if python let you do that then the buffer would start pointing into nothingness. So that's great, but it means that if we go creating a buffer inside this function, then we need to make sure that buffer is gone again when we return to the user, or else they'll be very confused when attempting to resize their bytearray randomly fails depending on whether the buffer's been garbage collected or not.

Extra annoyances:

- the ability to explicitly release buffers was only added in cffi 0.12
- it's impossible to actually test this, because currently on cpython the refcount gc rooms destructors deterministically, while on pypy they haven't actually implemented the buffer locking functionality. But I think it's still a good idea, for future proofing.

dholth:
Good news, cffi 0.8 does support with from_buffer() as ...:

dholth:
OK @njsmith how do you like this one? It has a compat no-op context manager for from_buffer.

Unfortunately from_buffer is smarter than the test mock. I moved the length test past from_buffer in one of the commits because the error message from from_buffer() "wrong type" is nicer than the error message from len() "doesn't have a length"

src/OpenSSL/SSL.py
```diff
@@ -1752,28 +1753,24 @@ def sendall(self, buf, flags=0):
...
+                result = _lib.SSL_write(
+                    self._ssl,
+                    data + total_sent,
+                    min(left_to_send, 2147483647)
+                )
+                self._raise_ssl_error(self._ssl, result)
+                total_sent += result
+                left_to_send -= result
+
+        return total_sent
```
dholth: bugfix

dholth:
OK folks, it's about as good as I'm likely to make it.

Includes a bugfix for `sendall()` which previously did not return the number of bytes written.

Includes a test to make sure `bio_write` throws not when passed bytes, bytearray.

njsmith:
I'm not very familiar with pyopenssl's internals, but FWIW this looks fine to me.

njsmith:
@dholth btw, you have a flake8 failure:

```
tests/test_ssl.py:2103:80: E501 line too long (80 > 79 characters)
```

dholth:
Missed it by that much

On Tue, Jul 9, 2019, at 5:59 PM, Nathaniel J. Smith wrote:
> @dholth <https://github.com/dholth> btw, you have a flake8 failure:

> `tests/test_ssl.py:2103:80: E501 line too long (80 > 79 characters)
`

dholth:
This likely improves performance too, with the simplified checking and doesn't-copy nature of from_buffer.

rtkbkish:
This change was a lifesaver.
We were POSTing on different threads using pypy. We kept getting Error: [('SSL routines', 'SSL3_WRITE_PENDING', 'bad write retry')] when a write was retried after a WantWriteError. 
Updating to this release cleared up the issue.

PR #856, opened 2019-07-29
Stop lawyers from emailing Hynek

hynek:
I've just got another email asking me for a copyright notice for some compliance stuff, they can ask in public, this is a public project. Also there's no reason for it to carry my name, given how little I do nowadays.


reaperhulk:
`cryptography` doesn't even list a maintainer in its setup call, maybe we should just remove it?

hynek:
Even better! ✨

hynek:
(I have removed it)

PR #859, opened 2019-07-31
change empty CRL get_revoked to empty tuple

thecubic:
resolves remaining issue in #258 by switching empty `get_revoked` return to effective empty tuple.

This was breaking functional tests in a pyOpenSSL integration.


reaperhulk:
Ugh, how did we miss reviewing this for so long. I'm concerned about changing this because while it obviously makes more sense, this has been like this for 7 years now so presumably there are people out there depending on the `None` behavior.

reaperhulk:
@thecubic after an almost comically long period of time I think we can merge this is if you can add a changelog entry noting this and linking to this PR (in backwards incompatible changes please).

alex:
CRLs have been removed from pyOpenSSL.

PR #876, opened 2019-11-18
ALPN: complete handshake without accepting a client's protocols.

markrwilliams:
The callback passed to `SSL_CTX_set_alpn_select_cb` can return
`SSL_TLSEXT_ERR_NOACK` to allow the handshake to continue without
accepting any of the client's offered protocols.

This commit introduces `NO_OVERLAPPING_PROTOCOLS`, which the Python
callback passed to `Context.set_alpn_select_callback` can return to
achieve the same thing.

It does not change the previous meaning of an empty string, which
still terminates the handshake.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -30,6 +30,8 @@ Changes:
 
 - Support ``bytearray`` in ``SSL.Connection.send()`` by using cffi's from_buffer.
   `#852 <https://github.com/pyca/pyopenssl/pull/852>`_
+- The ``OpenSSL.SSL.Context.set_alpn_select_callback`` can return a new ``NO_OVERLAPPING_PROTOCOLS`` sentinel value
+  to allow a TLS handshake to complete without an application protocol.
 
 
 ----
--- a/src/OpenSSL/SSL.py
+++ b/src/OpenSSL/SSL.py
@@ -1507,7 +1507,7 @@ class Context(object):
             one of those bytestrings to indicate the chosen protocol, the
             empty bytestring to terminate the TLS connection, or the
             :py:obj:`NO_OVERLAPPING_PROTOCOLS` to indicate that no offered
-            protocol was selected.
+            protocol was selected, but that the connection should not be aborted.
         """
         self._alpn_select_helper = _ALPNSelectHelper(callback)
         self._alpn_select_callback = self._alpn_select_helper.callback
--- a/src/OpenSSL/SSL.py
+++ b/src/OpenSSL/SSL.py
@@ -428,19 +428,7 @@ class _NpnSelectHelper(_CallbackExceptionHelper):
         )
 
 
-class _Sentinel(object):
-    """
-    A sentinel value.
-    """
-
-    def __init__(self, value):
-        self._value = value
-
-    def __repr__(self):
-        return "<Sentinel {!r}>".format(self._value)
-
-
-NO_OVERLAPPING_PROTOCOLS = _Sentinel("NO_OVERLAPPING_PROTOCOLS")
+NO_OVERLAPPING_PROTOCOLS = object()
 
 
 class _ALPNSelectHelper(_CallbackExceptionHelper):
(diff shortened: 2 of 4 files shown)


CHANGELOG.rst
```diff
@@ -26,7 +26,8 @@ Deprecations:
 Changes:
 ^^^^^^^^
 
-*none*
+- The ``OpenSSL.SSL.Context.set_alpn_select_callback`` can return a new ``NO_OVERLAPPING_PROTOCOLS`` sentinel value
+  to allow a TLS handshake to complete without an application protocol.
```
alex: This is missing a word, "the callback passed to" perhaps.

src/OpenSSL/SSL.py
```diff
@@ -427,6 +427,21 @@ def wrapper(ssl, out, outlen, in_, inlen, arg):
...
+    """
+
+    def __init__(self, value):
+        self._value = value
+
+    def __repr__(self):
+        return "<Sentinel {!r}>".format(self._value)
+
+
+NO_OVERLAPPING_PROTOCOLS = _Sentinel("NO_OVERLAPPING_PROTOCOLS")
```
alex: This is way too fancy, just use `object()`

src/OpenSSL/SSL.py
```diff
@@ -435,6 +450,10 @@ class _ALPNSelectHelper(_CallbackExceptionHelper):
     def __init__(self, callback):
         _CallbackExceptionHelper.__init__(self)
 
+        SSL_TLSEXT_ERR_OK = _lib.SSL_TLSEXT_ERR_OK
+        SSL_TLSEXT_ERR_NOACK = _lib.SSL_TLSEXT_ERR_NOACK
+        SSL_TLSEXT_ERR_ALERT_FATAL = _lib.SSL_TLSEXT_ERR_ALERT_FATAL
```
alex: Is there a reason we're mapping these to locals?

src/OpenSSL/SSL.py
```diff
@@ -1475,8 +1502,11 @@ def set_alpn_select_callback(self, callback):
 
         :param callback: The callback function.  It will be invoked with two
             arguments: the Connection, and a list of offered protocols as
-            bytestrings, e.g ``[b'http/1.1', b'spdy/2']``.  It should return
-            one of those bytestrings, the chosen protocol.
+            bytestrings, e.g ``[b'http/1.1', b'spdy/2']``.  It can return
+            one of those bytestrings to indicate the chosen protocol, the
+            empty bytestring to terminate the TLS connection, or the
+            :py:obj:`NO_OVERLAPPING_PROTOCOLS` to indicate that no offered
+            protocol was selected.
```
alex:
```suggestion
            protocol was selected, but that the connection should not be aborted.
```

tests/test_ssl.py
```diff
@@ -1960,6 +1960,42 @@ def select(conn, options):
...
+            agreeing to an application protocol by returning
+            ``NO_OVERLAPPING_PROTOCOLS``.
+            """
+            refusal_args = []
+
+            def refusal(conn, options):
+                refusal_args.append((conn, options))
+                return NO_OVERLAPPING_PROTOCOLS
+
+            client_context = Context(TLSv1_METHOD)
```
alex: Can this use `SSLv23_METHOD`?

src/OpenSSL/SSL.py
markrwilliams: I was scared about a reference cycle to `_lib` that passed through the C callback, which I assumed was the reason why the previous code returned 0 and 2 instead of these variables.

tests/test_ssl.py
markrwilliams: In this modern era almost certainly.

src/OpenSSL/SSL.py
alex: Gosh, unless there's a specific failing test I'd much prefer to write normal code.

Absent some other evidence, the usual reason code is bad in pyOpenSSL is because it's old and bad, not because it was well thought out.

tests/test_ssl.py
```diff
@@ -1960,6 +1960,42 @@ def select(conn, options):
...
+        def test_alpn_no_server_overlap(self):
+            """
+            A server can allow a TLS handshake to complete without
+            agreeing to an application protocol by returning
+            ``NO_OVERLAPPING_PROTOCOLS``.
+            """
+            refusal_args = []
+
+            def refusal(conn, options):
+                refusal_args.append((conn, options))
```
alex: You never do anything with this?

src/OpenSSL/SSL.py
```diff
@@ -452,24 +455,32 @@ def wrapper(ssl, out, outlen, in_, inlen, arg):
...
+                outbytes = callback(conn, protolist)
+                any_accepted = True
+                if outbytes is NO_OVERLAPPING_PROTOCOLS:
+                    outbytes = b''
+                    any_accepted = False
+                elif not isinstance(outbytes, _binary_type):
+                    raise TypeError(
+                        "ALPN callback must return a bytestring or the "
+                        "special NO_OVERLAPPING_PROTOCOLS sentinel value."
+                    )
```
alex: Can you add a test for this path?

Issue #872, opened 2019-10-29
No more examples?


bortzmeyer:
https://www.pyopenssl.org/en/stable/ says "There are also examples in the pyOpenSSL repository that may help you getting started" but the link (https://github.com/pyca/pyopenssl/tree/master/examples) yields a 404. I cannot find where the examples are.

alex:
That's right, we removed the examples since they had bitrotted. We should update the docs.

PR #883, opened 2019-12-14
A simple example of a HTTP client. Addresses #872

bortzmeyer:


reaperhulk:
If we're going to re-add examples we need a way to have them run in CI to prevent the bitrot we saw in past years. CI is also failing because of flake8 issues and check-manifest problems. The latter is failing since this needs to be excluded in `MANIFEST.in`.

PR #893, opened 2020-03-02
Add DHparams class and methods to dump and load

schwabe:
This add support for loading, dumping and generating Diffie-Hellman
parameters to pyOpenSSL. The interface is modelled to be similar to
the existing interfaces.

### DIFF
--- a/doc/api/crypto.rst
+++ b/doc/api/crypto.rst
@@ -72,6 +72,13 @@ Certificate revocation lists
 
 .. autofunction:: load_pkcs12
 
+Diffie-Hellman parameters
+~~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. autofunction:: dump_dhparams
+
+.. autofunction:: load_dhparams
+
 Signing and verifying signatures
 --------------------------------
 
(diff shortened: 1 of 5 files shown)


schwabe:
Note I originally also had a `check` method but the DH_check method seems to be missing in the exported OpenSSL methods. I am not familar with cffi so I did not investigate why this is the case:

```python
    def check(self):
        """
        Check the consistency of Diffie-Hellman parameters.

        This is the Python equivalent of OpenSSL's ``DH_check_``.

        :return: ``True`` if dh is consistent.

        :raise OpenSSL.crypto.Error: if the dh is inconsistent.

        :raise TypeError: if dh is of a type which cannot be checked.
            Only Diffie-Hellman parameters can currently be checked.
        """

        codes = 0 
        result = _lib.DH_check(self._dh, codes)
        if result:
            return True
        _raise_current_error()
```

reaperhulk:
In general we'd prefer users to generate DH params via the APIs in cryptography (https://cryptography.io/en/latest/hazmat/primitives/asymmetric/dh/). Is there a specific reason you want them in this project, which depends on cryptography anyway and has an API we'd prefer to slowly deprecate?

schwabe:
Main goal was to use the SSL_CTX_set_tmp_dh from the OpenSSL part of the library without needing to go through a file. But when you say deprecation, can you point me to the document part//what is being deprecated. I am sorry that I was completely oblivious to that.

schwabe:
Closing the PR as cryptography API works well enough at least for my usecase

PR #894, opened 2020-03-03
Allow accessing a connection's verfied certificate chain

### LINKED ISSUE #740: Retrieve chain after certificate validation - what should API look like?
It would be nice if we could fetch and examine the chain after calling verify_certificate (Which calls X509_verify_cert). https://pyopenssl.org/en/stable/api/crypto.html#OpenSSL.crypto.X509StoreContext)

I'm wondering what you think the API should look like. Fetch the entire chain at once and return it to the user, or allows the user to actually iterate over the chain?

Here's an example that shows how to parse the chain after a X509_verify_cert call: https://github.com/openssl/openssl/blob/master/apps/verify.c#L243

I'm willing to implement this feature - I don't think it should be too hard, but it can be quite valuable.

ShaneHarvey:
Add X509StoreContext.get_verified_chain using X509_STORE_CTX_get1_chain.
Add Connection.get_verified_chain using SSL_get0_verified_chain if
available (ie OpenSSL 1.1+) and X509StoreContext.get_verified_chain
otherwise.

Motivation: [PyMongo](https://github.com/mongodb/mongo-python-driver/) is implementing OCSP support and we're using PyOpenSSL to do so (thanks!). As part of the client's OCSP callback we would ideally have access to the **verified** peer certificate chain to find the peer's issuer cert. This change allows us to do that by exposing the verified peer certificate chain.

References:
- https://www.openssl.org/docs/man1.1.0/man3/X509_STORE_CTX_get1_chain.html
- https://www.openssl.org/docs/man1.1.0/man3/SSL_get0_verified_chain.html
- The MongoDB server's polyfill implementation of SSL_get0_verified_chain for OpenSSL 1.0.1/1.0.2: https://github.com/mongodb/mongo/blob/r4.4.0/src/mongo/util/net/ssl_manager_openssl.cpp#L312-L327

Fixes #740.

### DIFF
--- a/src/OpenSSL/SSL.py
+++ b/src/OpenSSL/SSL.py
@@ -2168,7 +2168,7 @@ class Connection(object):
 
         .. versionadded:: 20.0
         """
-        if hasattr(_lib, 'SSL_get0_verified_chain'):
+        if hasattr(_lib, "SSL_get0_verified_chain"):
             # OpenSSL 1.1+
             cert_stack = _lib.SSL_get0_verified_chain(self._ssl)
             if cert_stack == _ffi.NULL:
--- a/src/OpenSSL/SSL.py
+++ b/src/OpenSSL/SSL.py
@@ -28,6 +28,7 @@ from OpenSSL.crypto import (
     X509Name,
     X509,
     X509Store,
+    X509StoreContext,
 )
 
 __all__ = [
@@ -2144,6 +2145,48 @@ class Connection(object):
             pycert = X509._from_raw_x509_ptr(cert)
             result.append(pycert)
         return result
+
+    def get_verified_chain(self):
+        """
+        Retrieve the verified certificate chain of the peer including the
+        peer's end entity certificate. It must be called after a session has
+        been successfully established. If peer verification was not successful
+        the chain may be incomplete, invalid, or None.
+
+        :return: A list of X509 instances giving the peer's verified
+                 certificate chain, or None if it does not have one.
+
+        .. versionadded:: 20.0
+        """
+        if hasattr(_lib, 'SSL_get0_verified_chain'):
+            # OpenSSL 1.1+
+            cert_stack = _lib.SSL_get0_verified_chain(self._ssl)
+            if cert_stack == _ffi.NULL:
+                return None
+
+            result = []
+            for i in range(_lib.sk_X509_num(cert_stack)):
+                # TODO could incref instead of dup here
+                cert = _lib.X509_dup(_lib.sk_X509_value(cert_stack, i))
+                pycert = X509._from_raw_x509_ptr(cert)
+                result.append(pycert)
+            return result
+
+        pycert = self.get_peer_certificate()
+        if pycert is None:
+            return None
+
+        cert_stack = _lib.SSL_get_peer_cert_chain(self._ssl)
+        if cert_stack == _ffi.NULL:
+            return None
+
+        pystore = self._context.get_cert_store()
+        if pystore is None:
+            return None
+
+        pystorectx = X509StoreContext(pystore, pycert)
+        pystorectx._add_chain(cert_stack)
+        return pystorectx.get_verified_chain()
 
     def want_read(self):
         """
--- a/src/OpenSSL/SSL.py
+++ b/src/OpenSSL/SSL.py
@@ -2127,6 +2127,22 @@ class Connection(object):
             return X509._from_raw_x509_ptr(cert)
         return None
 
+    @staticmethod
+    def _cert_stack_to_list(cert_stack):
+        """
+        Internal helper to convert a STACK_OF(X509) to a list of X509
+        instances.
+        """
+        result = []
+        for i in range(_lib.sk_X509_num(cert_stack)):
+            cert = _lib.sk_X509_value(cert_stack, i)
+            _openssl_assert(cert != _ffi.NULL)
+            res = _lib.X509_up_ref(cert)
+            _openssl_assert(res >= 1)
+            pycert = X509._from_raw_x509_ptr(cert)
+            result.append(pycert)
+        return result
+
     def get_peer_cert_chain(self):
         """
         Retrieve the other side's certificate (if any)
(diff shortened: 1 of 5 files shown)


ShaneHarvey:
@alex/@reaperhulk will this PR ever be considered? I'd be happy to rebase it.

src/OpenSSL/SSL.py
```diff
@@ -2243,6 +2244,48 @@ def get_peer_cert_chain(self):
...
+        if hasattr(_lib, 'SSL_get0_verified_chain'):
+            # OpenSSL 1.1+
+            cert_stack = _lib.SSL_get0_verified_chain(self._ssl)
+            if cert_stack == _ffi.NULL:
+                return None
+
+            result = []
+            for i in range(_lib.sk_X509_num(cert_stack)):
+                # TODO could incref instead of dup here
+                cert = _lib.X509_dup(_lib.sk_X509_value(cert_stack, i))
```
reaperhulk, line 2268: Yeah let's `X509_up_ref` this instead of duping. Be sure to assert on the result for any call you make (X509_up_ref returns 1 on 1.1.0+ and the refcount in 1.0.2 I believe so it should probably be `_openssl_assert(res >= 1)`.

src/OpenSSL/SSL.py
```diff
@@ -2243,6 +2244,48 @@ def get_peer_cert_chain(self):
...
+                pycert = X509._from_raw_x509_ptr(cert)
+                result.append(pycert)
+            return result
+
+        pycert = self.get_peer_certificate()
+        if pycert is None:
+            return None
+
+        cert_stack = _lib.SSL_get_peer_cert_chain(self._ssl)
+        if cert_stack == _ffi.NULL:
```
reaperhulk, line 2278: When does this happen? If it should never happen we should do `_openssl_assert`

src/OpenSSL/crypto.py
```diff
@@ -1765,6 +1772,47 @@ def verify_certificate(self):
...
+          indicate which certificate caused the error.
+
+        .. versionadded:: 20.0
+        """
+        # Always re-initialize the store context in case
+        # :meth:`verify_certificate` is called multiple times.
+        #
+        # :meth:`_init` is called in :meth:`__init__` so _cleanup is called
+        # before _init to ensure memory is not leaked.
+        self._cleanup()
```
reaperhulk, line 1791: What happens if you call verify_certificate multiple times without reinitialization?

src/OpenSSL/crypto.py
```diff
@@ -1765,6 +1772,47 @@ def verify_certificate(self):
...
+        self._init()
+        ret = _lib.X509_verify_cert(self._store_ctx)
+        if ret <= 0:
+            self._cleanup()
+            raise self._exception_from_context()
+
+        # Note: X509_STORE_CTX_get1_chain returns a deep copy of the chain.
+        cert_stack = _lib.X509_STORE_CTX_get1_chain(self._store_ctx)
+        if cert_stack == _ffi.NULL:
+            # TODO: This is untested.
```
reaperhulk, line 1801: If we can't trigger this branch use `_openssl_assert`

src/OpenSSL/crypto.py
```diff
@@ -1765,6 +1772,47 @@ def verify_certificate(self):
...
+            self._cleanup()
+            return None
+
+        result = []
+        for i in range(_lib.sk_X509_num(cert_stack)):
+            cert = _lib.sk_X509_value(cert_stack, i)
```
reaperhulk, line 1807: Any call like this should also have a `_openssl_assert(cert != _ffi.NULL)` please

tests/test_ssl.py
```diff
@@ -2522,6 +2522,63 @@ def test_get_peer_cert_chain_none(self):
...
 
+    def test_get_verified_chain(self):
+        """
+        `Connection.get_verified_chain` returns a list of certificates
+        which the connected server returned for the certification verification.
+        """
+        chain = _create_certificate_chain()
+        [(cakey, cacert), (ikey, icert), (skey, scert)] = chain
+
+        serverContext = Context(TLSv1_METHOD)
```
reaperhulk, line 2533: Could you change all these to `SSLv23_METHOD`? `TLSv1_METHOD` won't work in Ubuntu 20.04 (which is what our CI just switched to) because TLS 1.0 is disabled. `SSLv23_METHOD` is an awful name but will allow up to TLS 1.3.

reaperhulk:
For a little context here, we're generally reluctant to add new features to pyOpenSSL because we don't have the time/bandwidth to review things effectively and generally we want people using `pyca/cryptography` where possible. However, `pyca/cryptography` has no intention of having a TLS API any time soon so features like this are appropriate to build and review here. Even if we only review when we're feeling exceedingly guilty...

reaperhulk:
Tests are failing right now

reaperhulk:
Please also add a CHANGELOG.rst entry for this new feature!

src/OpenSSL/SSL.py
```diff
@@ -2137,13 +2154,43 @@ def get_peer_cert_chain(self):
...
+        # Should never be NULL because the peer presented a certificate.
+        cert_stack = _lib.SSL_get_peer_cert_chain(self._ssl)
+        _openssl_assert(cert_stack != _ffi.NULL)
+
+        pystore = self._context.get_cert_store()
+        if pystore is None:
+            return None
+
+        pystorectx = X509StoreContext(pystore, pycert)
+        pystorectx._add_chain(cert_stack)
```
reaperhulk, line 2192: Let's just set `_chain` directly and remove the setter since it doesn't hold any logic.

src/OpenSSL/crypto.py
```diff
@@ -1712,11 +1712,18 @@ def __init__(self, store, certificate):
         self._store_ctx = _ffi.gc(store_ctx, _lib.X509_STORE_CTX_free)
         self._store = store
         self._cert = certificate
+        self._chain = _ffi.NULL
         # Make the store context available for use after instantiating this
         # class by initializing it now. Per testing, subsequent calls to
         # :meth:`_init` have no adverse affect.
         self._init()
 
+    def _add_chain(self, chain):
```
reaperhulk, line 1721: This can be removed once the other comment is addressed.

src/OpenSSL/crypto.py
reaperhulk: Okay. This feels inelegant but that describes all of pyOpenSSL. I suspect if I asked why not just never init until you need it that I'd find out these objects can be passed to other parts of pyOpenSSL which emphatically expect them to have been init'd already.

src/OpenSSL/crypto.py
ShaneHarvey: Done.

src/OpenSSL/SSL.py
ShaneHarvey: Done.

reaperhulk:
Thanks for working with us!

PR #897, opened 2020-03-11
Fix PKey.check for some broken keys

Mrmaxmeier:
RSA_check_key is documented to return 1 for valid keys.
It (currently) returns 0 or -1 for invalid ones.
The previous code accepted invalid keys if RSA_check_key returns -1!


alex:
Can you include a testcase for this?

Mrmaxmeier:
Thanks for the quick response! Would it be reasonable to issue a patch release for this?
The issue is somewhat security-sensitive and a PyPI release would simplify things for me :slightly_smiling_face:

PR #907, opened 2020-04-06
Fixing issue #798

### LINKED ISSUE #798: Using X509Store's set_time function raises an ValueError exception on Windows
Hi there,

when trying to use `X509Store`'s `set_time()` function I get an exception on my Windows machine:
```
self = <OpenSSL.crypto.X509Store object at 0x0000000004B9EEF0>
vfy_time = datetime.datetime(2018, 10, 10, 15, 47, 49, 836)
 
    def set_time(self, vfy_time):
        """
            Set the time against which the certificates are verified.
 
            Normally the current time is used.
 
            .. note::
 
              For example, you can determine if a certificate was valid at a given
              time.
 
            .. versionadded:: 17.0.0
 
            :param datetime vfy_time: The verification time to set on this store.
            :return: ``None`` if the verification time was successfully set.
            """
        param = _lib.X509_VERIFY_PARAM_new()
        param = _ffi.gc(param, _lib.X509_VERIFY_PARAM_free)
 
>       _lib.X509_VERIFY_PARAM_set_time(param, int(vfy_time.strftime('%s')))
E       ValueError: Invalid format string
 
C:\Program Files\Python36\lib\site-packages\OpenSSL\crypto.py:1671: ValueError
```

According to this Stackoverflow post, using `strftime('%s')` on a Python `datetime.datetime` object is a rather bad idea: https://stackoverflow.com/a/41811725

The `%s` directive also is not included in Python's `strftime()` documentation, however for some reason it still works on Unix based systems: https://docs.python.org/3.7/library/datetime.html#strftime-strptime-behavior 

Best regards,
Patrick

jalberdi004:
Fixing issue #798, thanks to @reaperhulk; removing undocumented '%s' option to convert the date to an `int` and getting the date in a more robust way.

Closes #798


src/OpenSSL/crypto.py
```diff
@@ -1641,10 +1641,11 @@ def set_time(self, vfy_time):
         :param datetime vfy_time: The verification time to set on this store.
         :return: ``None`` if the verification time was successfully set.
         """
+        import calendar
```
alex, line 1644: imports should go at teh top of the file

setup.py
```diff
@@ -45,15 +45,7 @@ def find_meta(meta):
...
-LONG = (
-    read_file("README.rst") + "\n\n" +
-    "Release Information\n" +
-    "===================\n\n" +
-    re.search(r"(\d{2}.\d.\d \(.*?\)\n.*?)\n\n\n----\n",
-              read_file("CHANGELOG.rst"), re.S).group(1) +
-    "\n\n`Full changelog " +
-    "<{uri}en/stable/changelog.html>`_.\n\n"
-).format(uri=URI)
+LONG = 'nothing relevant'
```
alex, line 48: Why was this change made?

jalberdi004: Not intended for the PR (I'll remove it as soon as possible).

The error occurs while installing PyOpenSSL in Windows machines (I tried in 3 different machines). The command that fails is:
```bash
python setup.py --install
```
With the output:
```
File "<string>", line 1, in <module>
File "C:\Users\USER\pyopenssl\setup.py", line 54, in <module>
"\n\n`Full changelog " +
AttributeError: 'NoneType' object has no attribute 'group'
```

src/OpenSSL/crypto.py
```diff
@@ -1,4 +1,5 @@
 import datetime
+import calendar
```
alex, line 2:
```suggestion
import calendar
import datetime
```

src/OpenSSL/crypto.py
jalberdi004: Moved

alex (dismissed):

jalberdi004:
There are python2.7 failures. But those are happening before tests are started, in setup.py
https://travis-ci.org/github/pyca/pyopenssl/jobs/671712647
https://travis-ci.org/github/pyca/pyopenssl/jobs/671712626
```
ERROR:   flake8: commands failed

The command "if [[ "$(uname -s)" == 'Darwin' ]]; then

  # set our flags to use homebrew openssl

  if [[ "${OPENSSL}"  == "1.1.1" ]]; then

    export LDFLAGS="-L/usr/local/opt/openssl@1.1/lib"

    export CFLAGS="-I/usr/local/opt/openssl@1.1/include"

    export PATH="/usr/local/opt/openssl@1.1/bin:$PATH"

  else

    export LDFLAGS="-L/usr/local/opt/openssl/lib"

    export CFLAGS="-I/usr/local/opt/openssl/include"

    export PATH="/usr/local/opt/openssl/bin:$PATH"

  fi
```

alex:
the flake8 issue is real, the failing test is related to cryptography dropping openssl 1.0.1. we'll have to solve that outside of this PR.

jalberdi004:
Good, thank you for the info

reaperhulk:
Since this was originally done the CI has been changed significantly and we black'd the codebase. If you could rebase this we can get it merged though!

jalberdi004:
It fails with the following error (https://travis-ci.com/github/pyca/pyopenssl/jobs/401351492): 
```
flake8 run-test: commands[0] | black --check .

[5934] /home/travis/build/pyca/pyopenssl$ /home/travis/build/pyca/pyopenssl/.tox/flake8/bin/black --check .

would reformat /home/travis/build/pyca/pyopenssl/src/OpenSSL/crypto.py

Oh no! 💥 💔 💥

1 file would be reformatted, 23 files would be left unchanged.

ERROR: InvocationError for command /home/travis/build/pyca/pyopenssl/.tox/flake8/bin/black --check . (exited with code 1)

___________________________________ summary ____________________________________

ERROR:   flake8: commands failed

The command "~/.venv/bin/tox -v" exited with 1.
```

I don't understand what does it mean and how to fix it.

alex:
All the python files int his repo are formatted with black. That error is saying that one of these files is not correctly formatted.

jalberdi004:
Now is solved the problem with formatting

reaperhulk (approved):

orosam:
Unfortunately, `timegm(timetuple())` also disregards time zones, just the other way around. 
Let me demonstrate with a small script (best results with python3, plus `python-dateutil` is also needed):

```python
from __future__ import print_function

from calendar import timegm
from datetime import datetime
from os import putenv

from dateutil import tz

TZ = "CET"  # it's UTC + 2 at the tested time below
putenv("TZ", TZ)

tzs = [
    ("UTC", tz.gettz("UTC")),
    ("Naive (local zone is {})".format(TZ), None),
    ("EEST", tz.gettz("EET")),  # It will be UTC + 3 hours at the tested time 
]

dates = [(name, datetime(2020, 8, 1, 11, 0, 0, tzinfo=tzinfo)) for name, tzinfo in tzs]

for name, date in dates:
    print("{}: {!r}".format(name, date))

    try:
        print("  python3 timestamp()        : {}".format(int(date.timestamp())))
    except AttributeError:
        pass

    print("  timestamp with timetuple   : {} ({!r})".format(timegm(date.timetuple()), date.timetuple()))
    print("  timestamp with utctimetuple: {} ({!r})".format(timegm(date.utctimetuple()), date.utctimetuple()))
    print("  timestamp with old strftime: {}".format(date.strftime("%s")))
    print()
```

Result:
```
UTC: datetime.datetime(2020, 8, 1, 11, 0, tzinfo=tzfile('/usr/share/zoneinfo/UTC'))
  python3 timestamp()        : 1596279600
  timestamp with timetuple   : 1596279600 (time.struct_time(tm_year=2020, tm_mon=8, tm_mday=1, tm_hour=11, tm_min=0, tm_sec=0, tm_wday=5, tm_yday=214, tm_isdst=0))
  timestamp with utctimetuple: 1596279600 (time.struct_time(tm_year=2020, tm_mon=8, tm_mday=1, tm_hour=11, tm_min=0, tm_sec=0, tm_wday=5, tm_yday=214, tm_isdst=0))
  timestamp with old strftime: 1596276000

Naive (local zone is CET): datetime.datetime(2020, 8, 1, 11, 0)
  python3 timestamp()        : 1596272400
  timestamp with timetuple   : 1596279600 (time.struct_time(tm_year=2020, tm_mon=8, tm_mday=1, tm_hour=11, tm_min=0, tm_sec=0, tm_wday=5, tm_yday=214, tm_isdst=-1))
  timestamp with utctimetuple: 1596279600 (time.struct_time(tm_year=2020, tm_mon=8, tm_mday=1, tm_hour=11, tm_min=0, tm_sec=0, tm_wday=5, tm_yday=214, tm_isdst=0))
  timestamp with old strftime: 1596272400

EEST: datetime.datetime(2020, 8, 1, 11, 0, tzinfo=tzfile('/usr/share/zoneinfo/EET'))
  python3 timestamp()        : 1596268800
  timestamp with timetuple   : 1596279600 (time.struct_time(tm_year=2020, tm_mon=8, tm_mday=1, tm_hour=11, tm_min=0, tm_sec=0, tm_wday=5, tm_yday=214, tm_isdst=1))
  timestamp with utctimetuple: 1596268800 (time.struct_time(tm_year=2020, tm_mon=8, tm_mday=1, tm_hour=8, tm_min=0, tm_sec=0, tm_wday=5, tm_yday=214, tm_isdst=0))
  timestamp with old strftime: 1596272400
```

So while the old "strftime(%s)" always treated `vfy_time` as a datetime in local time, the current implementation always treats it as a timestamp in UTC. Both disregards time zones.

Using `utctimetuple()` could be better, although it would also treat timezone-naive datetime objects as being in UTC, which might be surprising. But that could be documented.

PR #910, opened 2020-04-24
Add SSL.Context.set_keylog_callback

mhils:
This builds on the bindings added in https://github.com/pyca/cryptography/pull/5187. :)

The PR should be ready, but this depends on the next minor cryptography release.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -24,7 +24,8 @@ Deprecations:
 Changes:
 ^^^^^^^^
 
-*none*
+- Added ``Context.set_keylog_callback`` to log key material.
+  `#910 <https://github.com/pyca/pyopenssl/pull/910>`_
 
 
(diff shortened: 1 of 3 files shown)


rugk:
https://github.com/pyca/cryptography/pull/5187 ha sbeen merged. What is blocking this?

mhils:
There's no cryptography release that includes https://github.com/pyca/cryptography/pull/5187 yet. 

Tests are passing here because `Cryptography_HAS_KEYLOG` always returns false and the actual code is never tested (as indicated by the decrease in code coverage).

mhils:
I've rebase the PR onto master, all tests are passing now with cryptography 3.0 - thanks! 🍰 

Codecov is complaining about one missed line, but that seems to be a false-positive to me.

mhils:
Yes, we suffer from that too. Thanks for the super quick merge again! ❤️

PR #917, opened 2020-06-24
Fix generated test X.509 certificates.

davidben:
From RFC 5280, section 4.1.2.9:

>   [Extensions] MUST only appear if the version is 3 (Section 4.1.2.1).
>   If present, this field is a SEQUENCE of one or more certificate
>   extensions.  The format and content of certificate extensions in the
>   Internet PKI are defined in Section 4.2.

X509 objects default to v1, so the test certs need a set_version(2) call. (Note v3 is encoded as 2.)


davidben:
Well, v3, but yeah the world needs better certificate-building APIs. :-(

PR #921, opened 2020-07-25
Deprecated pkcs7 and pkcs12 support

alex:


src/OpenSSL/crypto.py
```diff
@@ -3010,6 +3010,14 @@ def load_pkcs7_data(type, buffer):
     return pypkcs7
 
 
+load_pkcs7_data = utils.deprecated(
+    load_pkcs7_data,
+    __name__,
+    "PKCS#7 support in pyOpenSSL is deprecated.",
```
reaperhulk, line 3016: We should update this to say the same about cryptography's APIs once we merge support.

alex: Do you want to wait until PKCS7 is merged into pyca/cryptography then?

reaperhulk: Eeeh, I don't care too much. We'll get that merged pretty quickly so if you want to do it now that's fine.

PR #922, opened 2020-08-03
use larger keys in ssl tests

alex:


tiran:
2048 is good enough for OpenSSL [security level](https://www.openssl.org/docs/manmaster/man3/SSL_CTX_get_security_level.html) 2. ``@SECLEVEL=3`` requires 3072bit RSA. Most distros (Debian, RHEL) have adopted ``@SECLEVEL=2``, so this should be good enough for a while.

reaperhulk:
We have several other changes that need to be made here, but this is a good start.

PR #927, opened 2020-08-03
update cert fixtures and simplify tests

reaperhulk:
This regenerates the key/cert hierarchy we use with 3072-bit RSA keys and removes the cleartextCertificatePEM/key since they were literally identical to root_cert_pem and root_cert_key. That necessitated going through and updating the file name in a bunch of tests.


tests/test_crypto.py
```diff
@@ -2809,39 +2965,33 @@ def test_dump_certificate_bad_type(self):
...
         """
-        key = load_privatekey(FILETYPE_PEM, cleartextPrivateKeyPEM)
-        dumped_pem = dump_privatekey(FILETYPE_PEM, key)
+        key = load_privatekey(FILETYPE_PEM, root_key_pem)
 
         dumped_der = dump_privatekey(FILETYPE_ASN1, key)
-        # XXX This OpenSSL call writes "writing RSA key" to standard out.  Sad.
-        good_der = _runopenssl(dumped_pem, b"rsa", b"-outform", b"DER")
-        assert dumped_der == good_der
-        key2 = load_privatekey(FILETYPE_ASN1, dumped_der)
```
alex, line 2836: This was the only call to `load_privatekey` with `FILETYPE_ASN1`  in the tests. So now it's uncovered. lol

reaperhulk: These tests are so bad.

PR #929, opened 2020-08-03
focal time

reaperhulk:


tests/test_ssl.py
```diff
@@ -1560,7 +1563,7 @@ def _load_tmp_dh_test(self, dhfilename):
             dhfile.write(dhparam)
 
         context.load_tmp_dh(dhfilename)
-        # XXX What should I assert here? -exarkun
+        assert _lib.ERR_peek_error() == 0
```
alex, line 1566: This doesn't seem right. What is it for?

reaperhulk: `load_tmp_dh` started failing because the DH key was too small on focal and no assertion was present in the tests so it wasn't caught properly. We should more correctly pull in the make_assert func

alex: The right fix is that `load_tmp_dh` should actually check the return code of `_lib.SSL_CTX_set_tmp_dh`, not one assert in a test.

PR #933, opened 2020-08-07
Context.set_verify: allow omission of callback

mhils:
This PR allows it to call `set_verify` without specifying an explicit callback:

> If no verify_callback is specified, the default callback will be used. Its return value is identical to preverify_ok, so that any verification failure will lead to a termination of the TLS/SSL handshake with an alert message, if SSL_VERIFY_PEER is set.

This simplifies the most common case to `context.set_verify(SSL.VERIFY_PEER)`. Now that OpenSSL has builtin hostname validation, it's probably a good thing to discourage people from writing custom certificate verification.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -29,6 +29,8 @@ Changes:
 - Added ``OpenSSL.SSL.Connection.get_verified_chain`` to retrieve the
   verified certificate chain of the peer.
   `#894 <https://github.com/pyca/pyopenssl/pull/894>`_.
+- Allow omission of the verification callback in ``Context.set_verify``.
+  `#933 <https://github.com/pyca/pyopenssl/pull/933>`_
 
 
 19.1.0 (2019-11-18)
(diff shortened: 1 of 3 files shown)


reaperhulk:
Based on the tests this appears to be true for every openssl in our CI. Do you happen to know when OpenSSL started defaulting to this behavior?

mhils:
I don't, but going by the docs this has been the case since at least 1.0.2: https://www.openssl.org/docs/man1.0.2/man3/SSL_CTX_set_verify.html
The next pyOpenSSL release is supposed to drop 1.0.1, so we should be good here?

reaperhulk:
Yep, 1.0.2 is minimum supported going forward so that’s fine.

CHANGELOG.rst
```diff
@@ -29,6 +29,8 @@ Changes:
 - Added ``OpenSSL.SSL.Connection.get_verified_chain`` to retrieve the
   verified certificate chain of the peer.
   `#894 <https://github.com/pyca/pyopenssl/pull/894>`_.
+- Allow omission of the verification callback in ``Context.set_verify``.
```
reaperhulk, line 32: One last request, please document that this will use openssl’s default verification both here and anywhere else we need.

mhils: done!

mhils:
Thanks for the super quick merge! ❤️

PR #943, opened 2020-09-08
Allow using an OpenSSL hashed directory for verification in X509Store

orosam:
Add `X509Store.load_locations()` to set a CA bundle file and/or an OpenSSL-style hashed CA/CRL lookup directory. 
This exposes `X509_STORE_load_locations()` from OpenSSL, similar to what the already existing `SSL.Context.load_verify_locations()` does for `SSL_CTX_load_verify_locations()`.

### DIFF
--- a/src/OpenSSL/crypto.py
+++ b/src/OpenSSL/crypto.py
@@ -19,6 +19,7 @@ from OpenSSL._util import (
     exception_from_error_queue as _exception_from_error_queue,
     byte_string as _byte_string,
     native as _native,
+    path_string as _path_string,
     UNSPECIFIED as _UNSPECIFIED,
     text_to_bytes_and_warn as _text_to_bytes_and_warn,
     make_assert as _make_assert,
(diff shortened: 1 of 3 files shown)


src/OpenSSL/crypto.py
```diff
@@ -1674,6 +1675,39 @@ def set_time(self, vfy_time):
...
+        ``c_rehash`` tool included with OpenSSL.  Either, but not both, of
+        *cafile* or *capath* may be ``None``.
+
+        :param cafile: In which file we can find the certificates (``bytes`` or
+                       ``unicode``).
+        :param capath: In which directory we can find the certificates
+                       (``bytes`` or ``unicode``).
+
+        :return: ``None`` if the locations were set successfully.
+        """
```
reaperhulk, line 1694: What happens if both are NULL or both are provided?

tests/test_crypto.py
```diff
@@ -3884,6 +3949,38 @@ def test_get_verified_chain_invalid_chain_no_root(self):
         assert exc.value.args[0][2] == "unable to get issuer certificate"
         assert exc.value.certificate.get_subject().CN == "intermediate"
 
+    @pytest.fixture
+    def ca_file(self, tmpdir):
```
reaperhulk, line 3953: I'm not particularly familiar with pytest's tmpdir fixture. Are these automatically cleaned up as part of the test suite?

src/OpenSSL/crypto.py
orosam: When both are None, the `X509_STORE_load_locations()` call will return 0, so an `OpenSSL.crypto.Error` will be raised. It's not very helpful, a `ValueError` might have been better, by checking the args explicitly. I did not want to diverge from `OpenSSL.SSL.Context.load_verify_locations()`, which behaves similarly, so I kept it this way. But since `OpenSSL.SSL.Error` is different from `OpenSSL.crypto.Error` anyway, I could add the check and raise a ValueError.

When both are provided, both "locations" will be used. Even better, `load_locations` can be called multiple times, to add more and more locations. Internally, OpenSSL just adds them to the store as `X509_LOOKUP`s. Should I update the docstring with this info?

tests/test_crypto.py
orosam: Yes, in a way. This fixture creates a unique temporary directory(*) for each test function invocation using it (so e.g. a new one for every parametric call), and performs a cleanup at the end of the pytest run. This cleanup keeps the temporary directories of the last 3 pytest runs, and deletes all others.
I admit I didn't give too much thought to this, as the `tmpdir` fixture is already used extensively in `test_ssl.py`, and indirectly via the `tmpfile` fixture defined in `conftest.py`.

(*) Technically the directory structure is like this: `pytest_base_temporary_directory/per_pytest_run_temporary_directory/per_test_invocation_temporary_directory`.)

reaperhulk: I had totally forgotten about the use in `test_ssl` because none of the maintainers of this project really like to think about it much 😂 . Thanks for the explanation anyway, this is fine.

src/OpenSSL/crypto.py
reaperhulk: Since pyOpenSSL is generally a thin wrapper around OpenSSL's API I'm fine with it raising `crypto.Error`. If you could update the docstring here then I think this is pretty much good to go. Thanks!

orosam: I've updated and also extended the docstring to mention that this method can be called multiple times, and added some tests.

One more question: I see that many methods have a `::versionadded` line. Should I add one for `load_locations()`?

orosam: (... and removed 2 empty lines in another force-push, sorry...)

reaperhulk: Yes a `.. versionadded:: 20.0` would be appropriate.

orosam: Done.

reaperhulk (approved):

PR #948, opened 2020-10-09
 Allow using additional untrusted certificates for chain building in X509StoreContext

orosam:
The additional certificates provided in the new `chain` parameter will be
untrusted but may be used to build the chain.

This makes it easier to validate a certificate against a store which
contains only root ca certificates, and the intermediates come from e.g.
the same untrusted source as the certificate to be verified.

This PR is based on the work done by @akgood for PR #473.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -24,6 +24,9 @@ Deprecations:
 Changes:
 ^^^^^^^^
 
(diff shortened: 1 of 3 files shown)


reaperhulk (approved):

orosam:
Thanks for the quick merge!

PR #952, opened 2020-10-26
Handle datetimes with time zone information in crypto.X509Store.set_time()

orosam:
#907 fixed the issue with `set_time()` not working on Windows.
It also changed `set_time()`'s behavior in an incompatible way: instead of
treating `vfy_time` always being in local time (regardless if it had a time
zone attached or not), it now treats `vfy_time` as a time in UTC.

This patch improves on that by taking the time zone into account, and also
documents the incompatible change.

Note that it is not always possible to convert a timestamp in an arbitrary
time zone into UTC unambiguously, due to repeated or skipped local times
around DST changes. The best is to use a timezone-aware `vfy_time`
using the UTC time zone.


orosam:
Two checks appear to be failing due to cryptography no longer allowing OpenSSL 1.0.2.

alex:
*flips table*

Ok, I guess we'll need to figure that out before we can review your PR. THank you for sending it!

alex:
Is it possible to write a test case for this?

PR #953, opened 2020-10-26
Drop CI for OpenSSL 1.0.2

alex:


src/OpenSSL/SSL.py
```diff
@@ -28,7 +28,6 @@
     X509Name,
     X509,
     X509Store,
-    X509StoreContext,
```
reaperhulk, line 31: Did we really get rid of this as part of 1.0.2 deprecation? Removing a whole class is a lot... 😮

alex: It's no longer needed by this module, you can import it from crypto as always.

reaperhulk: Oh oops, thought this was where it was exported for some reason.

reaperhulk:
As long as our minimum version is cryptography 2.8 (https://github.com/pyca/pyopenssl/blob/master/tox.ini#L13) it seems like we should be supporting 1.0.2 still.

alex:
I'll bump our minimum version them.

reaperhulk (approved):

PR #954, opened 2020-11-09
Retry on `WANT_WRITE` & `WANT_READ` in `sendall()`

### LINKED ISSUE #176: WANT_WRITE_ERROR occurs for sendall
The OpenSSL documentation says that in the event of a WANT_WRITE_ERROR or WANT_READ_ERROR, the same OpenSSL method call is to be repeated, otherwise you will get a bad write retry error.  

**See here:**
https://www.openssl.org/docs/ssl/SSL_write.html
http://stackoverflow.com/questions/2997218/why-am-i-getting-error1409f07fssl-routinesssl3-write-pending-bad-write-retr

For pyOpenSSL.sendAll() this is problematic.  Because python's sendAll() call does not return the number of bytes already sent if an error is thrown, it's impossible to continue with an identical SSL.Write() call.  The state of the last SSL.Write() call is only known inside the pyOpenSSL.sendAll() function, and since this function has thrown an exception, this state is lost.

To illustrate the problem, see the following source for SSL.sendall() in pyOpenSSL:
```python
    left_to_send = len(buf)
    total_sent = 0
    data = _ffi.new("char[]", buf)
  
    while left_to_send:
        result = _lib.SSL_write(self._ssl, data + total_sent, left_to_send)
        self._raise_ssl_error(self._ssl, result)
        total_sent += result
        left_to_send -= result
```

Specifically, inside the while loop, there are repeated calls to _lib.SSL_write().  _lib is the openSSL library, so the correct contract is that if SSL_write throws a WantWriteError or WantReadError, it should be handled by sleeping for a short duration and re-issuing the same call.  However, pyOpenSSL does not do this, and the caller cannot either.

As a workaround, I implemented a wrapper around pyOpenSSL, and implemented my own sendAll() to handle any WANT_WRITE_ERROR or WANT_READ_ERRORS in a loop, re-issuing the same send() call that the error threw (after a small sleep).

Finally, keep in mind that any write or read calls that are made to pyOpenSSL can fail with either a WANT_WRITE_ERROR or a WANT_READ_ERROR, because these calls *could* trigger an SSL handshake.  If network buffers are full or nearly full when the handshake triggers, these errors can be thrown.  For this reason, both error conditions must be handled by all network calls, either in pyOpenSSL or by developer using pyOpenSSL.

**Recommended fix:**
pyOpenSSL should write exception handlers to catch WANT_WRITE_ERROR and WANT_READ_ERRORS inside all functions to catch those exceptions, sleep for a small period of time, and re-issue the same SSL_write() or SSL_read() call.  The specific implementation will need to be considerate of blocking vs non-blocking mode, as well as timeouts, and is likely not trivial.

Thank you for all the work you do on pyOpenSSL.  It really is a fantastic library.

webknjaz:
This change introduces retries in `OpenSSL.SSL.Connection.sendall()`
when `WANT_WRITE_ERROR` or `WANT_READ_ERROR` happen.

It relies on `SSL_MODE_ENABLE_PARTIAL_WRITE` being set on the context,
that changes the mode of `SSL_write()` to return errors only if zero
bytes has been sent making it safe to retry in these cases.

Ideally, the calling code is supposed to `poll()`/`select()` the
socket to know when it's okay to attempt the next retry (hence it is
readable or writable) but it's not available in the `sendall()` method
and just retrying the operation is good enough.

Fixes #176

Refs:
* http://openssl.6102.n7.nabble.com/SSL-MODE-ACCEPT-MOVING-WRITE-BUFFER-td6421.html
* https://stackoverflow.com/a/28992313/595220
* https://www.openssl.org/docs/manmaster/man3/SSL_write.html
* https://stackoverflow.com/a/20817394/595220


reaperhulk:
Is there a way we can cover this with a test? And is `SSL_MODE_ENABLE_PARTIAL_WRITE` being set anywhere?

webknjaz:
> And is `SSL_MODE_ENABLE_PARTIAL_WRITE` being set anywhere?

It's set in the context initializer: https://github.com/pyca/pyopenssl/blob/124a013/src/OpenSSL/SSL.py#L660. Also, the docstring of `send()` seems to imply the same: https://github.com/pyca/pyopenssl/blob/124a013/src/OpenSSL/SSL.py#L1623-L1625 — I did not understand why is that at first but now it makes sense.

> Is there a way we can cover this with a test?

I'm not sure how to properly trigger this in a stable manner at the moment. Although, the suspicion is that maybe if we could limit the buffer size of the test socket and feed a sufficiently big chunk of data into it, it should work.
https://github.com/pyca/pyopenssl/blob/124a0134fdb7decb0136b4b6f7892b87b919e74e/tests/test_ssl.py#L2659-L2690 seems to be doing this exactly.

webknjaz:
I'm trying to come up with with a reproducer @ https://github.com/pyca/pyopenssl/pull/955 but it still needs polishing.

CHANGELOG.rst
```diff
@@ -25,6 +25,10 @@ Deprecations:
 Changes:
 ^^^^^^^^
 
+- Fixed the inability of ``OpenSSL.SSL.Connection.sendall()`` to
+  keep with sending data over the wire after ``SSL_ERROR_WANT_READ``
```
webknjaz, line 29:
```suggestion
  keep sending data over the wire after ``SSL_ERROR_WANT_READ``
```

PR #955, opened 2020-11-09
Add a test for `WANT_READ` during `sendall()`

webknjaz:
This is my attempt to trigger `WANT_READ` during `sendall()` but it is incomplete at the moment.
Any hints on how to make the reproducer more stable are welcome.


webknjaz:
@reaperhulk combining this with #954 causes `ssl3_read_bytes` and I do not understand why. Any idea what I'm doing wrong?

webknjaz:
@julianz- any idea how to write a reliable test for this?

PR #956, opened 2020-11-10
Keep reference to SSL verify_call in Connection object

schwabe:
If a set_verify is used on a context before and after a Connection
the reference in the SSL* object still points to the old _verify_helper
object. Since this object has no longer any references to it, the
callback can result in a segfault.

This commit fixes the issues by ensuring that as long as the
Connection object/SSL* object lives a reference to the callback
function is held.

Note: I have not checked if other callback functions suffer the same problem

### DIFF
--- a/src/OpenSSL/SSL.py
+++ b/src/OpenSSL/SSL.py
@@ -1485,6 +1485,12 @@ class Connection(object):
         # avoid them getting freed.
         self._alpn_select_callback_args = None
 
+        # Reference the verify_callback of the Context. This ensures that if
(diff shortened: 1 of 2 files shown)


alex:
Is it possible to write a test case for this?

schwabe:
I added a test case. The test case reliable crashes for me on Ubuntu 20.04 and macOS 10.15.7. I ordered the test after the fix so the segfault does not break git bisect etc.

reaperhulk (approved):

PR #963, opened 2020-11-26
Added set_ciphersuites() API

schlenk:
Add the set_ciphersuites API to set TLS 1.3 ciphersuites properly.

See:
https://www.openssl.org/docs/manmaster/man3/SSL_CTX_set_ciphersuites.html
https://wiki.openssl.org/index.php/TLS1.3#Ciphersuites

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -24,7 +24,9 @@ Deprecations:
 
 Changes:
 ^^^^^^^^
-
+- Added the ``OpenSSL.SSL.Context.set_ciphersuites`` method to
+  allow setting the TLS 1.3 ciphersuites.
+  `#XXX <https://github.com/pyca/pyopenssl/pull/999`_
 - Added a new optional ``chain`` parameter to ``OpenSSL.crypto.X509StoreContext()``
   where additional untrusted certificates can be specified to help chain building.
   `#948 <https://github.com/pyca/pyopenssl/pull/948>`_
--- a/src/OpenSSL/SSL.py
+++ b/src/OpenSSL/SSL.py
@@ -1129,6 +1129,25 @@ class Context(object):
                 ],
             )
 
+    def set_ciphersuites(self, ciphersuites):
+        """
+        Set the list of TLS 1.3 ciphersuites to be used in this context.
+
+        See the OpenSSL manual for more information (e.g.
+        :manpage:`ciphers(1)`).
+
+        :param bytes ciphersuites: An OpenSSL ciphersuites string.
+        :return: None
+        """
+        ciphersuites = _text_to_bytes_and_warn("ciphersuites", ciphersuites)
+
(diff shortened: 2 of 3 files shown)


alex:
What's the motivation? All the TLS 1.3 ciphersuites are secure (and good) which removes a lot of the motivation that exists in TLS<=1.2 for micromanaging them.

schlenk:
Motivation is three part:
a) Cannot configure the CCM variants
b) Looked at the Twisted API for cipher suites and it is broken due to not understanding the existing API (https://twistedmatrix.com/trac/ticket/10058)
c) Fixing some in-house enterprise stuff where people want to configure that stuff...

tiran:
IMHO it's bad practice to configure any TLS and crypto-related settings in an application. Cipher suites, key strength, TLS versions, and trust store should be configured globally on operating system level. In the past few years Linux distributions have invested into better infrastructure for system-wide crypto policies.

> What's the motivation? All the TLS 1.3 ciphersuites are secure (and good) which removes a lot of the motivation that exists in TLS<=1.2 for micromanaging them.

I have used the same argument to object against ``SSL_CTX_set_ciphersuites`` for Python's ssl module.

alex:
> IMHO it's bad practice to configure any TLS and crypto-related settings in an application. Cipher suites, key strength, TLS versions, and trust store should be configured globally on operating system level. In the past few years Linux distributions have invested into better infrastructure for system-wide crypto policies.

I don't agree with this at all. This is a very sysadmin/distro centric view of the world, and I think it maps quite poorly to lots of real world applications.

Nevertherless, I'm loathe to expand the pyOpenSSL API here.

schlenk:
Hi @tiran 

yes, agreed that i should be done on the OS level usually and that TLS 1.3 ciphers are great and all that.
I basically ship a whole python distro/openssl with custom options/apache httpd etc. as part of some framework. So probably should tweak the "framework global" openssl settings instead. I can fork/patch my local versions as needed.

schwabe:
There is also the use that might need to configure to a different default when the distribution changes the default setting. E.g. Centos8 does this:

```
cat ./crypto-policies/back-ends/opensslcnf.config

CipherString = @SECLEVEL=2:kEECDH:kRSA:kEDH:kPSK:kDHEPSK:kECDHEPSK:-aDSS:-3DES:!DES:!RC4:!RC2:!IDEA:-SEED:!eNULL:!aNULL:!MD5:-SHA384:-CAMELLIA:-ARIA:-AESCCM8
Ciphersuites = TLS_AES_256_GCM_SHA384:TLS_CHACHA20_POLY1305_SHA256:TLS_AES_128_GCM_SHA256:TLS_AES_128_CCM_SHA256
MinProtocol = TLSv1.2
```

So is probably more a feature parity between TLS <=1.2 and TLS 1.3+ than anything else.

PR #966, opened 2020-11-27
Tox.ini; Test on Python 3.9 and make flake8 stricter

cclauss:


alex:
If we want to do this, there's some flake8 fixes that need to happen.

reaperhulk:
We have a line-length of 79 set on black (see pyproject.toml), so we don't want 88 on flake8.

reaperhulk (approved):

PR #982, opened 2020-12-16
remove CB_ISSUER_CHECK

neheb:
Deprecated under OpenSSL 1.1

Signed-off-by: Rosen Penev <rosenp@gmail.com>


reaperhulk:
Deprecated or removed? Is this part of the goal of removing this value from cryptography itself?

neheb:
OpenSSL 1.1 still defines this under normal circumstances (OPENSSL_NO_DEPRECATED disabled)

The goal is to remove https://github.com/pyca/cryptography/pull/4920/commits/dc04a560e446387e4de0b7be841091cc562c2d60 from my PR. Which means remove it everywhere it's used.

reaperhulk (approved):
Since this has no effect on OpenSSL versions we support let's do it.

PR #985, opened 2020-12-21
Add SSL_CTX_set_min_proto_version/SSL_CTX_set_max_proto_version bindings

### LINKED ISSUE #860: add SSL_CTX_set_min_proto_version and SSL_CTX_set_max_proto_version
I'd like to be able to create TLS connections using only TLS 1.3, and I'd like to be able to create TLS connections using only TLS 1.2 (or other older TLS versions).

In the C API provided by openssl 1.1.0 and later, there is no TLSv1_3_METHOD at all, and the SSL_OP_NO_TLSvN scheme is deprecated in favor of specifying min/max protocol versions, using the new APIs SSL_CTX_set_min_proto_version() and SSL_CTX_set_max_proto_version().

Can you provide wrappers for the SSL_CTX_set_min_proto_version() and SSL_CTX_set_max_proto_version() library functions, and import the family of constants such as TLS1_3_VERSION and TLS1_2_VERSION for use with the new min/max API?

For context on the upstream OpenSSL API changes, you might refer to https://github.com/openssl/openssl/pull/1804#issuecomment-257230640 and https://github.com/openssl/openssl/issues/6964

mhils:
This PR adds the SSL_CTX_set_min_proto_version/SSL_CTX_set_max_proto_version protocol selection mechanisms along with TLS_method/TLS_client_method/TLS_server_method to pyOpenSSL.

depends on https://github.com/pyca/cryptography/pull/5662, fixes #860

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,7 +11,6 @@ jobs:
         PYTHON:
           # Base builds
           - {VERSION: "2.7", TOXENV: "py27"}
-          - {VERSION: "3.5", TOXENV: "py35"}
           - {VERSION: "3.6", TOXENV: "py36"}
           - {VERSION: "3.7", TOXENV: "py37"}
           - {VERSION: "3.8", TOXENV: "py38"}
@@ -26,7 +25,6 @@ jobs:
           - {VERSION: "pypy3", TOXENV: "pypy3-cryptographyMaster"}
           # -cryptographyMinimum
           - {VERSION: "2.7", TOXENV: "py27-cryptographyMinimum"}
-          - {VERSION: "3.5", TOXENV: "py35-cryptographyMinimum"}
           - {VERSION: "3.6", TOXENV: "py36-cryptographyMinimum"}
           - {VERSION: "3.7", TOXENV: "py37-cryptographyMinimum"}
           - {VERSION: "3.8", TOXENV: "py38-cryptographyMinimum"}
@@ -66,7 +64,7 @@ jobs:
       matrix:
         TEST:
           - {CONTAINER: "stretch", TOXENV: "py27"}
-          - {CONTAINER: "stretch", TOXENV: "py35"}
+          - {CONTAINER: "ubuntu-bionic", TOXENV: "py36"}
     name: "${{ matrix.TEST.TOXENV }} on ${{ matrix.TEST.CONTAINER }}"
     steps:
       - uses: actions/checkout@v2
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -71,6 +71,7 @@ jobs:
       - run: tox -v
         env:
           TOXENV: ${{ matrix.TEST.TOXENV }}
+          RUSTUP_HOME: /root/.rustup
       - name: Upload coverage
         run: |
           curl -o codecov.sh -f https://codecov.io/bash || curl -o codecov.sh -f https://codecov.io/bash || curl -o codecov.sh -f https://codecov.io/bash
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -18,6 +18,8 @@ Changes:
 
 - Raise an error when an invalid ALPN value is set.
   `#993 <https://github.com/pyca/pyopenssl/pull/993>`_
+- Added ``OpenSSL.SSL.Context.set_min_proto_version`` and ``OpenSSL.SSL.Context.set_max_proto_version``
+  to set the minimum and maximum supported TLS version `#985 <https://github.com/pyca/pyopenssl/pull/985>`_.
 
 20.0.1 (2020-12-15)
 -------------------
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -10,6 +10,9 @@ The third digit is only for regressions.
 Backward-incompatible changes:
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
+- The minimum ``cryptography`` version is now 3.3.
+- Drop support for Python 3.5
+
 Deprecations:
 ^^^^^^^^^^^^^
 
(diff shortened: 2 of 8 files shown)


mhils:
This should generally be ready for review, but depends on a new cryptography release (and then a bump of the minimum version here). 😃

mhils:
Tests for 3.6 and above are passing now. 😃 

2.7 and 3.5 are failing because they are not supported by the latest cryptography release anymore. I presume you want to drop them for pyOpenSSL as well soon-ish? :)

reaperhulk:
We'll try to keep 2.7 support in pyOpenSSL until the minimum version has to be bumped to support new bindings. These bindings were recently added, but were they in the 3.3 release?

reaperhulk:
I could have scrolled up in this PR. I guess we didn't, which means this requires 3.4? 🤣

mhils:
- SSL_CTX_set_min_proto_version/SSL_CTX_set_max_proto_version landed in https://github.com/pyca/cryptography/pull/5595 on December 1st.
- 3.3 was released on December 8th.
- The version constants landed in https://github.com/pyca/cryptography/pull/5662 on December 21st.

One could ship a 3.3 release with hardcoded constants if it needs to be, but bumping to 3.4 would be cleaner.

Also - having just taken a quick look at the cryptography tracker - thank you all for the fantastic work! ❤️ All of my projects happily accepted 3.4, everything worked out of the box, and I can't wait to see more memory-safe code being used. :)

reaperhulk:
Given pyOpenSSL's common use in more legacy environments I think we probably should just hardcode the constants (with comments that state when we move to 3.4+ of cryptography we can switch to getting them from cryptography) and allow the use of 3.3 still. I'd like to keep 2.7 support here another year if possible.

And thanks for the kind words -- it's going to be a rough couple weeks 😄

reaperhulk:
We dropped py3.5 in 3.3 so you'll need to update the CI config to remove the py35 configs. Sorry about that!

mhils:
No worries, I was just about to ask if you want me to do it. :)

.github/workflows/ci.yml
```diff
@@ -66,7 +64,7 @@ jobs:
       matrix:
         TEST:
           - {CONTAINER: "stretch", TOXENV: "py27"}
-          - {CONTAINER: "stretch", TOXENV: "py35"}
+          - {CONTAINER: "ubuntu-bionic", TOXENV: "py36"}
```
mhils, line 67: I tried changing the distro test to ubuntu-bionic (stretch only has 3.5), which fails due to the old pip version + cryptography 3.4. Do you want to update pip here, or specify RUSTUP_HOME as it is done in the cryptography tests [here](https://github.com/pyca/cryptography/blob/5fdc9c1b8563c2ffe99d6dff3f0632fb9dc29936/.github/workflows/ci.yml#L134)?

reaperhulk: Let's upgrade pip here. No need to compile cryptography if we don't have to...

reaperhulk:
Also needs a changelog rebase

.github/workflows/ci.yml
```diff
@@ -66,10 +64,12 @@ jobs:
...
     name: "${{ matrix.TEST.TOXENV }} on ${{ matrix.TEST.CONTAINER }}"
     steps:
       - uses: actions/checkout@v2
+      - run: python3 -m pip install -U pip
+        if: ${{ matrix.TEST.CONTAINER == "ubuntu-bionic" }}
```
reaperhulk, line 72: This is invalid syntax in an `if`. You can remove the `${{ }}` wrapper.

mhils: That explains a lot, thanks 😄 🙈

mhils:
Turns out the culprit was that constants require single quotes, not double quotes. 🤷‍♂️ 

Any hints on the py27 job failing with "ERROR: InterpreterNotFound: python2.7"?

reaperhulk:
Looks like https://github.com/pyca/infra/blob/main/runners/debian/Dockerfile#L20 isn't doing its job. Looking inside the container that's because `/etc/debian_version` no longer contains a name and instead just has a version number. I wonder how long this has been true...

reaperhulk:
https://github.com/pyca/infra/pull/340

mhils:
Ok, removing `PIP_NO_BINARY=cryptography` in tox.ini fixes the ubuntu-bionic tests:
https://github.com/pyca/pyopenssl/blob/d290855aab9f12d7cf739c63aad9ca3699d936f7/tox.ini#L19

`PIP_NO_BINARY=:all:` was introduced by @hynek in https://github.com/pyca/pyopenssl/commit/8b5ee6f6546915cee150788bfc92c3be5fcee990 ("wheels make testing of various OpenSSLs on OS X impossible") and then changed to only apply to cryptography in https://github.com/pyca/pyopenssl/commit/0c21cbd3a64f59f04a670c4ed1220f3d7a320398 ("Banning all breaks setuptools installation"). 

Is there a particular (security) reason why cryptography is not installed as a wheel?

If yes, what approach shall we take for the ubuntu-bionic container build? Without wheels, [it errors building cryptography](https://github.com/pyca/pyopenssl/runs/2052281536?check_suite_focus=true#step:5:212).
If no, this PR should be good to go.

reaperhulk:
I believe this was originally done so we could test alternate versions of OpenSSL against pyOpenSSL rather than just the ones our wheels ship. That seems like it's probably still worthwhile since we *don't* test that all of pyOpenSSL works on the set of OpenSSL versions cryptography tests against in its own matrix. So we should probably just make it so we can build the rust components in the bionic container.

mhils:
Alright, this was in fact very straightforward in the end! 😃

reaperhulk (approved):

reaperhulk:
Thanks for working through this @mhils

mhils:
Always welcome, thanks y'all! 😊

PR #993, opened 2021-02-17
Check return code of SSL_[CTX_]set_alpn_protos

### LINKED ISSUE #992: check return code of `SSL_CTX_set_alpn_protos`?
as found by @mhils and me in https://github.com/mitmproxy/mitmproxy/issues/4451#issuecomment-780270030 ff,
pyopenssl does not check the return code of `SSL_CTX_set_alpn_protos()`,
https://github.com/pyca/pyopenssl/blob/d182dab3f7b51c3f3d5c13687f1abba81acd78e4/src/OpenSSL/SSL.py#L137882
while that function fails on input as simple as an empty list.

mhils:
This PR fixes #992. 😃

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -16,6 +16,9 @@ Deprecations:
 Changes:
 ^^^^^^^^
 
+- Raise an error when an invalid ALPN value is set.
+  `#993 <https://github.com/pyca/pyopenssl/pull/993>`_
+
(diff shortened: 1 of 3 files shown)


reaperhulk:
Looks like this needs to be blackened.

mhils:
Fixed! :)

reaperhulk (approved):

PR #1027, opened 2021-07-08
Accept pathlib.Path as a valid path

ltworf:
And also whatever can be converted to bytes.

Way more pythonic now!

### DIFF
--- a/src/OpenSSL/_util.py
+++ b/src/OpenSSL/_util.py
@@ -1,3 +1,4 @@
+import os
 import sys
 import warnings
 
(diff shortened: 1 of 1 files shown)


ltworf:
Hello?

Plz?

alex:
I think this would be more appropriately handled with `os.fspath`.

ltworf:
Done

src/OpenSSL/_util.py
```diff
@@ -96,12 +97,17 @@ def path_string(s):
 
     :return: An instance of :py:class:`bytes`.
     """
-    if isinstance(s, bytes):
-        return s
-    elif isinstance(s, str):
-        return s.encode(sys.getfilesystemencoding())
+    try:
+        strpath = os.fspath(s)
+    except Exception:
```
alex, line 102: Why do we need to catch exceptions here? What case is that handling?

ltworf:
```
In [2]: os.fspath(1)
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-2-b00c080840d9> in <module>
----> 1 os.fspath(1)

TypeError: expected str, bytes or os.PathLike object, not int
```

ltworf: Just because you seem to want to provide your own message. I'd be fine to not handle it and remove the else case later on as well.

alex: Yes, I think allowing the exception from `os.fspath` to propagate is fine.

alex (approved):
Thanks!

Issue #1032, opened 2021-08-01
merge types-pyopenssl into pyopenssl


graingert:
It would be useful to merge the types-pyopenssl into pyopenssl.

It could start by just pasting the .pyi files into the repo and running check-stubs, but could also be iteratively improved to use inline type annotations and mypy type checking

See also https://github.com/twisted/twisted/pull/1484#issuecomment-890476456

adiroiban:
Thanks.
I think that an initial .pyi import would help...as is better to have them available by default with the openssl package.

graingert:
And running the stubcheck here found an annotation bug!

alex:
Now that we've dropped py2, we'd rather just integrate types directly into the code, not import the stubs.

lovetox:
Hi, i don’t think merging the pyi files is worth it, the project code base is really small.

Would a PR be appreciated? I could work on that.

reaperhulk:
We'd be happy to take PRs for inline typing. I'd suggest submitting it as incremental PRs: first adding a CI job with mypy + config and then a series of PRs adding types (and updating the config to remove exclusions, etc) until it's complete.

PR #1033, opened 2021-08-02
import stubs from typeshed

graingert:
fixes https://github.com/pyca/pyopenssl/issues/1032


reaperhulk (commented):
I would prefer that we just run the mypy tests in the flake tox job and not add all these new environments.

graingert:
> I would prefer that we just run the mypy tests in the flake tox job and not add all these new environments.

You need to run mypy once for each Python version and platform

reaperhulk:
Why? We don’t do this on cryptography. Are the types so fragile that I need to be concerned about them breaking on different versions 3.6+?

graingert:
> Why? We don’t do this on cryptography. Are the types so fragile that I need to be concerned about them breaking on different versions 3.6+?

Yeah I already had type errors on CI that passed locally because I only ran on one interpreter

penguinolog:
> Are the types so fragile that I need to be concerned about them breaking on different versions 3.6+?

On python 3.6 - yes. Later `from __future__ import annotations` solve most problems

graingert:
There's two different problems,
The first is that the typing syntax is different on different versions of Python you use `__future__ annotations` to use the new syntax.

The second is that different Python versions have different types and functions you run mypy on each Python version to solve this

lovetox:
This can should be obsolete once #1089 is merged

mhils:
Superseded by #1089 with inlined annotations. If someone would be willing to port over the annotations for `SSL.py` that would be most wonderful. 😃

PR #1039, opened 2021-09-07
Fill in notBefore/notAfter in X509 _PKeyInteractionTestsMixin tests

davidben:
While the tests currently pass without it, this is because OpenSSL's encoder doesn't notice that it is emitting garbage. See https://github.com/openssl/openssl/issues/16538

Fill in a placeholder validity period so the tests both better mirror real X.509 signing code and do not rely on this bug.


alex (approved):
I have a keen dislike for openssl x509

PR #1045, opened 2021-09-30
Misc CI cleanups

njsmith:
- Update branch names to match upstream
- Drop py2 testing
- Appease stricter checks in latest Sphinx version

Split off from #1026 as requested.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -1,5 +1,5 @@
 [tox]
-envlist = {pypy3,py36,py37,py38,py39}{,-cryptographyMain,-cryptographyMinimum}{,-randomorder},py37-twistedTrunk,pypi-readme,check-manifest,flake8,docs,coverage-report
+envlist = {pypy,pypy3,py27,py36,py37,py38,py39}{,-cryptographyMain,-cryptographyMinimum}{,-randomorder},py37-twistedTrunk,pypi-readme,check-manifest,flake8,docs,coverage-report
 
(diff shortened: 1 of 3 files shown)


alex:
Looks like CI is failing. Is this related to this PR or was it pre-broken?

njsmith:
Oh it was pre-broken, but I think I just pushed the piece to unbreak it while you were commenting

tox.ini
```diff
@@ -9,8 +9,7 @@ extras =
     test
 deps =
     coverage>=4.2
-    cryptographyMaster: git+https://github.com/pyca/cryptography.git
-    cryptographyMinimum: cryptography==3.3
```
alex, line 13: Without this line, how does -minimum work?

njsmith: Huh, when I told magit to revert that line I wasn't expecting it to delete it entirely. A bit disturbing that nothing breaks with that missing. Anyway, nice catch, thanks :-) And I reverted it properly now...

reaperhulk:
As long as we support that cryptography 3.3 minimum we should keep testing python 2.7, otherwise lgtm.

njsmith:
Well, I don't really care about support for py2 or old cryptographys, but I'm happy to go with whatever.

Does anyone care, or should we just toss them out and move on?

reaperhulk (approved):

PR #1047, opened 2021-10-04
py27 going, going, gone

reaperhulk:
let CI be my guide


njsmith (approved):
LGTM. Codecov seems to be claiming that there's a line somewhere that's no longer being covered, which might mean there's another line that can be removed? I failed at deciphering its output though. Do you care?

njsmith:
Meh, having stared at the codecov report some more I'm pretty sure the complaint about new uncovered lines is just some kind of artifact.

Issue #1046, opened 2021-10-04
OpenSSL.SSL.Context missing from Intersphinx mapping in 21.0.0


pquentin:
# Steps to reproduce

Inspect published Sphinx mapping:

```bash
$ curl https://www.pyopenssl.org/en/21.0.0/objects.inv > objects.inv && python -m sphinx.ext.intersphinx objects.inv | grep Context
        OpenSSL.SSL.ContextType                  api/ssl.html#OpenSSL.SSL.ContextType
        openssl-context                          Context objects                         : api/ssl.html#openssl-context
        openssl-x509storecontext                 X509StoreContext objects                : api/crypto.html#openssl-x509storecontext
        openssl-x509storecontexterror            X509StoreContextError objects           : api/crypto.html#openssl-x509storecontexterror
```

# Expected output

Just like 20.0.1:

```bash
$ curl https://www.pyopenssl.org/en/20.0.1/objects.inv > objects.inv && python -m sphinx.ext.intersphinx objects.inv | grep Context   
 https://www.pyopenssl.org/en/stable
 https://www.pyopenssl.org/en/stable
	OpenSSL.SSL.Context                      api/ssl.html#OpenSSL.SSL.Context
	OpenSSL.crypto.X509StoreContext          api/crypto.html#OpenSSL.crypto.X509StoreContext
	OpenSSL.crypto.X509StoreContextError     api/crypto.html#OpenSSL.crypto.X509StoreContextError
	OpenSSL.SSL.ContextType                  api/ssl.html#OpenSSL.SSL.ContextType
	OpenSSL.SSL.Context.add_client_ca        api/ssl.html#OpenSSL.SSL.Context.add_client_ca
    ...
	OpenSSL.SSL.Context.use_privatekey_file  api/ssl.html#OpenSSL.SSL.Context.use_privatekey_file
	OpenSSL.crypto.X509StoreContext.get_verified_chain api/crypto.html#OpenSSL.crypto.X509StoreContext.get_verified_chain
	OpenSSL.crypto.X509StoreContext.set_store api/crypto.html#OpenSSL.crypto.X509StoreContext.set_store
	OpenSSL.crypto.X509StoreContext.verify_certificate api/crypto.html#OpenSSL.crypto.X509StoreContext.verify_certificate
	openssl-context                          Context objects                         : api/ssl.html#openssl-context
	openssl-x509storecontext                 X509StoreContext objects                : api/crypto.html#openssl-x509storecontext
[...]

PR #1048, opened 2021-10-04
Testing .readthedocs.yml changes

njsmith:


njsmith:
OK, looks like this change in fact worked.

Index before: https://www.pyopenssl.org/en/latest/genindex.html

Index after: https://pyopenssl--1048.org.readthedocs.build/en/1048/genindex.html

So this fixes #1046

reaperhulk (approved):

PR #1078, opened 2022-01-25
Docs: Fix param type for methods where digest is passed

lovetox:


lovetox:
This would also need fixing in the python typeshed, i would do this if this gets merged

reaperhulk (approved):
Thanks for fixing the documentation here!

PR #1083, opened 2022-01-30
Handle no expire date in X509.has_expire()

lovetox:
This is my naive suggestion to handle this case, obviously the return value of this method is very critical.

Alternative would be to raise an exception with a nice error text.


alex:
Can you provide a test case for this? I'm not sure I understand when this would happen.

lovetox:
I sadly can’t. I don’t have experience with certificates and don’t even know if such a cert would be legal.

But anyway `get_notAfter()` can return `Ǹone`. If we assume that will be never the case we should raise a Exception or add an `assert is not None` statement.

Both will tell typecheckers about this assumption and correctly infer the type in the code which follows.

The fact that this codepath was there and the NoneType Exception not reported, tells me it is either not very likely or not possible. So i changed the PR to reflect that.

It now explicitly raises an Exception instead of just failing with NoneType Error.

reaperhulk:
`get_notAfter` can return `None` when you simply construct a raw `X509()` and don't call any setter methods. This is pretty unfortunate for typing, but is definitely a thing that can happen. You can trigger an error in `    time_string = self.get_notAfter().decode("utf-8")` by simply doing `x = X509(); x.has_expired()`

lovetox:
@reaperhulk 
I will use your example to write a test, but there is still no answer to the question how you want to handle if it returns `None`.

Until now it throws a decoding exception, in my MR i changed this to a explicit exception. But we could also return `False`. Because without date there is no expire?

alex (approved):

PR #1086, opened 2022-01-30
X509Name: Use functools.totalordering for comparisons

lovetox:
- Reduce the magic
- Make it more readable
- Make it easier to add type annotations in the future

### DIFF
--- a/src/OpenSSL/crypto.py
+++ b/src/OpenSSL/crypto.py
@@ -1,9 +1,8 @@
 import calendar
 import datetime
-
 from base64 import b16encode
+import functools
 from functools import partial
-from operator import __eq__, __ne__, __lt__, __le__, __gt__, __ge__
 
 from cryptography import utils, x509
 from cryptography.hazmat.primitives.asymmetric import dsa, rsa
@@ -528,6 +527,7 @@ def get_elliptic_curve(name):
     raise ValueError("unknown curve name", name)
 
 
+@functools.total_ordering
 class X509Name:
     """
     An X.509 Distinguished Name.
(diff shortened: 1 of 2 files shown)


src/OpenSSL/crypto.py
```diff
@@ -642,23 +642,17 @@ def __getattr__(self, name):
...
+            return NotImplemented
 
-    __eq__ = _cmp(__eq__)
-    __ne__ = _cmp(__ne__)
+        return _lib.X509_NAME_cmp(self._name, other._name)
 
-    __lt__ = _cmp(__lt__)
-    __le__ = _cmp(__le__)
+    def __eq__(self, other):
+        return self._cmp(other) == 0
```
alex, line 652: This isn't quite right -- in the event `other` is not an `X509Name` then this is going to do `NotImplemented == 0` which isn't what we want.

lovetox: whoops, that should of course be raising

alex: To be clear: We want to return `NotImplemented` from `__eq__` if the other side isn't an `X509Name` since that ensures that the RHS's `__eq__` can be called.

lovetox: Yeah i just found that out when looking up the difference between NotImplemented and NotImplementedError

I will think about how to solve this

lovetox: I wanted to avoid the code duplication, but i think its not worth it for the 2 isinstance calls.

alex (approved):

alex:
Oop, looks like there's merge conflicts unfortunately.

lovetox:
I think i fixed the conflicts

alex (commented):
Ooof sorry, one last request, can you add a test case that covers the `NotImplemented` branch in `__lt__`.

Thanks!

alex (approved):

PR #1089, opened 2022-01-30
Add inline type annotations

### LINKED ISSUE #1032: merge types-pyopenssl into pyopenssl

lovetox:
- Prepare Package and CI
- Everything is typed except SSL.py
- Extensive use of Any type
- mypy yields no errors with this PR on strict mode
- On one method `# type: ignore` had to be used
- The only logic change which had to be done is factored out into its own commit

I would like to request a review for this PR, its a good starting point on the way to having the whole codebase type checked.

The package is not yet marked as `typed` because SSL.py is missing annotations which contains public API.

Fixes #1032

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -42,6 +42,7 @@ jobs:
           # Meta
           - {VERSION: "3.9", TOXENV: "check-manifest"}
           - {VERSION: "3.9", TOXENV: "flake8"}
+          - {VERSION: "3.6", TOXENV: "py36-mypy"}
           - {VERSION: "3.9", TOXENV: "docs"}
     name: "${{ matrix.PYTHON.TOXENV }}"
     steps:
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,5 +1,5 @@
 include             LICENSE MANIFEST.in *.rst tox.ini .coveragerc
-exclude             codecov.yml .readthedocs.yml
+exclude             codecov.yml .readthedocs.yml mypy.ini
 recursive-include   tests           *.py
 recursive-include   doc             *
 prune               doc/_build
--- a/mypy.ini
+++ b/mypy.ini
@@ -1,0 +1,24 @@
+[mypy]
+warn_unused_configs = True
+follow_imports = skip
+strict = True
+
+exclude = (?x)(
+    SSL\.py$
+  )
+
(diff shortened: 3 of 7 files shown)


lovetox:
Ok seems i have no clue about tox, maybe someone can help out with getting mypy running in the CI

lovetox:
Ci runs now mypy but fails because of dependency on #1083

lovetox:
Hi,

CI runs without error now, MR is ready for Review.

mhils (commented):
Thanks for the PR! 😃 

Could you move mypy into a separate toxenv (similar to flake8)? AFAIK there's no benefit in running mypy on different Python versions (please correct me if I'm wrong). By moving it into a separate job we take it off the critical path.

The type hints so far look good to me. 👍

lovetox:
It seems im unable to add another env, i tried (see commits in tox.ini) but it does not add the env to CI, im new to tox though maybe i did something wrong

mhils:
You need to add a job in the workflow yaml in `/.github`. :)

lovetox:
Ok works now.

i opted to run mypy on the lowest supported version. This has the benefit of failing if anyone uses syntax or typing imports not supported by python 3.6. But one could argue this would be catched by other tests because this project has a high test coverage.

Tell me if you want to run it on latest version.

mhils:
Sounds sensible to me. Once the minimum Python version is >=3.7, it
probably makes a lot of sense to opt into `from __future__ import
annotations` and just use the latest version. :)

mhils (commented):
CI is currently not running what you think it is, see https://github.com/pyca/pyopenssl/runs/6436553928?check_suite_focus=true#step:5:53. I think that's because of the mismatch between `py-36-mypy` (tox.ini), `mypy` (tox.ini), and `py36-mypy` (ci.yml).

lovetox:
Should be ok now

mhils (approved):
Looks good to me now, thanks a bunch! 😃 We still need SSL.py to be annotated to reap the benefits of this, but thank you for getting this started! 🍰

PR #1109, opened 2022-03-13
Implement Context constructor in terms of new OpenSSL APIs

alex:


src/OpenSSL/SSL.py
```diff
@@ -680,31 +680,24 @@ class Context:
     """
 
     _methods = {
-        SSLv2_METHOD: "SSLv2_method",
```
reaperhulk, line 683: If we're going to remove this here (which I'm fine with), then we probably want to remove the top level constants too? The latter is an instantly breaking change for people who import it, while the current approach breaks at runtime.

I do wonder how many people will tell us "I use pyopenssl to do bad TLS deliberately"...?

alex: Done

reaperhulk (approved):

PR #1116, opened 2022-04-26
repair CI

reaperhulk:


src/OpenSSL/SSL.py
```diff
@@ -1685,6 +1685,19 @@ def _raise_ssl_error(self, ssl, result):
...
                 # TODO: This is untested.
                 _raise_current_error()
+        elif error == _lib.SSL_ERROR_SSL and _lib.ERR_peek_error() != 0:
+            # In 3.0.x an unexpected EOF no longer triggers syscall error
+            # but we want to maintain compatibility so we check here and
+            # raise syscall if it is an EOF.
+            peeked_error = _lib.ERR_peek_error()
+            reason = _lib.ERR_GET_REASON(peeked_error)
+            # 294 is SSL_R_UNEXPECTED_EOF_WHILE_READING in OpenSSL 3.0.x
+            if reason == 294:
```
alex, line 1695: Can we expose this constant in cryptography and backport for next week's 37.0.1? I hate this -- all it takes is one of the fork's doing this with a different version number and we've got a mess.

tests/test_crypto.py
```diff
@@ -4090,7 +4090,11 @@ def test_untrusted_self_signed(self):
         with pytest.raises(X509StoreContextError) as exc:
             store_ctx.verify_certificate()
 
-        assert exc.value.args[0][2] == "self signed certificate"
+        # OpenSSL 1.1.x and 3.0.x have different error messages
+        assert exc.value.args[0][2] in (
+            "self signed certificate",
+            "self-signed certificate",
+        )
```
alex, line 4097:
```suggestion
        assert exc.value.args[0][2] in [
            "self signed certificate",
            "self-signed certificate",
        ]
```

tests/test_ssl.py
```diff
@@ -517,14 +517,19 @@ def test_set_cipher_list_no_cipher_match(self, context):
...
+                "SSL_CTX_set_cipher_list",
+                "no cipher match",
+            ),
+            # 3.0.x
+            (
+                "SSL routines",
+                "",
+                "no cipher match",
+            ),
         )
```
alex, line 533:
```suggestion
        assert excinfo.value.args[0][0] in [
            # 1.1.x
            (
                "SSL routines",
                "SSL_CTX_set_cipher_list",
                "no cipher match",
            ),
            # 3.0.x
            (
                "SSL routines",
                "",
                "no cipher match",
            ),
        ]
```

tests/test_ssl.py
```diff
@@ -564,13 +569,20 @@ def test_set_session_id_fail(self, context):
         with pytest.raises(Error) as e:
             context.set_session_id(b"abc" * 1000)
 
-        assert [
+        assert e.value.args[0][0] in (
+            # 1.1.x
```
alex, line 573:
```suggestion
        assert e.value.args[0][0] in [
            # 1.1.x
```

tests/test_ssl.py
```diff
@@ -564,13 +569,20 @@ def test_set_session_id_fail(self, context):
...
-            )
-        ] == e.value.args[0]
+            ),
+            # 3.0.x
+            (
+                "SSL routines",
+                "",
+                "ssl session id context too long",
+            ),
+        )
```
alex, line 585:
```suggestion
        ]
```

src/OpenSSL/SSL.py
```diff
@@ -1685,6 +1685,20 @@ def _raise_ssl_error(self, ssl, result):
...
                 _raise_current_error()
+        elif error == _lib.SSL_ERROR_SSL and _lib.ERR_peek_error() != 0:
+            # In 3.0.x an unexpected EOF no longer triggers syscall error
+            # but we want to maintain compatibility so we check here and
+            # raise syscall if it is an EOF.
+            peeked_error = _lib.ERR_peek_error()
+            reason = _lib.ERR_GET_REASON(peeked_error)
+            try:
+                _openssl_assert(
+                    reason == _lib.SSL_R_UNEXPECTED_EOF_WHILE_READING
```
alex, line 1696: do we need to check `Cryptography_HAS_UNEXPECTED_EOF_WHILE_READING`

alex:
This should also bump the minimum pyca/cryptography version

reaperhulk:
The current code path will raise `AttributeError` and fall back if the constant is unavailable, which would mean it's invalid for any cryptography compiled against OpenSSL 3 that is not also version 37.0.2. In practice this is really just 37.0.0 and 37.0.1 wheels. Should we force absolute latest cryptography to close this hole? I'm not opposed, it just feels like we should explicitly decide. 😄 

We can remove the try/except if we move the requirement to 37.0.2 as well

alex:
I _strongly_ dislike catching AttributeError, as it makes it extremely easy to accidentally catch exceptions from the wrong place. I'm pro moving to cryptography>=37.0.2

This is a pretty basic dependency resolution thing and is unlikely to cause problems, thank god.

alex (approved):

Issue #1114, opened 2022-04-26
AttributeError: module 'lib' has no attribute 'X509_V_FLAG_CB_ISSUER_CHECK'


mvellasco:
Importing the `crypto` module from OpenSSL throws an AttributeError. This is happening under version 20.0.1 of the library.

```
File "/usr/local/lib/python3.8/site-packages/OpenSSL/__init__.py", line 8, in <module>
  from OpenSSL import crypto, SSL
File "/usr/local/lib/python3.8/site-packages/OpenSSL/crypto.py", line 1556, in <module>
     class X509StoreFlags(object):
File "/usr/local/lib/python3.8/site-packages/OpenSSL/crypto.py", line 1577, in X509StoreFlags
     CB_ISSUER_CHECK = _lib.X509_V_FLAG_CB_ISSUER_CHECK
AttributeError: module 'lib' has no attribute 'X509_V_FLAG_CB_ISSUER_CHECK'
```

alex:
Can you provide instructinos for reproducing this from a clean environment?

I strongly suspect you are mixing incompatible versions of pyOpenSSL and cryptography and one/both of them should be upgraded.

mvellasco:
Doing `pip install pyopenssl==20.0.1` alone and them going to the terminal and trying the import above gives the error for me. I'll check my version of cryptography as well to see if that's the issue.

alex:
You'll need to use a more recent pyopenssl version, or pin to an older cryptography version.

mvellasco:
Downgrading cryptography worked like a charm. Thank you very much @alex !

reaperhulk:
For those seeing this issue: You can upgrade `pyOpenSSL` to latest to resolve it or downgrade `cryptography`, but the latter is not a good long term solution since you will (of course) not receive security updates.

reaperhulk:
We re-added some symbols in 37.0.1 to minimize breakage. We still encourage users to upgrade to the latest pyOpenSSL as these will be removed again in a future release.

PR #1117, opened 2022-04-27
Fix error module 'lib' has no attribute 'X509_V_FLAG_CB_ISSUER_CHECK'

moylop260:
Release today:
 - https://pypi.org/project/cryptography/37.0.0

And it is raising the following error:
 - `module 'lib' has no attribute 'X509_V_FLAG_CB_ISSUER_CHECK'`

So current pyopenssl version is only compatible with cryptography<37.0.0

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -96,7 +96,7 @@ if __name__ == "__main__":
         package_dir={"": "src"},
         install_requires=[
             # Fix cryptographyMinimum in tox.ini when changing this!
-            "cryptography>=35.0",
+            "cryptography>=35.0,<37.0.0",
         ],
         extras_require={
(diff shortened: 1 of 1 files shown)


alex:
This is not correct, see https://github.com/pyca/pyopenssl/issues/1114.

The _current_ release of pyOpenSSL is very much compatible with the latest cryptography. You are using an old pyOpenSSL with a brand new cryptography. Upgrade your pyOpenSSL and everything will be great.

moylop260:
Using a limited version less-than is immutable the version installed

So, using a incompatible version it will raises error for all previous version of pyopenssl installing out-of-the-box for different reasons

But I got your point

Thank you for reply

PR #1121, opened 2022-05-11
add `Connection.use_(certificate|privatekey)`

mhils:
(depends on cryptography 38)

Follow-up to https://github.com/pyca/cryptography/pull/7210.  :)
I've added comments to point out the duplication between Context and Connection. Let me know if you want those out, happy to adjust to whatever style you prefer.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -25,6 +25,8 @@ Changes:
 - Add ``OpenSSL.SSL.Connection.set_verify`` and ``OpenSSL.SSL.Connection.get_verify_mode``
   to override the context object's verification flags.
   `#1073 <https://github.com/pyca/pyopenssl/pull/1073>`_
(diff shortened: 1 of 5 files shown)


reaperhulk:
cryptography 38 is out now 😄

reaperhulk:
Conflict here due to the new pinning, but we should get this reviewed ASAP 😄

reaperhulk (approved):

balloob:
This PR has not updated the changelog. Changelog for 22.1 still mentions Cryptography 37 as the minimum version.

reaperhulk:
@balloob Thanks for the report, we've got a changelog fix up in https://github.com/pyca/pyopenssl/pull/1150 although we may also need to spin some knobs on readthedocs as well since this won't fix the `stable` tag.

PR #1122, opened 2022-05-11
Fix CI

mhils:
readme_renderer 35.0 warns if there are duplicate implicit target names.
This breaks with pyOpenSSL's readme (which has duplicate subsections, e.g. "Deprecations:").

I'm happy to also add a whole bunch of anchors if you prefer that. I didn't feel that's necessarily better. :)

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -49,7 +49,7 @@ deps =
     readme_renderer
 skip_install = true
 commands =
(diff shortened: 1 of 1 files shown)


mhils:
Completely missed #1116, sorry 🙈

PR #1133, opened 2022-07-06
Make `X509StoreContextError`'s message friendlier

### LINKED ISSUE #1132: Make `X509StoreContextError` more structured?
First of all, thanks for maintaining pyOpenSSL!

This is how an `X509StoreContextError` is currently constructed: it's a `list` of errors, combined with the cert that caused the error. The last member of the error list is a string representation of `X509_STORE_CTX_get_error`:

https://github.com/pyca/pyopenssl/blob/02db1a024d04cf6669670f773fd6c5d3a7275626/src/OpenSSL/crypto.py#L1873-L1895

As a result, it's inconvenient to introspect the different ways in which certificate verification can fail. 

My proposal: the `errors` list stays the same, but the last member (the string representation) is also copied into a `message` attribute. In other words, something like this:

```py
    def _exception_from_context(self) -> X509StoreContextError:
        """
        Convert an OpenSSL native context error failure into a Python
        exception.
        When a call to native OpenSSL X509_verify_cert fails, additional
        information about the failure can be obtained from the store context.
        """
        message = _ffi.string(
            _lib.X509_verify_cert_error_string(
                _lib.X509_STORE_CTX_get_error(self._store_ctx)
            )
        ).decode("utf-8"),
        errors = [
            _lib.X509_STORE_CTX_get_error(self._store_ctx),
            _lib.X509_STORE_CTX_get_error_depth(self._store_ctx),
            message,
        ]
        # A context error should always be associated with a certificate, so we
        # expect this call to never return :class:`None`.
        _x509 = _lib.X509_STORE_CTX_get_current_cert(self._store_ctx)
        _cert = _lib.X509_dup(_x509)
        pycert = X509._from_raw_x509_ptr(_cert)
        return X509StoreContextError(message, errors, pycert)
```

Let me know what you think. I'm happy to submit a PR with these changes, if there's interest.

woodruffw:
The `message` is now a string, rather than a list of errors. The list of errors is separately preserved.

Closes #1132.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -12,6 +12,9 @@ Backward-incompatible changes:
 
 - Remove support for SSLv2 and SSLv3.
 - The minimum ``cryptography`` version is now 37.0.2.
+- The ``OpenSSL.crypto.X509StoreContextError`` exception has been refactored,
+  changing its internal attributes.
+  `#1133 <https://github.com/pyca/pyopenssl/pull/1133>`_
 
 Deprecations:
 ^^^^^^^^^^^^^
--- a/src/OpenSSL/crypto.py
+++ b/src/OpenSSL/crypto.py
@@ -1776,7 +1776,9 @@ class X509StoreContextError(Exception):
     :type certificate: :class:`X509`
     """
 
-    def __init__(self, message: str, errors: List[Any], certificate: X509) -> None:
+    def __init__(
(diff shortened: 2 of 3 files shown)


woodruffw:
Okay, updated the changelog. The exception class appears to be autodoc'd, so I don't think any doc changes are necessary. But let me know if I'm wrong.

PR #1136, opened 2022-07-18
Add py.typed file

daviddavis:
More info: https://peps.python.org/pep-0561/#packaging-type-information


daviddavis:
That makes sense. Thank you for the prompt response.

PR #1137, opened 2022-07-23
Fix docs in `SSL.Context.get_alpn_proto_negotiated`

sobolevn:
I am working on better types in https://github.com/python/typeshed/blob/master/stubs/pyOpenSSL/OpenSSL/SSL.pyi
I've noticed that your current docs are not quite correct, because it stated "returns an empty string", while the actual code did: `return b""`. It was a bit confusing.

### DIFF
--- a/src/OpenSSL/SSL.py
+++ b/src/OpenSSL/SSL.py
@@ -2648,7 +2648,7 @@ class Connection:
         Get the protocol that was negotiated by ALPN.
 
         :returns: A bytestring of the protocol name.  If no protocol has been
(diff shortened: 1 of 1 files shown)


mhils (approved):
Thanks! 😃 

FWIW, we'd be happy to get more type annotations into pyOpenSSL itself so that we eventually don't need typeshed anymore. If you want to contribute type annotations here, you are more than welcome!

sobolevn:
Thanks! I will finish typeshed first, then I can reapply them to the source (if their state is good enough).

PR #1139, opened 2022-08-12
Don't test that invalid RSA keys can be imported

davidben:
test_check_pr_897 asserts that an invalid key is correctly detected as invalid. However, in doing so, it also asserts that the invalid key is considered *valid* at parse time.

Ideally, the underlying cryptography library would just call RSA_check_key during parsing, but it would then fail this test. Make the test more tolerant by allow either parsing or checking to throw an error.

### DIFF
--- a/tests/test_crypto.py
+++ b/tests/test_crypto.py
@@ -1230,8 +1230,8 @@ class TestPKey:
         """
         `PKey.check` raises `OpenSSL.crypto.Error` if provided with broken key
         """
-        pkey = load_privatekey(FILETYPE_PEM, rsa_p_not_prime_pem)
         with pytest.raises(Error):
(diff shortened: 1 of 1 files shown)


tests/test_crypto.py
```diff
@@ -1230,8 +1230,8 @@ def test_check_pr_897(self):
         """
         `PKey.check` raises `OpenSSL.crypto.Error` if provided with broken key
```
alex, line 1231: Can you update this docstring to say that either `load_privatekey` or `check` raises

davidben: Done. Also I missed that we actually ran into this in two different places, so updated the other test too.

alex (approved):

PR #1140, opened 2022-08-13
Switch to the new utils.deprecation spelling

davidben:
The new spelling was introduced in https://github.com/pyca/cryptography/pull/6923 and is more friendly to type checkers.

Version-wise, that PR appears to be in cryptography 37.0.0, which is now beyond the minimum version for pyOpenSSL.


alex:
can you run `black` to get CI happy

davidben:
Done.

alex (approved):

PR #1163, opened 2022-12-02
Make some handshakes in tests less flaky.

tjni:
The test `test_set_keylog_callback` (and a couple others) are consistently failing for me locally on my M1 Mac. I don't know exactly why this happens, but I hypothesize it's related to the non-blocking nature of the handshake and perhaps delays in my system while the handshake is taking place. By adding a `select` call to block when `WantReadError` is raised, I'm able to get these tests to pass on my machine.

In addition, I noticed `pytest` warning me that the sockets created in the `_handshake_test` method were not closed, so I'm wrapping the body of the method now in a `with` block.

### DIFF
--- a/tests/test_ssl.py
+++ b/tests/test_ssl.py
@@ -7,6 +7,7 @@ Unit tests for :mod:`OpenSSL.SSL`.
 
 import datetime
 import gc
+import select
 import sys
 import uuid
 from errno import (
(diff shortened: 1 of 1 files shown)


mhils:
Thanks, @tjni! FYI I'm tackling the 3.6 problem here already: #1164

mhils (approved):

PR #1166, opened 2022-12-15
Add support for X509_V_FLAG_PARTIAL_CHAIN

vEpiphyte:
Add support for the X509_V_FLAG_PARTIAL_CHAIN constant. This is similar to the following cPython changes made to support partial certificate chain verification:

https://bugs.python.org/issue40849
https://github.com/python/cpython/pull/20463/files

While the cPython changes are in 3.10+, this would allow users with older python libraries to take advantage of this flag.


alex:
thanks for the PR.

test failures look to be related to tox4 -- https://github.com/pyca/pyopenssl/pull/1167 should fix those.

alex:
ok, if you can rebase/merge main we can test/review/merge

PR #1169, opened 2022-12-16
Fix CRL nextUpdate handling.

### LINKED ISSUE #1168: CRL.set_nextUpdate sometimes doesn't work and other setters violate const-correctness
Filing this as an issue since it'll be a couple of PRs. There are a few issues with how the time-setting functions in pyOpenSSL are implemented:

* `CRL.set_lastUpdate` mutates the result of `X509_CRL_get0_lastUpdate`. While this works, it's not const-correct
* `CRL.set_nextUpdate` does the same, but it also doesn't always work! nextUpdate may be NULL, in which case pyOpenSSL calls `ASN1_TIME_set_string` on NULL, which silently does a different operation.
* `Revoked.set_rev_date` also violates const

The `X509` are fine, but since the others need to use setters, rather than mutable getters, may as well make them all consistent at this point. The problem is cryptography doesn't bind `X509_REVOKED_set_revocationDate`. So I'm thinking what I'll do is:

* First PR, fix the `CRL` ones
* Second PR, add bindings to cryptography.io
* Third PR, align everything on setters, not to be merged yet
* @alex and @reaperhulk manage the cross-project coordination however they feel like and merge the third PR when ready.

davidben:
When setting the nextUpdate field of a CRL, this code grabbed the nextUpdate ASN1_TIME field from the CRL and set its time. But nextUpdate is optional in a CRL so that field is usually NULL. But OpenSSL's ASN1_TIME_set_string succeeds when the destination argument is NULL, so it was silently a no-op.

Given that, the call in a test to set the nextUpdate field suddenly starts working and sets the time to 2018, thus causing the CRL to be considered expired and breaking the test. So this change also changes the expiry year far into the future.

Additionally, the other CRL and Revoked setters violate const in the API.

Fixes #1168.

### DIFF
--- a/src/OpenSSL/crypto.py
+++ b/src/OpenSSL/crypto.py
@@ -168,10 +168,33 @@ def _set_asn1_time(boundary: Any, when: bytes) -> None:
     """
     if not isinstance(when, bytes):
         raise TypeError("when must be a byte string")
+    if boundary == _ffi.NULL:
+        # ASN1_TIME_set_string validates the string without writing anything
+        # when the destination is NULL.
+        raise ValueError("boundary must not be NULL")
 
     set_result = _lib.ASN1_TIME_set_string(boundary, when)
     if set_result == 0:
         raise ValueError("Invalid string")
+
+
+def _new_asn1_time(when: bytes) -> Any:
+    """
+    Behaves like _set_asn1_time bit returns a new ASN1_TIME object.
+
+    @param when: A string representation of the desired time value.
+
+    @raise TypeError: If C{when} is not a L{bytes} string.
+    @raise ValueError: If C{when} does not represent a time in the required
+        format.
+    @raise RuntimeError: If the time value cannot be set for some other
+        (unspecified) reason.
+    """
+    ret = _lib.ASN1_TIME_new()
+    _openssl_assert(ret != _ffi.NULL)
+    ret = _ffi.gc(ret, _lib.ASN1_TIME_free)
+    _set_asn1_time(ret, when)
+    return ret
 
 
 def _get_asn1_time(timestamp: Any) -> Optional[bytes]:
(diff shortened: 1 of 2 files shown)


src/OpenSSL/crypto.py
```diff
@@ -168,12 +168,35 @@ def _set_asn1_time(boundary: Any, when: bytes) -> None:
...
+    @raise TypeError: If C{when} is not a L{bytes} string.
+    @raise ValueError: If C{when} does not represent a time in the required
+        format.
+    @raise RuntimeError: If the time value cannot be set for some other
+        (unspecified) reason.
+    """
+    ret = _lib.ASN1_TIME_new()
+    _openssl_assert(ret != _ffi.NULL)
+    ret = _ffi.gc(ret, _lib.ASN1_TIME_free)
+    _set_asn1_time(ret, when)
```
alex, line 196: This doesn't appear to have any callers left, should it just be inlinined here? I think there's a bunch of checks in it that are now unreachable

alex: Ignore me, there's still 1 caller

alex (dismissed):

src/OpenSSL/crypto.py
```diff
@@ -168,12 +168,35 @@ def _set_asn1_time(boundary: Any, when: bytes) -> None:
     """
     if not isinstance(when, bytes):
         raise TypeError("when must be a byte string")
+    if boundary == _ffi.NULL:
+        # ASN1_TIME_set_string validates the string without writing anything
+        # when the destination is NULL.
+        raise ValueError("boundary must not be NULL")
```
alex, line 174: Is it possible to write a test case that reaches this?

davidben: Replaced with assert per out-of-band conversation. (This is unreachable. It's just to avoid another such bug being masked.)

src/OpenSSL/crypto.py
davidben: Yeah, see comment above. :-) We could get rid of it if we did a multi-project thing and bound some more APIs.

alex (dismissed):

src/OpenSSL/crypto.py
```diff
@@ -168,12 +168,34 @@ def _set_asn1_time(boundary: Any, when: bytes) -> None:
...
+    _openssl_assert(boundary != _ffi.NULL)
 
     set_result = _lib.ASN1_TIME_set_string(boundary, when)
     if set_result == 0:
         raise ValueError("Invalid string")
 
 
+def _new_asn1_time(when: bytes) -> Any:
+    """
+    Behaves like _set_asn1_time bit returns a new ASN1_TIME object.
```
alex, line 182:
```suggestion
    Behaves like _set_asn1_time but returns a new ASN1_TIME object.
```

PR #1193, opened 2023-03-21
Expose secure renegotiation flag from TLS connection

ndmalc:
Hello,

This is a pull request to add support for the SSL_get_secure_renegotiation_support function from OpenSSL.

Some simple test is added.

Best regards

### DIFF
--- a/src/OpenSSL/SSL.py
+++ b/src/OpenSSL/SSL.py
@@ -2730,6 +2730,23 @@ class Connection:
 
         return _ffi.buffer(data[0], data_len[0])[:]
 
+    def get_secure_renegotiation_support(self):
+        """
+        Retrieve the secure renegotiation flag of the current connection.
+
+        :returns: A boolean representing the support of secure renegotiation (rfc5746)
+            for the current connection. True means that secure renegotiation is advertised
+            and supported by server. False means that secure renegotiation is not supported
+            or that client renegotiation is not supported at all.
+        :rtype: :class:`bool`
+        """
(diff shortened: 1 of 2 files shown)


reaperhulk:
These bindings have actually been removed from `cryptography` and will not be present in the next release. The CI should have caught this, but there's a bug in the `cryptographyMain` job that causes it to re-install the latest release (rather than using `main`) due to the version bounds in pyOpenSSL now.

I'm not entirely clear on the value of this though. RFC 5746 was published in early 2010. Is insecure renegotiation prevalent at this point?

mhils:
Anecdotally we've had some people request this for mitmproxy (https://github.com/mitmproxy/mitmproxy/issues/5449 and https://github.com/mitmproxy/mitmproxy/pull/5698), but I don't have hard numbers. Do we know which pyOpenSSL/cryptography release disabled insecure renegotiation? CHANGELOG.rst isn't conclusive unfortunately.

reaperhulk:
OpenSSL would have disabled it, we never set an explicit value. It may or may not be listed in their changelog though…

mhils:
https://www.openssl.org/news/cl30.txt says it's OpenSSL 3.0, so that means cryptography 37 (2022-04-26).

reaperhulk:
Those mitmproxy issues suggest it’d be helpful to have the insecure renegotiation flags from the other PR available. I’m not a huge fan but being able to communicate with old devices that can’t be updated makes sad sense. If you think we should be landing one or both of these PRs I can be okay with that. We’ll need to re-add the bindings to cryptography as well as fixing the cryptographyMain CI issue though.

PR #1194, opened 2023-03-21
Add Context options for legacy renegotiation support

ndmalc:
Hello,

This is a pull request to add support for Context options used to support legacy renegotiation disable by default in recent OpenSSL versions.

Best regards


reaperhulk:
As noted in the other PR, these flags have been removed from the next cryptography release. We'd have to have a very good reason to want to explicitly re-add an insecure feature though.

PR #1202, opened 2023-03-27
Expose X509_V_* constants

### LINKED ISSUE #1201: Expose constants that are no longer available in latest cryptography
In order to customize TLS validation logic in pyOpenSSL, some constants are sometimes needed. For example, Tahoe-LAFS currently uses X509_V_ERR_CERT_NOT_YET_VALID, X509_V_ERR_SELF_SIGNED_CERT_IN_CHAIN, and a few others (https://github.com/tahoe-lafs/tahoe-lafs/blob/d92470d233533bdc8ae1f014ca3a82b0ce74e3f3/src/allmydata/storage/http_client.py#L220).

Previous to `cryptography` v40, these were available there, but they no longer are.

@alex suggested that:

1. They get added to public interface of `pyOpenSSL`, since that is the consumer.
2. `cryptography` is then updated appropriately to expose just what pyOpenSSL needs.

I assume they should be exposed in `OpenSSL.SSL`?

* [x] PR to `cryptography` adding the constants is merged.
* [ ] PR to this repository conditionally adding these to `OpenSSL.SSL.__all__`.

itamarst:
Fixes #1201 

The `cryptography` half of this is in https://github.com/pyca/cryptography/pull/8626

**Potential user hazard:** As currently implemented, users of `pyOpenSSL` who depend on this may be setting up _their_ users for weird errors if they happen to install `cryptography` 40. So possibly this should also include either:

* Changing minimal required `cryptography` to 41, if/when https://github.com/pyca/cryptography/pull/8626 is merged and v41 released.
* Or, don't change the minimal version, but do change `cryptography` dependency to forbid version 40, so you'd get versions before and after.
* Or, just assume this functionality is so obscure no one is likely to hit this in practice and do nothing.


reaperhulk:
I think we're going to backport and release those bindings in 40.0.2. I think I'd prefer to avoid bumping up the minimum version as well, so we'll want to be more explicit (40.0.0 and 40.0.1) for the conditional branch. Please also add a comment that says when we bump to a 41+ minimum we can remove the branch 😄

itamarst:
Anyway, how's this?

src/OpenSSL/SSL.py
```diff
@@ -250,6 +252,127 @@
...
+    )
+    X509_V_ERR_HOSTNAME_MISMATCH = _lib.X509_V_ERR_HOSTNAME_MISMATCH
+    X509_V_ERR_EMAIL_MISMATCH = _lib.X509_V_ERR_EMAIL_MISMATCH
+    X509_V_ERR_IP_ADDRESS_MISMATCH = _lib.X509_V_ERR_IP_ADDRESS_MISMATCH
+    X509_V_ERR_APPLICATION_VERIFICATION = (
+        _lib.X509_V_ERR_APPLICATION_VERIFICATION
+    )
+for name in list(globals().keys()):
+    if name.startswith("X509_V_"):
+        __all__.append(name)
```
alex, line 373: Instead of dumping everything at the top level, let's namespace it ala: https://github.com/pyca/pyopenssl/blob/main/src/OpenSSL/crypto.py#L1623

src/OpenSSL/SSL.py
```diff
@@ -250,6 +252,127 @@
 SSL_CB_HANDSHAKE_START = _lib.SSL_CB_HANDSHAKE_START
 SSL_CB_HANDSHAKE_DONE = _lib.SSL_CB_HANDSHAKE_DONE
 
+X509_V_OK = _lib.X509_V_OK
+# cryptography v40.0.0 and .1 releases are missing the X509_V_ERR_* codes;
+# previously they were exposed but not formally part of the public API. Once
+# pyOpenSSL has minimal required cryptography version of 41 this code can be
+# run unconditionally. See https://github.com/pyca/pyopenssl/issues/1206
+if _cryptography_version not in ("40.0.0", "40.0.1"):
```
alex, line 260: Please check my capabilities, not version. This is very fragile.

reaperhulk:
Looks like this needs a rebase and addressing the outstanding comments. We're going to do a release in the next few days so this can land if it gets done!

itamarst:
Ugh got distracted and didn't finish this. Will try to get to this tomorrow or next week.

itamarst:
OK should be ready for review again.

alex:
not related, I'll do a PR momentarily to fix it.

src/OpenSSL/SSL.py
```diff
@@ -250,6 +251,128 @@
...
+    """
+
+    OK = _lib.X509_V_OK
+
+    # cryptography v40.0.0 and .1 releases are missing the X509_V_ERR_* codes;
+    # previously they were exposed but not formally part of the public API.
+    # Once pyOpenSSL has minimal required cryptography version of 41 this code
+    # can be run unconditionally. See
+    # https://github.com/pyca/pyopenssl/issues/1206
+    if hasattr(_lib, "X509_V_ERR_CERT_SIGNATURE_FAILURE"):
```
alex, line 273: Given that .0 and .1 are excluded in the `install_requires`, do we need this check? Is it reachable?

alex (commented):
One comment, besides that this LGTM.

src/OpenSSL/SSL.py
itamarst: Hmm. 

1. Someone could force override it.
2. Conceivably (though not AFAIK) some other past version might have it.
3. A future version could drop it (but then silently failing is bad).

So either this can be dropped, yeah, or the `setup.py` restriction can be dropped... Having inconsistent APIs based on what's installed is weird, so may as well drop this.

itamarst:
Addressed.

CHANGELOG.rst
```diff
@@ -20,6 +20,8 @@ Changes:
 ^^^^^^^^
 
 - Invalid versions are now rejected in ``OpenSSL.crypto.X509Req.set_version``.
+- Added ``X509_V_*` constants to ``OpenSSL.SSL``.
```
alex, line 23:
```suggestion
- Added ``X509VerificationCodes`` to ``OpenSSL.SSL``.
```

alex (approved):

lunaperezclaudio-svg:
2464707387

PR #1210, opened 2023-04-15
add support for Context.set_cert_store

reaperhulk:


src/OpenSSL/SSL.py
```diff
@@ -1399,6 +1399,24 @@ def get_cert_store(self):
...
+        Set the certificate store for the context.
+        :param store: A X509Store object.
+        :return: None
+        """
+        try:
+            _lib.SSL_CTX_set_cert_store(self._context, store._store)
+            # The store is now owned by the context, so we need to
+            # remove the gc free in the object. We do this after the
+            # set since set may not exist.
+            _ffi.gc(store._store, None)
```
alex, line 1413: This seems like the very wrong way to do this. We should be bumping the refcount here, not getting rid of the GC. Doing it this way means it's totally wrong if you put the same store on 2x ssl contexts

src/OpenSSL/SSL.py
```diff
@@ -1399,6 +1399,24 @@ def get_cert_store(self):
...
+        except AttributeError:
```
alex, line 1414: This also catches attribute errors if store is the wrong type. Bad.

tox.ini
```diff
@@ -19,6 +19,8 @@ extras =
 deps =
     coverage>=4.2
     cryptographyMinimum: cryptography==38.0.0
+    # special version to test paths for bindings we temporarily removed
+    cryptography40: cryptography==40.0.1
```
alex, line 23: Can we just bump the minimum, this is too complicated

src/OpenSSL/SSL.py
reaperhulk: We don't have that binding 😂 Good catch on the multiple adds though. We can't support this without adding the upref binding then.

reaperhulk:
This is blocked on a release that contains https://github.com/pyca/cryptography/pull/8737 now.

ankit-gautam23 (approved):
LGTM

src/OpenSSL/SSL.py
```diff
@@ -1507,6 +1507,16 @@ def get_cert_store(self):
         pystore._store = store
         return pystore
 
+    def set_cert_store(self, store):
```
alex, line 1510: Can we added type annotations? Seems like we should start adding them as we go.

tests/test_ssl.py
```diff
@@ -1716,6 +1716,12 @@ def test_get_cert_store(self):
         store = context.get_cert_store()
         assert isinstance(store, X509Store)
 
+    def test_set_cert_store(self):
+        context = Context(SSLv23_METHOD)
+        store = X509Store()
+        context.set_cert_store(store)
+        assert store._store == context.get_cert_store()._store
```
alex, line 1723: Can we make this test actually verify what we want?

src/OpenSSL/SSL.py
jlaine: +1 for type annotations :)

Issue #1235, opened 2023-08-07
HUE build failing with error message, need edge case handling


Madhukar98:
--- Building egg for pyOpenSSL-17.5.0
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/mnt/c/Users/Madhukar/Desktop/hue/build/env/local/lib/python2.7/site-packages/setuptools/sandbox.py", line 253, in run_setup
    raise
  File "/usr/lib/python2.7/contextlib.py", line 35, in __exit__
    self.gen.throw(type, value, traceback)
  File "/mnt/c/Users/Madhukar/Desktop/hue/build/env/local/lib/python2.7/site-packages/setuptools/sandbox.py", line 195, in setup_context
    yield
  File "/usr/lib/python2.7/contextlib.py", line 35, in __exit__
    self.gen.throw(type, value, traceback)
  File "/mnt/c/Users/Madhukar/Desktop/hue/build/env/local/lib/python2.7/site-packages/setuptools/sandbox.py", line 166, in save_modules
    saved_exc.resume()
  File "/mnt/c/Users/Madhukar/Desktop/hue/build/env/local/lib/python2.7/site-packages/setuptools/sandbox.py", line 141, in resume
    six.reraise(type, exc, self._tb)
  File "/mnt/c/Users/Madhukar/Desktop/hue/build/env/local/lib/python2.7/site-packages/setuptools/sandbox.py", line 154, in save_modules
    yield saved
  File "/mnt/c/Users/Madhukar/Desktop/hue/build/env/local/lib/python2.7/site-packages/setuptools/sandbox.py", line 195, in setup_context
    yield
  File "/mnt/c/Users/Madhukar/Desktop/hue/build/env/local/lib/python2.7/site-packages/setuptools/sandbox.py", line 250, in run_setup
    _execfile(setup_script, ns)
  File "/mnt/c/Users/Madhukar/Desktop/hue/build/env/local/lib/python2.7/site-packages/setuptools/sandbox.py", line 45, in _execfile
    exec(code, globals, locals)
  File "setup.py", line 54, in <module>
    "\n\n`Full changelog " +
AttributeError: 'NoneType' object has no attribute 'group'

alex:
You are using a very old version of pyOpenSSL (6 years old) with Python 2.7, which is no longer supported.

Can you reproduce this with a recent, supported, version of Python and pyOpenSSL?

Madhukar98:
[...]

PR #1236, opened 2023-08-07
Handling edgecase for search operation

Madhukar98:
W.R.T [issue-1235](https://github.com/pyca/pyopenssl/issues/1235), handling edge case when no pattern is detected.


alex:
This PR doesn't explain _why_ its necessary. This regular expression is being run against a file in the repo, so we know it will always match.

Issue #1238, opened 2023-08-10
X509Extension __str__() method raises exception for unknown extension types


ich199:
### Issue

When calling the `crypto.X509Extension` method [`__str__()`](https://github.com/pyca/pyopenssl/blob/b259bfbd660f5680783e1abfab27ed0d5bf91c1a/src/OpenSSL/crypto.py#L870), if the extension type is not supported by OpenSSL, an `exception_type` error is raised:

```python
python -c 'from OpenSSL.crypto import X509Extension; print(str(X509Extension(b"1.2.3.4.5.6.7", False, b"DER:05:00")))'
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/home/iain/pyopenssl-test/lib64/python3.11/site-packages/OpenSSL/crypto.py", line 882, in __str__
    _openssl_assert(print_result != 0)
  File "/home/iain/pyopenssl-test/lib64/python3.11/site-packages/OpenSSL/_util.py", line 71, in openssl_assert
    exception_from_error_queue(error)
  File "/home/iain/pyopenssl-test/lib64/python3.11/site-packages/OpenSSL/_util.py", line 57, in exception_from_error_queue
    raise exception_type(errors)
OpenSSL.crypto.Error: []
```

### Environment:

```bash
$ python --version
Python 3.11.4

$ pip list
Package      Version
------------ -------
cffi         1.15.1
cryptography 41.0.3
pip          23.2.1
pycparser    2.21
pyOpenSSL    23.2.0
setuptools   62.6.0

$ openssl version
OpenSSL 3.0.9 30 May 2023 (Library: OpenSSL 3.0.9 30 May 2023)
```

### Possible Cause
The current call to [`X509V3_EXT_print()`](https://github.com/pyca/pyopenssl/blob/b259bfbd660f5680783e1abfab27ed0d5bf91c1a/src/OpenSSL/crypto.py#L878C42-L878C42) in `crypto.X509Extension` sets the `flags` parameter (3rd parameter in the call) to `0`, which causes OpenSSL to [return an error](https://github.com/openssl/openssl/blob/master/include/openssl/x509v3.h.in#L489C42-L489C42) for unknown extension types:

```python
print_result = _lib.X509V3_EXT_print(bio, self._extension, 0, 0)
```

### Possible Fix

[...]

Issue #270, opened 2015-05-28
Crypto fail on SCT retrieval


Talanor:
When executing this code (http://hastebin.com/arinusefex.py) on "twitter.com" (which only displays basic informations and X509 extensions of the certificate), it raises an exception on the last extension, which oid is : 1.3.6.1.4.1.11129.2.4.2.

reaperhulk:
1.3.6.1.4.1.11129.2.4.2 is the OID for certificate transparency (RFC 6962). Since your install (and indeed almost all installs) of OpenSSL don't know this extension an error is raised when __str__ is called, which calls `X509V3_EXT_print`. pyOpenSSL should definitely just provide a __str__ of "unknown extension" or something when it encounters an OID it doesn't recognize.

Talanor:
Thank you. Additionally, would you have any idea as of why my openssl command seem to be able to display the informations properly and pyopenssl can't?

reaperhulk:
What version of openssl? I can parse CT Precertificate SCTs with 1.0.2, but it's possible your pyopenssl isn't linked against that. To test you can run:

```
python -c "import OpenSSL.SSL; print(OpenSSL.SSL.SSLeay_version(OpenSSL.SSL.SSLEAY_VERSION))"
```

Talanor:
It prints out ```OpenSSL 1.0.1f 6 Jan 2014```

and ```openssl version``` also returns ```OpenSSL 1.0.1f 6 Jan 2014```

reaperhulk:
Hmm, I'm not sure why you're having problems then. I believe `x509 -text` uses the same `X509V3_EXT_print` function we're calling, but maybe I'm mistaken?

Talanor:
After reading some code, it would seem (not sure at all, correct me if i'm wrong), that cffi compiles python wrappers to call ```X509V3_EXT_print``` (and a bunch of other functions). Maybe the problem comes from the wrapper?

PR #1239, opened 2023-08-10
fix X509Extension __str__() method for unknow extension types

ich199:
fixes: https://github.com/pyca/pyopenssl/issues/1238
fixes: https://github.com/pyca/pyopenssl/issues/270


ich199:
tests seem to be failing due to https://github.com/pyca/pyopenssl/issues/1240

mhils:
Thanks @ich199! The change generally makes sense to me. Would you mind adding a `X509V3_EXT_ERROR_UNKNOWN` binding to cryptography and then use this here instead of defining a custom constant?

ich199:
Thanks @mhils, I've created the binding in cryptography as requested and that's been merged. I'm guessing it would be preferred to backport that to `41.0.x` rather than bumping the cryptography requirement here?

PR #1241, opened 2023-08-10
switch linting to ruff

reaperhulk:


tox.ini
```diff
@@ -41,16 +41,15 @@ commands =
...
     python -m twisted.trial -j4 --reporter=text twisted
 
-[testenv:flake8]
+[testenv:lint]
 basepython = python3
 deps =
     black
-    flake8
-    flake8-import-order
+    ruff==0.0.284
```
alex, line 48: This is literally our only pin, and we have no mechanism for bumping. We need to either unpin or do this in a dependabot friendly way IMO.

reaperhulk: We have a pin for mypy in the exact same way here. It definitely isn't dependabot friendly though, I agree 😄

alex: Its also not local dev friendly (since I know neither of us tox)

alex (approved):
Going to merge this to get us back to green

PR #1252, opened 2023-09-20
Add support for `cryptography` CRLs to `X509Store`

facutuesca:
As a follow up to deprecating CRL APIs (https://github.com/pyca/pyopenssl/issues/1249), there is one remaining function that won't be deprecated but accepts the `CRL` type: `X509Store.add_crl(crl: CRL)`. In order to [give users an alternative](https://github.com/pyca/pyopenssl/issues/1249#issuecomment-1726834527) once `CRL` is deprecated, this PR adds support to `add_crl()` so that it also accepts `cryptography`'s `X509.CertificateRevocationList`.

This PR also adds tests for the new code path, by parametrizing existing tests that use `add_crl` so that the CRL being passed as a parameter is either PyOpenSSL's or cryptography's. For example:
```python
    @pytest.mark.parametrize(
        "create_crl",
        [
            pytest.param(
                _make_test_crl,
                id="pyOpenSSL CRL",
            ),
            pytest.param(
                _make_test_crl_cryptography,
                id="cryptography CRL",
            ),
        ],
    )
    def test_verify_with_revoked(self, create_crl):
            #...
            root_crl = create_crl(self.root_cert, self.root_key, certs=[self.intermediate_cert])
            #...
 ```

### DIFF
--- a/src/OpenSSL/crypto.py
+++ b/src/OpenSSL/crypto.py
@@ -1720,7 +1720,7 @@ class X509Store:
         res = _lib.X509_STORE_add_cert(self._store, cert._x509)
         _openssl_assert(res == 1)
 
-    def add_crl(self, crl: "CRL") -> None:
+    def add_crl(self, crl: "_CRLInternal") -> None:
         """
         Add a certificate revocation list to this store.
 
@@ -2404,7 +2404,7 @@ class CRL:
     @classmethod
     def from_cryptography(
         cls, crypto_crl: x509.CertificateRevocationList
-    ) -> "CRL":
+    ) -> "_CRLInternal":
         """
         Construct based on a ``cryptography`` *crypto_crl*.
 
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: src/OpenSSL/crypto.py [hunk 1]
```diff
@@ -2428,7 +2448,7 @@ def get_revoked(self) -> Optional[Tuple[_RevokedInternal, ...]]:
             return tuple(results)
         return None
 
-    def add_revoked(self, revoked: Revoked) -> None:
+    def add_revoked(self, revoked: _RevokedInternal) -> None:
```
reaperhulk, line 2451: Did this fail mypy or are we avoiding deprecation warnings?

facutuesca: It's for the docs, see the current [type signature for this function](https://www.pyopenssl.org/en/latest/api/crypto.html#OpenSSL.crypto.CRL.add_revoked):
```
add_revoked(revoked: <cryptography.utils._DeprecatedValue object at 0x7f66c71ab1d0>) → None
Add a revoked (by value not reference) to the CRL structure
```

reaperhulk:
This still needs a changelog entry 😄

facutuesca:
> This still needs a changelog entry 😄

@reaperhulk My bad, fixed!

reaperhulk (approved):

Issue #1227, opened 2023-06-23
Stop using BN_set_word


reaperhulk:
We only use it to set a BN to `65537`. We can do this without that function (and then we can remove it in cryptography)

facutuesca:
@reaperhulk Could we close this one now?

PR #1253, opened 2023-09-21
Remove use of `BN_set_word`

facutuesca:
This replaces the only use of `BN_set_word` with a call to `BN_hex2bn`.

This fixes https://github.com/pyca/pyopenssl/issues/1227

### DIFF
--- a/src/OpenSSL/crypto.py
+++ b/src/OpenSSL/crypto.py
@@ -369,10 +369,13 @@ class PKey:
             if bits <= 0:
                 raise ValueError("Invalid number of bits")
 
-            # TODO Check error return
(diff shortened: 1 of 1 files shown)


alex:
I don't remember what @reaperhulk's motivation to remove `BN_set_word` was, but this seems more complicated and not worth it.

woodruffw:
Hmm, perhaps `BN_native2bn` is the correct replacement API here?

(But if the replacement is too complex regardless, we can drop this.)

facutuesca:
Closing due to solution not actually decreasing the bindings surface (we need to add a new binding in order to remove `BN_set_word`)

PR #1272, opened 2023-11-30
Fix various issues around X509_STORE_CTX reuse

davidben:
This is three commits that probably should be reviewed separately. The first is a minimal fix for a weird behavior in pyOpenSSL: inspecting an X509_STORE_CTX after you have cleaned it up is pretty odd.

The next two rework X509StoreContext to just create a new X509_STORE_CTX for each verification, which avoids a whole lot of historical mess around X509_STORE_CTX's transition from being stack-allocated to heap-allocated in the OpenSSL API.


src/OpenSSL/crypto.py
```diff
@@ -1941,28 +1929,8 @@ def cleanup(s: Any) -> None:
...
-        """
-        Internally cleans up the store context.
-
-        The store context can then be reused with a new call to :meth:`_init`.
-        """
-        _lib.X509_STORE_CTX_cleanup(self._store_ctx)
-
-    def _exception_from_context(self) -> X509StoreContextError:
+    @staticmethod
+    def _exception_from_context(store_ctx) -> X509StoreContextError:
```
alex, line 1933: This will need a type annotation. It can probable be `typing.Any`

davidben: Done.

alex (approved):

PR #1276, opened 2023-12-09
#1275 Add OpenSSL.SSL.Connection.session_reused API.

### LINKED ISSUE #1275: Use SSL_session_reused API
I guess that this should be `OpenSSL.SSL.Connection.session_reused()`

It should return the number 1 or 0 to keep it as close a possible to the OpenSSL API.

It should not return a boolean.

I see that `OpenSSL.SSL.Connection.renegotiate_pending()` return a boolean, but [SSL_renegotiate_pending](https://www.openssl.org/docs/manmaster/man3/SSL_renegotiate_pending.html) returns an int.

adiroiban:
Fixes #1275 

This add the SSL.Connection API added in https://github.com/pyca/cryptography/issues/9969

Still WIP


CHANGELOG.rst
```diff
@@ -4,6 +4,12 @@ Changelog
 Versions are year-based with a strict backward-compatibility policy.
 The third digit is only for regressions.
 
+
+- Added ``OpenSSL.SSL.Connection.session_reused()`` to query whether the
+  current session was reused during the last handshake.
+  [`#1275 <https://github.com/pyca/pyopenssl/issues/1275>`_]
```
adiroiban, line 10: Let me know if you want a link to a PR.

I have created the changelong before creating the PR... so at that time, I didn't had a PR ID.

src/OpenSSL/SSL.py
```diff
@@ -2674,6 +2674,25 @@ def set_session(self, session):
...
+        If both client and server agree on the session,
+        it will be reused and a flag is being set that can be queried by the
+        application.
+
+        Retruns `0` when a new session was negotiated.
+        Returns `1` when a the session was reused.
+
+        :returns: int
+
+        .. versionadded:: NEXT
```
adiroiban, line 2692: Not sure in which version it will be released.

src/OpenSSL/SSL.py
```diff
@@ -2674,6 +2674,25 @@ def set_session(self, session):
...
+        During the negotiation, a client can propose to reuse a session.
+        The server then looks up the session in its cache.
+        If both client and server agree on the session,
+        it will be reused and a flag is being set that can be queried by the
+        application.
+
+        Retruns `0` when a new session was negotiated.
+        Returns `1` when a the session was reused.
+
+        :returns: int
```
adiroiban, line 2690: I went with upstream API, but maybe is best to return a `bool`

tests/test_ssl.py
```diff
@@ -2755,43 +2758,143 @@ def test_set_session_wrong_args(self):
...
+    def test_session_reused(self):
+        """
+        `Connection.session_reused`, returns 0 for new connections..
+        """
+        ctx = Context(TLSv1_2_METHOD)
+        connection = Connection(ctx, None)
+
+        assert connection.session_reused() == 0
+
+    def test_client_set_session_tls1_2(self):
```
adiroiban, line 2770: I went with both TLS 1.2 and 1.3 tests since the session handling is a big different.

I will refactor the tests to share more code.

tests/test_ssl.py
```diff
@@ -2755,43 +2758,143 @@ def test_set_session_wrong_args(self):
...
-        ctx.use_certificate(cert)
-        ctx.set_session_id(b"unity-test")
+        server_ctx = Context(TLSv1_2_METHOD)
+        server_ctx.use_privatekey(key)
+        server_ctx.use_certificate(cert)
+        # !!!!
+        # I have no idea why it works when server-side cache is disabled.
+        # I guess that this might be because server and client are in the
+        # same process.
+        server_ctx.set_session_cache_mode(SSL.SESS_CACHE_OFF)
```
adiroiban, line 2787: I went with explicit context for server and client.

Somehow for this test for TLS 1.2, it works even when cache is off.

For my end to end test, in which I use 2 separate processes, the server cache needs to be enabled for session reuse.

tests/test_ssl.py
```diff
@@ -2755,43 +2758,143 @@ def test_set_session_wrong_args(self):
...
+        # This is a proxy: in general, we have no access to any unique
+        # identifier for the session (new enough versions of OpenSSL expose
+        # a hash which could be usable, but "new enough" is very, very new).
+        # Instead, exploit the fact that the master key is re-used if the
+        # session is re-used.  As long as the master key for the two
+        # connections is the same, the session was re-used!
+        assert originalServer.master_key() == resumedServer.master_key()
+        assert originalClient.master_key() == resumedClient.master_key()
+
+    def test_client_set_session_tls1_3(self):
```
adiroiban, line 2837: I have no idea why tls 1.3 fails in this test.

I have it working in a separate manual  proof of concept code, in which I have the server with pyOpenSSL and the client is curl.

Issue #1278, opened 2024-01-10
Add support for retrieving negotiated SRTP profile


jlaine:
The `aiortc` project uses `pyOpenSSL` to perform a DTLS handshake and export SRTP keying material. Currently, the only enable a single SRTP profile using `ctx.set_tlsext_use_srtp(b"SRTP_AES128_CM_SHA1_80")` but we would like to start enabling GCM profiles (https://github.com/aiortc/aiortc/issues/1012).

Enabling multiple SRTP profiles means that when the handshake completes, we need to know which SRTP profile was selected, for which OpenSSL provides `SSL_get_selected_srtp_profile`:

https://www.openssl.org/docs/man3.0/man3/SSL_get_selected_srtp_profile.html

This function is present in `cryptography`'s bindings, so I would like to add a `Connection.get_selected_srtp_profile` method to wrap it, and return `bytes`:

- empty if no SRTP profile was negotiated - or should this be `None`?
- the name of the SRTP profile if one was negotiated

Using `bytes` as the return value makes the API consistent with `Context.set_tlsext_use_srtp`.

I'm happy to submit a PR if this sounds acceptable.

jlaine:
I think I borked the change log, I added my change at the bottom of what I assumed was the "latest" code but it looks as though I added it to an existing release!!

jlaine:
> Ooops, if you can send a cleanup PR moving it ot the right place, that'd be great.

Sure, what's the usual phrasing for the "next release" section?

alex:
It'll be `24.0.0 (UNRELEASED)` because calver. See https://github.com/pyca/pyopenssl/commit/5ba8ce10ed7c318e57516a7ec8447cbb5626d3f9 for what the last bump looked like.

PR #1279, opened 2024-01-10
Add Connection.get_selected_srtp_profile (#1278)

jlaine:
If an SRTP profile was negotiated as part of the handshake, make it possible to retrieve the name of the profile. This is needed to determine which profiles were offered using `Context.set_tlsext_use_srtp` was actually selected.


mhils (approved):
Returning `b""` and not `None` feels a bit weird, but we do the same for ALPN, so it's probably the right decision in terms of consistency. LGTM!

jlaine:
Thanks @mhils !

PR #1281, opened 2024-01-10
Replace RSA-512 test keys with RSA-2048

davidben:
RSA-512 was factored in 1999. Since none of these tests are specific to the size of the RSA key, best to test inputs that are representative of real-world inputs, lest the underlying cryptography library start enforcing post-1999 security levels.


mhils (approved):
People like to copy - even from test code -, so this is a definitely good! 👍

PR #1287, opened 2024-01-25
Set SSL_MODE_ACCEPT_MOVING_WRITE_BUFFER in calling OpenSSL

julianz-:
See cherrypy/cheroot#245 for discussion.


mhils:
See https://github.com/pyca/pyopenssl/pull/1242 for more context.

CHANGELOG.rst
```diff
@@ -4,6 +4,21 @@ Changelog
 Versions are year-based with a strict backward-compatibility policy.
 The third digit is only for regressions.
 
+24.1.0 (2024-01-25)
+-------------------
+Backward-incompatible changes:
```
mhils, line 9:
```suggestion
24.1.0 (UNRELEASED)
-------------------

Backward-incompatible changes:
```

CHANGELOG.rst
```diff
@@ -4,6 +4,21 @@ Changelog
...
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+- modified SSL.py to set SSL_MODE_ACCEPT_MOVING_WRITE_BUFFER in calling OpenSSL. See cherrypy/cheroot#245 for discussion.
```
mhils, line 11: We can reword this to emphasize the behavior change ("pyOpenSSL now sets `SSL_MODE_ACCEPT_MOVING_WRITE_BUFFER` by default, matching CPython's behavior") over the implementation work that has been done. The specific file that has been modified is likely not as interesting for readers. :)

Also, it's probably useful to link (see below for syntax) to this PR here, which then has further links to the previous PR and the cherrypy discussion.

mhils (commented):
Thanks for pushing this further! 🍰 I think we had general agreement to do this in #1242, so this looks good to merge after some minor docs fixes. :)

We can ignore codecov, that seems to be a bug in determining coverage.

CHANGELOG.rst
```diff
@@ -4,6 +4,22 @@ Changelog
 Versions are year-based with a strict backward-compatibility policy.
 The third digit is only for regressions.
 
+24.1.0 (UNRELEASED)
+-------------------
+Backward-incompatible changes:
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+- ``pyOpenSSL`` now sets ``SSL_MODE_ACCEPT_MOVING_WRITE_BUFFER`` by default, matching CPython's behavior. 
+- The minimum ``cryptography`` version is now 42.0.0
+`#1287 <https://github.com/pyca/pyopenssl/pull/1287>`_.
```
mhils, line 13:
```suggestion
- ``pyOpenSSL`` now sets ``SSL_MODE_ACCEPT_MOVING_WRITE_BUFFER`` by default, matching CPython's behavior. 
  `#1287 <https://github.com/pyca/pyopenssl/pull/1287>`_.
- The minimum ``cryptography`` version is now 42.0.0
```

(see the rendering preview on GitHub :))

CHANGELOG.rst
julianz-: Thanks! I made those changes you suggested. Regarding code coverage I noticed in SSL.py, one area coverage is missing is here:

```

172 | except AttributeError:
173 | # Hardcode constants for cryptography < 3.4, see
174 | # https://github.com/pyca/pyopenssl/pull/985#issuecomment-775186682
175 | SSL3_VERSION = 768
176 | TLS1_VERSION = 769
177 | TLS1_1_VERSION = 770
178 | TLS1_2_VERSION = 771
179 | TLS1_3_VERSION = 772

```

If we are requiring cryptography >= 42.0.0, can the exception block be replaced by `pass` or something?

mhils: Thanks for spotting that! We can do away with the entire try/catch block(s) here and only keep the contents. On a second look that's probably what codecov is complaining about, I was mesmerized by the diff and didn't notice the "indirect changes" tab. 🫣

julianz-: Looks like that did the trick! Can you approve the merge?

tests/test_crypto.py
```diff
@@ -1740,7 +1740,7 @@ def test_construction(self):
         certificate = X509()
         assert isinstance(certificate, X509)
         assert type(certificate).__name__ == "X509"
-        assert type(certificate) == X509
+        assert type(certificate) is X509
```
webknjaz, line 1743: Are these identity check changes related to the PR?

src/OpenSSL/SSL.py
```diff
@@ -163,20 +163,11 @@
 DTLS_SERVER_METHOD = 11
 DTLS_CLIENT_METHOD = 12
 
-try:
```
webknjaz, line 166: Should this be a part of the PR?

CHANGELOG.rst
```diff
@@ -4,6 +4,21 @@ Changelog
...
+- ``pyOpenSSL`` now sets ``SSL_MODE_ACCEPT_MOVING_WRITE_BUFFER`` by default, matching CPython's behavior. `#1287 <https://github.com/pyca/pyopenssl/pull/1287>`_.
+- The minimum ``cryptography`` version is now 42.0.0.
+
+Deprecations:
+^^^^^^^^^^^^^
+
+Changes:
+^^^^^^^^
+
+
```
webknjaz, line 20: Maybe, don't add these empty sections?

CHANGELOG.rst
```diff
@@ -4,6 +4,21 @@ Changelog
 Versions are year-based with a strict backward-compatibility policy.
 The third digit is only for regressions.
 
+24.1.0 (UNRELEASED)
+-------------------
```
webknjaz, line 8:
```suggestion
-------------------

```

src/OpenSSL/SSL.py
webknjaz: Ah, nevermind. I see this was agreed earlier.

tests/test_crypto.py
julianz-: Not really - I added those to get the PR to pass all the tests. Should I separate these things out into a separate PR?

webknjaz: Normally, yes — it's a good idea to submit atomic changes as it's harder to reason about things when a PR title/description refers to the change in a single line and the diff is flooded with an unproportional amount of arbitrary changes that don't support its stated purpose. This affects the cognitive complexity of the patch as the mental burden for the reader/reviewer gets higher.

However, the maintainers may still decide to accept such PRs on occasion. Especially since the activities aren't happening as often in this repo. And ultimately it's up to them.

Though, what I'm talking about is always applicable in general — smaller patches require less time to accept, because they are easier to understand and when there's more changes, a problem with parts of them would block/postpone the bits that would be otherwise acceptable.

julianz-: I submitted a separate PR for these fixes and that's now been approved. Do I need to scrap this PR and create a new one? Or can this one now be accepted anyway?

CHANGELOG.rst
julianz-: I was just following the way all previous entries in the Changelog have been done. They keep these empty entries no?

tests/test_crypto.py
webknjaz: You can perform a rebase and that'll make the diff here smaller.

CHANGELOG.rst
webknjaz: Alright, makes sense.

CHANGELOG.rst
webknjaz: @julianz- you can accept this with a click of a button right here.

CHANGELOG.rst
```diff
@@ -4,6 +4,21 @@ Changelog
...
+Backward-incompatible changes:
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
webknjaz, line 10:
```suggestion
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

```

webknjaz (commented):
Add some empty line separators under title marks.

tests/test_crypto.py
julianz-: I did the rebase but was a bit of a mess. Somehow my branch I rebased still says it is one commit behind the origin.

julianz-: I synced the PR with all commits now so it should be good to go now

webknjaz: > I did the rebase but was a bit of a mess. Somehow my branch I rebased still says it is one commit behind the origin.

That usually happens when you forget to `git fetch --all`. And rebasing should be done against the remote branch (the local one with the same name tends to be outdate).

webknjaz: Once fetched, if your pointer to this upstream repository is called `origin`, use `git rebase origin/main`. If you use your fork as origin, then you may want to add a separate remote pointing to this repo too. You can do something like `git remote add upstream git@github.com:pyca/pyopenssl.git` and `git fetch upstream`. Then your rebase command would be `git rebase upstream/main` (assuming you're on the local `jan23-2024` branch when running it). If you're having troubles with empty commits, it's possible to clean them up with an interactive rebase, I suppose.

CHANGELOG.rst
webknjaz: I have a collection of articles related to Git and contributions: https://gist.github.com/webknjaz/a5a4fb374b7579de827e6bedb93a5220 / https://gist.github.com/webknjaz/cb7d7bf62c3dda4b1342d639d0e78d79 / https://gist.github.com/webknjaz/a7362787a80067af8621a85a71746ca1 / https://gist.github.com/webknjaz/2aa9c7a43b51c1385260ff87e0cbb9e3.
Maybe, there's some references that can help you with Git.

setup.py
```diff
@@ -93,11 +93,7 @@ def find_meta(meta):
         packages=find_packages(where="src"),
         package_dir={"": "src"},
         install_requires=[
-<<<<<<< HEAD
```
webknjaz: @julianz- I see you may be struggling with Git a little. I'll try to help.

If you want to melt the last commit into the previous one, do the following:
```console
$ git rebase -i HEAD^^
```
That will open an editor (for me it's NeoVim, you can have something different set up, like VS Code).
In the opened text file, in the "rebase plan", find the last entry with the commit 9c22966.
On that line, change the first word "pick" to just one letter "f".
Save the change and close the editor.
At this point, you'll end up with the commit you want, and you can force-push that.

julianz-: I just separated out some code into a different PR which was approved. But I am now having difficulty in rebasing this PR so that it doesn't include that code change. The rebased version keeps saying it's one commit behind the upstream main version even though I rebased against that version. I did 

```git fetch --all```

as you advised previously for this kind of circumstance - and which worked before.

Comment (julianz-) [reply setup.py]: @webknjaz Ok I got past that problem (not sure why but using  git merge-base jan23-2024 upstream/main was not giving me the right result) but now the code coverage is failing even though I already dealt with the block of code that never runs because of the cryptography version on the previous PR. Let me know what you think. Thanks!

Comment (webknjaz) [reply setup.py]: @julianz- the coverage report page (https://app.codecov.io/gh/pyca/pyopenssl/pull/1287/indirect-changes#1beeb2c0c32f16d580bcd3b7da99eecc-R229) indicates an indirect coverage change @ https://github.com/pyca/pyopenssl/blob/482b95e/src/OpenSSL/SSL.py#L230.
However, that line isn't covered on `main`. So maybe, another rebase will clear that confusion. If it doesn't — ignore it.

Comment (alex) [reply setup.py]: It's showing up because `OP_LEGACY_SERVER_CONNECT` always exists with the new, higher, cryptography minimum version.

You can remove the `except AttributeError`

**Comment by webknjaz** [comment]
@mhils @alex it appears that https://app.codecov.io/gh/pyca/pyopenssl treats `master` as the default branch, which is why it may get confused and assign coverage drops to pull requests that have nothing to do with said lines.

File: setup.py (hunk 9 above)
Comment (webknjaz) [reply setup.py]: Most likely, the underlying OpenSSL lib used in tests got upgraded to a version that always provides that constant. cc @mhils

**Comment by alex** [comment]
I've changed the default branch to be main

File: setup.py (hunk 9 above)
Comment (webknjaz) [reply setup.py]: > You can remove the `except AttributeError`

Ah, makes sense. I wasn't entirely sure. I think the confusion of `main` vs. `master` threw me off to make an incorrect assumption. Seems like you're right and it is indeed related to the change in this PR.

Comment (julianz-) [reply setup.py]: @alex @webknjaz I removed the check on OP_LEGACY_SERVER_CONNECT. That's seems to have cleared the coverage problem. Thx!

**Comment by webknjaz** [comment]
Is there a smoke test we could include in this PR?

**Comment by julianz-** [comment]
The only test for setting the mode currently is:

```
def test_set_mode(self):
        """
        `Context.set_mode` accepts a mode bitvector and returns the
        newly set mode.
        """
        context = Context(SSLv23_METHOD)
        assert MODE_RELEASE_BUFFERS & context.set_mode(MODE_RELEASE_BUFFERS)
```

This checks that setting the mode for MODE_RELEASE_BUFFERS returns the same bit. I guess we could add another check to make sure passing SSL_MODE_ACCEPT_MOVING_WRITE_BUFFER also returns the appropriate value? Not sure whether that counts as a smoke test but it's something perhaps?

**Comment by webknjaz** [comment]
Maybe, find some existing test where a write buffer is passed, copy it and pass a moving buffer there?

@mhils ideas?

**Comment by mhils** [comment]
We could possibly adapt something like this: https://github.com/pyca/pyopenssl/blob/main/tests/test_ssl.py#L2837

If that's easy to add I'm all for it, but I also feel that not having an elaborate test here is not the end of the world. There's precedent (SSL_MODE_ENABLE_PARTIAL_WRITE has no real test) and, more importantly, if this fails in the future we should get a very explicit 'bad write retry' error. Does CPython have a dedicated test for this?

This looks good to merge otherwise. @alex @reaperhulk, if you want to make a judgement call here please just merge. :)

**Comment by julianz-** [comment]
@mhils Strangely, when I try running pytest locally on my branch I am getting an exception on that test that makes it fail the test:

```
E       OpenSSL.SSL.Error: [('system library', '', ''), ('system library', '', ''), ('system library', '', ''), ('system library', '', ''), ('SSL routines', '', 'certificate verify failed')]

../../../../Library/Python/3.9/lib/python/site-packages/OpenSSL/_util.py:57: Error
____________________________________________________________________ TestConnection.test_wantWriteError _____________________________________________________________________
```

That's just a fragment of the output so maybe not very meaningful but as I understand it, the test is meant to throw WantWriteError but for some reason although it's expected the exception is not being considered a success? How is this supposed to work?

**Comment by webknjaz** [comment]
> although it's expected the exception is not being considered a success?

In your log, a more generic exception happens (`OpenSSL.SSL.Error`), not `OpenSSL.SSL.WantWriteError`. This is why `pytest.raises()` doesn't match it as expected.

**Comment by julianz-** [comment]
@webknjaz  

> > although it's expected the exception is not being considered a success?
> 
> In your log, a more generic exception happens (`OpenSSL.SSL.Error`), not `OpenSSL.SSL.WantWriteError`. This is why `pytest.raises()` doesn't match it as expected.

Ok to make any progress on this, I'm trying to understand the first test that is failing when I run pytest locally - test_set_default_verify_paths() in test_ssl.py. It's basically saying "certificate verify failed". I tried debugging this using the following in the command line:

``` openssl s_client -connect "encrypted.google.com":443 ```

and get:

``` Verify return code: 18 (self signed certificate)```

I assume this generates an error because the return code is not 0. But how is this supposed to work?  The test relies on Google's certificate but there seems to be problem with the cert?

**Comment by webknjaz** [comment]
> I tried debugging this using the following in the command line:
> 
> `openssl s_client -connect "encrypted.google.com":443`
> 
> and get:
> 
> ` Verify return code: 18 (self signed certificate)`
> 
> I assume this generates an error because the return code is not 0. But how is this supposed to work? The test relies on Google's certificate but there seems to be problem with the cert?

Yes, it is customary to interpret any non-zero return code as an error. I believe, this convention originated in POSIX.

Not sure how you got a self-signed certificate, though. Have you set up your DNS / `/etc/hosts` to remap the domain to something else?

The actual host responds with the correct certificate and the handshake succeeds:
```console
$ openssl s_client -connect "encrypted.google.com":443
CONNECTED(00000004)
depth=2 C = US, O = Google Trust Services LLC, CN = GTS Root R1
verify return:1
depth=1 C = US, O = Google Trust Services LLC, CN = GTS CA 1C3
verify return:1
depth=0 CN = *.google.com
verify return:1
---
Certificate chain
 0 s:CN = *.google.com
   i:C = US, O = Google Trust Services LLC, CN = GTS CA 1C3
   a:PKEY: id-ecPublicKey, 256 (bit); sigalg: RSA-SHA256
   v:NotBefore: Feb 19 08:03:54 2024 GMT; NotAfter: May 13 08:03:53 2024 GMT
 1 s:C = US, O = Google Trust Services LLC, CN = GTS CA 1C3
   i:C = US, O = Google Trust Services LLC, CN = GTS Root R1
   a:PKEY: rsaEncryption, 2048 (bit); sigalg: RSA-SHA256
   v:NotBefore: Aug 13 00:00:42 2020 GMT; NotAfter: Sep 30 00:00:42 2027 GMT
 2 s:C = US, O = Google Trust Services LLC, CN = GTS Root R1
   i:C = BE, O = GlobalSign nv-sa, OU = Root CA, CN = GlobalSign Root CA
   a:PKEY: rsaEncryption, 4096 (bit); sigalg: RSA-SHA256
   v:NotBefore: Jun 19 00:00:42 2020 GMT; NotAfter: Jan 28 00:00:42 2028 GMT
---
Server certificate
-----BEGIN CERTIFICATE-----
MIIOXTCCDUWgAwIBAgIRAJF3iEqsDpoECedLdRgGyygwDQYJKoZIhvcNAQELBQAw
RjELMAkGA1UEBhMCVVMxIjAgBgNVBAoTGUdvb2dsZSBUcnVzdCBTZXJ2aWNlcyBM
TEMxEzARBgNVBAMTCkdUUyBDQSAxQzMwHhcNMjQwMjE5MDgwMzU0WhcNMjQwNTEz
MDgwMzUzWjAXMRUwEwYDVQQDDAwqLmdvb2dsZS5jb20wWTATBgcqhkjOPQIBBggq
hkjOPQMBBwNCAAQ7UkR1HmtagYbBBsPVqVik2xZO85oyD4jqvtxb2zKUdBCN3n+E
YFxtMs4KiRDvMzp3C/xWfNaMoF3X7/sDUm3ro4IMPjCCDDowDgYDVR0PAQH/BAQD
AgeAMBMGA1UdJQQMMAoGCCsGAQUFBwMBMAwGA1UdEwEB/wQCMAAwHQYDVR0OBBYE
FO620PFJar/4etLB8PcVeSgTJybuMB8GA1UdIwQYMBaAFIp0f6+Fze6VzT2c0OJG
FPNxNR0nMGoGCCsGAQUFBwEBBF4wXDAnBggrBgEFBQcwAYYbaHR0cDovL29jc3Au
cGtpLmdvb2cvZ3RzMWMzMDEGCCsGAQUFBzAChiVodHRwOi8vcGtpLmdvb2cvcmVw
by9jZXJ0cy9ndHMxYzMuZGVyMIIJ7wYDVR0RBIIJ5jCCCeKCDCouZ29vZ2xlLmNv
bYIWKi5hcHBlbmdpbmUuZ29vZ2xlLmNvbYIJKi5iZG4uZGV2ghUqLm9yaWdpbi10
ZXN0LmJkbi5kZXaCEiouY2xvdWQuZ29vZ2xlLmNvbYIYKi5jcm93ZHNvdXJjZS5n
b29nbGUuY29tghgqLmRhdGFjb21wdXRlLmdvb2dsZS5jb22CCyouZ29vZ2xlLmNh
ggsqLmdvb2dsZS5jbIIOKi5nb29nbGUuY28uaW6CDiouZ29vZ2xlLmNvLmpwgg4q
Lmdvb2dsZS5jby51a4IPKi5nb29nbGUuY29tLmFygg8qLmdvb2dsZS5jb20uYXWC
DyouZ29vZ2xlLmNvbS5icoIPKi5nb29nbGUuY29tLmNvgg8qLmdvb2dsZS5jb20u
bXiCDyouZ29vZ2xlLmNvbS50coIPKi5nb29nbGUuY29tLnZuggsqLmdvb2dsZS5k
ZYILKi5nb29nbGUuZXOCCyouZ29vZ2xlLmZyggsqLmdvb2dsZS5odYILKi5nb29n
bGUuaXSCCyouZ29vZ2xlLm5sggsqLmdvb2dsZS5wbIILKi5nb29nbGUucHSCDyou
Z29vZ2xlYXBpcy5jboIRKi5nb29nbGV2aWRlby5jb22CDCouZ3N0YXRpYy5jboIQ
Ki5nc3RhdGljLWNuLmNvbYIPZ29vZ2xlY25hcHBzLmNughEqLmdvb2dsZWNuYXBw
cy5jboIRZ29vZ2xlYXBwcy1jbi5jb22CEyouZ29vZ2xlYXBwcy1jbi5jb22CDGdr
ZWNuYXBwcy5jboIOKi5na2VjbmFwcHMuY26CEmdvb2dsZWRvd25sb2Fkcy5jboIU
Ki5nb29nbGVkb3dubG9hZHMuY26CEHJlY2FwdGNoYS5uZXQuY26CEioucmVjYXB0
Y2hhLm5ldC5jboIQcmVjYXB0Y2hhLWNuLm5ldIISKi5yZWNhcHRjaGEtY24ubmV0
ggt3aWRldmluZS5jboINKi53aWRldmluZS5jboIRYW1wcHJvamVjdC5vcmcuY26C
EyouYW1wcHJvamVjdC5vcmcuY26CEWFtcHByb2plY3QubmV0LmNughMqLmFtcHBy
b2plY3QubmV0LmNughdnb29nbGUtYW5hbHl0aWNzLWNuLmNvbYIZKi5nb29nbGUt
YW5hbHl0aWNzLWNuLmNvbYIXZ29vZ2xlYWRzZXJ2aWNlcy1jbi5jb22CGSouZ29v
Z2xlYWRzZXJ2aWNlcy1jbi5jb22CEWdvb2dsZXZhZHMtY24uY29tghMqLmdvb2ds
ZXZhZHMtY24uY29tghFnb29nbGVhcGlzLWNuLmNvbYITKi5nb29nbGVhcGlzLWNu
LmNvbYIVZ29vZ2xlb3B0aW1pemUtY24uY29tghcqLmdvb2dsZW9wdGltaXplLWNu
LmNvbYISZG91YmxlY2xpY2stY24ubmV0ghQqLmRvdWJsZWNsaWNrLWNuLm5ldIIY
Ki5mbHMuZG91YmxlY2xpY2stY24ubmV0ghYqLmcuZG91YmxlY2xpY2stY24ubmV0
gg5kb3VibGVjbGljay5jboIQKi5kb3VibGVjbGljay5jboIUKi5mbHMuZG91Ymxl
Y2xpY2suY26CEiouZy5kb3VibGVjbGljay5jboIRZGFydHNlYXJjaC1jbi5uZXSC
EyouZGFydHNlYXJjaC1jbi5uZXSCHWdvb2dsZXRyYXZlbGFkc2VydmljZXMtY24u
Y29tgh8qLmdvb2dsZXRyYXZlbGFkc2VydmljZXMtY24uY29tghhnb29nbGV0YWdz
ZXJ2aWNlcy1jbi5jb22CGiouZ29vZ2xldGFnc2VydmljZXMtY24uY29tghdnb29n
bGV0YWdtYW5hZ2VyLWNuLmNvbYIZKi5nb29nbGV0YWdtYW5hZ2VyLWNuLmNvbYIY
Z29vZ2xlc3luZGljYXRpb24tY24uY29tghoqLmdvb2dsZXN5bmRpY2F0aW9uLWNu
LmNvbYIkKi5zYWZlZnJhbWUuZ29vZ2xlc3luZGljYXRpb24tY24uY29tghZhcHAt
bWVhc3VyZW1lbnQtY24uY29tghgqLmFwcC1tZWFzdXJlbWVudC1jbi5jb22CC2d2
dDEtY24uY29tgg0qLmd2dDEtY24uY29tggtndnQyLWNuLmNvbYINKi5ndnQyLWNu
LmNvbYILMm1kbi1jbi5uZXSCDSouMm1kbi1jbi5uZXSCFGdvb2dsZWZsaWdodHMt
Y24ubmV0ghYqLmdvb2dsZWZsaWdodHMtY24ubmV0ggxhZG1vYi1jbi5jb22CDiou
YWRtb2ItY24uY29tghRnb29nbGVzYW5kYm94LWNuLmNvbYIWKi5nb29nbGVzYW5k
Ym94LWNuLmNvbYIeKi5zYWZlbnVwLmdvb2dsZXNhbmRib3gtY24uY29tgg0qLmdz
dGF0aWMuY29tghQqLm1ldHJpYy5nc3RhdGljLmNvbYIKKi5ndnQxLmNvbYIRKi5n
Y3BjZG4uZ3Z0MS5jb22CCiouZ3Z0Mi5jb22CDiouZ2NwLmd2dDIuY29tghAqLnVy
bC5nb29nbGUuY29tghYqLnlvdXR1YmUtbm9jb29raWUuY29tggsqLnl0aW1nLmNv
bYILYW5kcm9pZC5jb22CDSouYW5kcm9pZC5jb22CEyouZmxhc2guYW5kcm9pZC5j
b22CBGcuY26CBiouZy5jboIEZy5jb4IGKi5nLmNvggZnb28uZ2yCCnd3dy5nb28u
Z2yCFGdvb2dsZS1hbmFseXRpY3MuY29tghYqLmdvb2dsZS1hbmFseXRpY3MuY29t
ggpnb29nbGUuY29tghJnb29nbGVjb21tZXJjZS5jb22CFCouZ29vZ2xlY29tbWVy
Y2UuY29tgghnZ3BodC5jboIKKi5nZ3BodC5jboIKdXJjaGluLmNvbYIMKi51cmNo
aW4uY29tggh5b3V0dS5iZYILeW91dHViZS5jb22CDSoueW91dHViZS5jb22CFHlv
dXR1YmVlZHVjYXRpb24uY29tghYqLnlvdXR1YmVlZHVjYXRpb24uY29tgg95b3V0
dWJla2lkcy5jb22CESoueW91dHViZWtpZHMuY29tggV5dC5iZYIHKi55dC5iZYIa
YW5kcm9pZC5jbGllbnRzLmdvb2dsZS5jb22CG2RldmVsb3Blci5hbmRyb2lkLmdv
b2dsZS5jboIcZGV2ZWxvcGVycy5hbmRyb2lkLmdvb2dsZS5jboIYc291cmNlLmFu
ZHJvaWQuZ29vZ2xlLmNughpkZXZlbG9wZXIuY2hyb21lLmdvb2dsZS5jboIYd2Vi
LmRldmVsb3BlcnMuZ29vZ2xlLmNuMCEGA1UdIAQaMBgwCAYGZ4EMAQIBMAwGCisG
AQQB1nkCBQMwPAYDVR0fBDUwMzAxoC+gLYYraHR0cDovL2NybHMucGtpLmdvb2cv
Z3RzMWMzL2ZWSnhiVi1LdG1rLmNybDCCAQMGCisGAQQB1nkCBAIEgfQEgfEA7wB1
ADtTd3U+LbmAToswWwb+QDtn2E/D9Me9AA0tcm/h+tQXAAABjcCbl6wAAAQDAEYw
RAIgSsPqmrwU1+TmKxlq0lWFp8HhAWMNr8TpdCsZZ2hIZIQCIHHVd134qMevyD/v
xTWo8mVLjIJIbBUcO8Lm0alcvvkXAHYAdv+IPwq2+5VRwmHM9Ye6NLSkzbsp3GhC
Cp/mZ0xaOnQAAAGNwJuV/QAABAMARzBFAiA90a0s0Fw86i60HTH7XDtVIHnOE9sr
iosICjMRaNAzHgIhANdDshIUQaZkQM2lWPLTvmT3DKZqVMFnA5Aq425HslpPMA0G
CSqGSIb3DQEBCwUAA4IBAQBAjDmy7+UAyQqr2eYerPnwfAaSkD+3kcrbCW0Z656D
rejG3DE2yJ3q/Ao8OqZfg5hC/YpOaMvZMTwvdmLb6urWpgGdgEGVaWW+SHGrckVJ
scLqOVJ1ceXWMhMnp5QHtIhMHsBVwKPT+QM058b6oro2xamIACbAGC6eaem/TF0e
05holHHlBFPZk94PdLfB9f+nzobuWk6K+IaNihsCSgea5C31W1eWW9sE9Z9i3UXa
jbkTdgNtgRKT5HOoz4V+VPeR4uR/VBrQLrx1FsdICP6fCzG4lj1k9dJl3BLRk4xk
RJvLoyFyoRJMiqSpMcxNg7YTgK1ttUUj5BWT6rofaTxp
-----END CERTIFICATE-----
subject=CN = *.google.com
issuer=C = US, O = Google Trust Services LLC, CN = GTS CA 1C3
---
No client certificate CA names sent
Peer signing digest: SHA256
Peer signature type: ECDSA
Server Temp Key: X25519, 253 bits
---
SSL handshake has read 6813 bytes and written 406 bytes
Verification: OK
---
New, TLSv1.3, Cipher is TLS_AES_256_GCM_SHA384
Server public key is 256 bit
Secure Renegotiation IS NOT supported
Compression: NONE
Expansion: NONE
No ALPN negotiated
Early data was not sent
Verify return code: 0 (ok)
---
^C
```

julianz-:
No I don't have anything relevant in /etc/hosts. Here's more of the error message:

`openssl s_client -connect "encrypted.google.com":443                                       
CONNECTED(00000005)
depth=0 OU = "No SNI provided; please fix your client.", CN = invalid2.invalid
verify error:num=18:self signed certificate
`

On searching for info about "No SNI provided", I see I can get corrects result by adding the flag -servername to the call to openssl: 

e.g. 

`openssl s_client -servername "encrypted.google.com":443 -connect "encrypted.google.com":443`

gives the correct certificate result i.e. error:num = 0. 

But how would I apply this fix to the test code in question? Why is it even necessary to provide the SNI on my machine but this test seems to pass on GitHub without any intervention?

webknjaz:
@julianz- if you add `--pdb` to the `pytest` command, are you able to inspect the exception object interactively?

Perhaps @mhils would be able to make a better advice...

julianz-:
Actually I tried that but the error seems to be coming from SSL internals.  After calling do_handshake() in SSL.py I can step through the python code but the error is coming from the C++ SSL library which throws an error that eventually gets translated in the command line as:

 OpenSSL.SSL.Error: [('system library', '', ''), ('system library', '', ''), ('system library', '', ''), ('system library', '', ''), ('SSL routines', '', 'certificate verify failed')]

src/OpenSSL/SSL.py
```diff
@@ -223,11 +223,8 @@
 except AttributeError:
     pass
 
-try:
-    OP_LEGACY_SERVER_CONNECT = _lib.SSL_OP_LEGACY_SERVER_CONNECT
-    __all__.append("OP_LEGACY_SERVER_CONNECT")
-except AttributeError:
-    pass
+OP_LEGACY_SERVER_CONNECT = _lib.SSL_OP_LEGACY_SERVER_CONNECT
+__all__.append("OP_LEGACY_SERVER_CONNECT")
```
jlaine, line 227: If this is no longer conditional, wouldn't it be better to move it up with the other definitions above and include it straight into `__all__`? This will need a rebase since I added an `int` type annotation, sorry :)

webknjaz:
@julianz- perhaps it's time to rebase and see if that's still an issue + maybe ask for hints from the maintainers?

webknjaz:
@mhils @alex would anyone be able to post any actionable advice on this PR, please?

alex:
I think to get this merged, the following need to happen:

- [ ] Dropping support for Python versions or anything like that should not be in this PR
- [ ] I don't understand why several tests were removed? Those seem orthagonal to what I understand this PR to do
- [ ] We need a test that shows the desired behavior with this PR (not just that the flag is set, but that the object behaves correctly.)

julianz-:
> @julianz- could you explain the changes in tests? And also, it's probably a good idea to submit dropping older Python versions from the CI matrix together with updating the packaging metadata and possibly the docs in a separate PR.

I had to remove Python 3.7 from part of the CI matrix because the tests seem to invoke the latest version of cryptography -  v46 - which dropped support for that version of Python. It's a little confusing especially as we specify a version of cryptography less than 46 in this PR. Is there another way to deal with this issue?

I removed tests from test_ssl.py that were failing the static type testing invoked by py313-mypy because they seemed to have been designed to test functionality that no longer exists. For example, this line in test_short_bytearray():
        
        count = server.send(bytearray(b"xy"))

As Connection.send only accepts a parameter of type bytes this code fails the type testing.  

I agree it's better to separate out the removal of these tests so will create a new PR for that if you agree with the above reasoning.

alex:
I am to get to it this weekend.

setup.py
```diff
@@ -94,7 +94,7 @@ def find_meta(meta):
         packages=find_packages(where="src"),
         package_dir={"": "src"},
         install_requires=[
-            "cryptography>=41.0.5,<46",
+            "cryptography>=42.0.0,<46",
```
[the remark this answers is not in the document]
julianz-: The use of SSL_MODE_ACCEPT_MOVING_WRITE_BUFFER requires cryptography v42 as this mode was removed from earlier versions of cryptography and I had to get it restored (the mode is part of the openSSL library).

tests/test_ssl.py
```diff
@@ -2997,6 +3147,261 @@ def test_wantWriteError(self) -> None:
...
+                1024 * 1024 * 64
+            ):  # Large loop count to ensure buffer fill
+                try:
+                    client_socket.send(msg)
+                    time.sleep(0.01)
+                except OSError as e:
+                    if e.errno == EWOULDBLOCK:
+                        print("Client socket buffer filled (EWOULDBLOCK hit).")
+                        return  # Buffer successfully filled, exit function
+                    raise  # pragma: no cover # Re-raise unexpected OSErrors
```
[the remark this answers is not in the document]
julianz-: Ok sure wlll replace.

tests/test_ssl.py
```diff
@@ -2997,6 +3147,261 @@ def test_wantWriteError(self) -> None:
 
     # XXX want_read
 
+    def _fill_client_buffer(self, client_socket: socket) -> None:
+        """
+        Attempts to fill the client's raw send buffer until
+        EWOULDBLOCK is hit.
+        """
+        print("--- Phase 1: Filling client socket buffer ---")
```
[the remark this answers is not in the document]
julianz-: Sure - will remove these.

tests/test_ssl.py
```diff
@@ -2997,6 +3147,261 @@ def test_wantWriteError(self) -> None:
...
+        """
+        After an `OpenSSL.SSL.WantWriteError` if the SSL connection processed
+        some data, the connection may expect a retry with the same buffer.
+        Using mode SSL_MODE_ACCEPT_MOVING_WRITE_BUFFER makes it possible
+        to use a different buffer location provided the length is the same.
+        """
+        mode = (
+            _lib.SSL_MODE_ENABLE_PARTIAL_WRITE
+            | _lib.SSL_MODE_ACCEPT_MOVING_WRITE_BUFFER
+        )
```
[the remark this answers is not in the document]
julianz-: Ok - will revise.

tests/test_ssl.py
```diff
@@ -413,6 +416,153 @@ def handshake_in_memory(
...
+                isinstance(first_error_tuple, tuple)
+                and len(first_error_tuple) >= 3
+            ):
+                reason = first_error_tuple[2]
+                if isinstance(reason, str):
+                    return reason
+    return None
+
+
+def create_ssl_nonblocking_connection(
```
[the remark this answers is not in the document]
alex: That's not correct, the loopback function uses `socket_pair` which creates a real socket.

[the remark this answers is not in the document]
julianz-: Ok I think I was mislead by the comment in loopback() which says: "Create a connected socket pair and force two connected SSL sockets to talk to each other via memory BIOs." Is this correct? Anyway, I revised my  create_ssl_nonblocking_connection() function to reuse handshake() rather than loopback() as I needed more granular control over the setup (such as setting buffer sizes). loopback() also sets the sockets in blocking mode while I need in non-blocking mode.

Have simplified the overall logic of the moving buffer tests eliminating an extra phase of buffer filling that proved unnecessary and also scrapped the variable buffer size probing as I am checking on buffer sizes allocated by the environment instead. I have replaced print statements with logger.debug() statements which can be turned on by running pytest with the --log-cli-level=DEBUG. These are extremely useful for debugging these tests if they don't work - some of CI environments are more challenging than others so I needed these.  I guess they could be removed - but if the tests stopped working in the future this would be very unhelpful!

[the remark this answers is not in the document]
alex: Yeah, that comment is just flat wrong. Sorry for the confusion I'll send a PR to fix that.

You should also be able to use the `loopback_server_factory`/`loopback_client_factory` to remove most of the rest of the code in this function.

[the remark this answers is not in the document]
julianz-: Great thanks! I used loopback_server_factory() to simplify create_ssl_nonblocking_connection() further as you suggested. loopback_client_factory() I didn't use as I needed control over the context to be able to set the mode but the replacement code is short anyway.

tests/test_ssl.py
```diff
@@ -2997,6 +3095,248 @@ def test_wantWriteError(self) -> None:
...
+            except SSL.WantWriteError:
+                logger.debug(
+                    f"After {i + 1} attempt(s) successfully induced a "
+                    f"WantWriteError with message size {buffer_size}. "
+                    "Buffer location in _attempt_want_write_error() "
+                    f"is {id(msg):#x}"
+                )
+                initial_want_write_triggered = True
+                break  # Exit loop as desired error was triggered
+            except Exception as e:  # pragma: no cover
```
alex, line 3127: In general, we strongly avoid no cover pragmas. You can simply omit this handling, if some other exception is raised, you can just let it propagate.

tests/test_ssl.py
```diff
@@ -2997,6 +3095,248 @@ def test_wantWriteError(self) -> None:
...
+                error_string = str(e)
+                logger.debug(f"Attempt {i} failed with error: {error_string}")
+
+        if not initial_want_write_triggered:
+            pytest.fail("Could not induce WantWriteError")  # pragma: no cover
+
```
alex, line 3133: Similarly, this can simply be `assert initial_want_write_triggered`

tests/test_ssl.py
julianz-: Good to know. Have revised to remove all but one of my no cover pragmas. The one remaining in _shutdown_connections() is forgivable I think but see what you think.

tests/test_ssl.py
julianz-: Done

.github/workflows/ci.yml
```diff
@@ -57,7 +57,7 @@ jobs:
         with:
           python-version: ${{ matrix.PYTHON.VERSION }}
       - run: python -m pip install tox
-      - run: tox -v
+      - run: tox -v -- --log-cli-level=DEBUG --show-capture=all
```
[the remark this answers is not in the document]
julianz-: Okay I will remove logging although I guess I was worried about how future proof these tests will be.  With changing environments, the buffer sizes might require tweaking to ensure success. Without the logging it might be harder to debug what is going on. 

The above change which was to turn on the logger I will remove.

tests/test_ssl.py
```diff
@@ -2997,6 +3095,225 @@ def test_wantWriteError(self) -> None:
...
+        server: Connection,
+        client_socket: socket,
+        server_socket: socket,
+    ) -> None:
+        """Helper to safely shut down SSL connections and close sockets."""
+        logger.debug("--- Cleanup: Shutting down connections ---")
+        try:
+            if client:
+                client.shutdown()
+        except Exception:
```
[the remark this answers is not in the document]
julianz-: Do we really care about the precise exception? By the time we have gotten here the test is pretty much complete - I'm just tearing down the sockets as part of house cleaning.

tests/test_ssl.py
```diff
@@ -2997,6 +3095,225 @@ def test_wantWriteError(self) -> None:
...
+            # An exception is usually thrown so it is caught and ignored.
+            pass
+        try:
+            if server:
+                server.shutdown()
+        except Exception:  # pragma: no cover
```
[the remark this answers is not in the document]
julianz-: Same comment as above

tests/test_ssl.py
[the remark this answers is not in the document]
julianz-: Maybe so - that's the only exception I am seeing when I shut down the client connection. And I see no exception when shutting down the server connection. So I can change to that.

tests/test_ssl.py
```diff
@@ -2997,6 +3095,225 @@ def test_wantWriteError(self) -> None:
...
+        server_socket: socket,
+    ) -> None:
+        """Helper to safely shut down SSL connections and close sockets."""
+        logger.debug("--- Cleanup: Shutting down connections ---")
+        try:
+            if client:
+                client.shutdown()
+        except Exception:
+            # An exception is usually thrown so it is caught and ignored.
+            pass
```
webknjaz, line 3217: I agree with Alex's comments: catch-all exceptions will inevitably hide unforeseen problems in the future. Additionally, it's best to scope exception handling to a single instruction as more lines within try/except create more unobvious control flow branches. Plus, something like a `contextlib.suppress()` CM with a clear code comment would relay the intent much better:
```suggestion
        if client:
            with contextlib.suppress(OpenSSL.SSL.Error):
                # When closing connections in the test teardown stage, we don't care
                # about possible TLS-level problems as the test was specifically
                # emulating corner case situations pre-shutdown. We just attempt
                # releasing resources if possible and disregard any possibly related
                # problems that may occur at this point.
                client.shutdown()
```

tests/test_ssl.py
```diff
@@ -2997,6 +3095,225 @@ def test_wantWriteError(self) -> None:
...
+        self._drain_server_buffers(server, server_socket)
+
+        # now attempt the moving buffer retry
+        result = self._perform_moving_buffer_test(
+            client, buffer_size, want_bad_retry
+        )
+        self._shutdown_connections(
+            client, server, client_socket, server_socket
+        )
+        return result
```
webknjaz, line 3271: This looks like a sort of thing that seems natural to wrap with `@contextlib.contextmanager` or (even better `@pytest.fixture` with a `yield` for cleanup and indirect params).

tox.ini
```diff
@@ -60,7 +60,7 @@ commands =
 extras =
     docs
 commands =
-     sphinx-build -W -b html doc doc/_build/html {posargs}
+     sphinx-build -W -b html doc doc/_build/html
```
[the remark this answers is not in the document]
julianz-: Yes

alex:
There's some lint errors, and you'll need to update the minimum version in noxfile.py, but the code itself looks good. Thanks for your persistence herer!

.github/workflows/ci.yml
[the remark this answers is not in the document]
julianz-: Ok logging has gone and replaced where possible with asserts

tox.ini
```diff
@@ -18,7 +18,7 @@ extras =
     test
 deps =
     coverage>=4.2
-    cryptographyMinimum: cryptography==41.0.5
+    cryptographyMinimum: cryptography==42.0.0
```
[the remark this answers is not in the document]
julianz-: Done

alex (approved):
Thank you!

julianz-:
Thank you @alex and @webknjaz!

julianz-:
> Thanks! Hopefully, this will unblock Cheroot once released.

Agreed! I wanted to create a PR for Cheroot once we had this.

webknjaz:
> > Thanks! Hopefully, this will unblock Cheroot once released.
> 
> Agreed! I wanted to create a PR for Cheroot once we had this.

It's been released a week ago. So that's finally unblocked!

julianz-:
Great thanks for the heads up!

Issue #1224, opened 2023-06-06
SSL_CTX_set_cipher_list missing implementation


BillyGLW:
In regards to set_ciphers method in Context class and taking into account of openssl docs - setting ciphers for TLSv1.3 (in order to achieve any effect) must be set via SSL_CTX_set_cipher_list - elsewhere set_ciphers wont take any effect or raise exception - so that, changes require to distinguish between protocol versions - propably here: https://github.com/pyca/pyopenssl/blob/main/src/OpenSSL/SSL.py#L1327

BillyGLW:
PR bump: https://github.com/pyca/pyopenssl/pull/1292

PR #1292, opened 2024-02-03
SSL_CTX_set_ciphersuites for tlsv3 context

BillyGLW:
Linking: https://github.com/pyca/pyopenssl/issues/1224

As cffi binding allows to use SSL_CTX_set_ciphersuites - adaptation for setting the ciphers in context of TLSv3 can be made.

### DIFF
--- a/src/OpenSSL/SSL.py
+++ b/src/OpenSSL/SSL.py
@@ -1320,6 +1320,26 @@ class Context:
         """
         _lib.SSL_CTX_set_tmp_ecdh(self._context, curve._to_EC_KEY())
 
+    def set_ciphersuites(self, cipher_list):
+        """
+        Set the list of ciphers to be used to configure the available TLSv1.3
+        ciphersuites for this context.
+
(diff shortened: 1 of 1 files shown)


src/OpenSSL/SSL.py
```diff
@@ -1320,6 +1320,26 @@ def set_tmp_ecdh(self, curve):
         """
         _lib.SSL_CTX_set_tmp_ecdh(self._context, curve._to_EC_KEY())
 
+    def set_ciphersuites(self, cipher_list):
```
jlaine, line 1323: `def set_ciphersuites(self, cipher_list: bytes) -> None` :)

jlaine:
Isn't it a bit unusual to hijack the test of another method, just because the two methods require bytes as input?

src/OpenSSL/SSL.py
BillyGLW: Done

PR #1308, opened 2024-06-21
Add type annotations for the `SSL` module

jlaine:
The type annotations provided in `types-OpenSSL` are at best incomplete, and are bound to fall out of sync with `PyOpenSSL`. This PR brings type annotations to all public APIs in the `SSL` module.

It also changes references from `unicode` to `str` in the docstrings.

NOTE: I was not sure how to type the first argument of `recv_into`.


src/OpenSSL/SSL.py
```diff
@@ -1113,11 +1131,13 @@ def _raise_passphrase_exception(self):
 
         _raise_current_error()
 
-    def use_privatekey_file(self, keyfile, filetype=_UNSPECIFIED):
+    def use_privatekey_file(
+        self, keyfile: _StrOrBytesPath, filetype=_UNSPECIFIED
```
alex, line 1135:
```suggestion
        self, keyfile: _StrOrBytesPath, filetype: int=_UNSPECIFIED
```

jlaine: Unfortunately the UNSPECIFIED sentinel is an object. Is there any reason we are not using `None` here?

alex: I don't see a reason, no.

alex:
If you rebase on main, CI should be green.

mhils (commented):
Strong +1 on inlining the annotations from types-pyopenssl, thank you for doing this @jlaine! 🍰

LGTM % @alex's comment.

src/OpenSSL/SSL.py
```diff
@@ -142,12 +144,18 @@
...
+OPENSSL_VERSION_NUMBER: int = _lib.OPENSSL_VERSION_NUMBER
+OPENSSL_VERSION: int = _lib.OPENSSL_VERSION
+OPENSSL_CFLAGS: int = _lib.OPENSSL_CFLAGS
+OPENSSL_PLATFORM: int = _lib.OPENSSL_PLATFORM
+OPENSSL_DIR: int = _lib.OPENSSL_DIR
+OPENSSL_BUILT_ON: int = _lib.OPENSSL_BUILT_ON
+
+SSLEAY_VERSION = OPENSSL_VERSION
+SSLEAY_CFLAGS = OPENSSL_CFLAGS
+SSLEAY_PLATFORM = OPENSSL_CFLAGS
```
alex, line 156:
```suggestion
SSLEAY_PLATFORM = OPENSSL_PLATFORM
```

jlaine:
I have rebased on top of `main`, let's see if CI is happy.

I also added `int` annotations to our constants, like `types-OpenSSL` does.

Some questions:
- Do we *only* support passing a `socket.socket` to `Connection`, or is something that quacks like a socket OK too?
- Related to the above : `Connection.connect` and `Connection.sock_shutdown` say they return "what the socket returns". If the socket is *garanteed* to be a `socket.socket`, that would be `None`, and there is no need to return anything. But as we *might* be supporting weird socket-like things, I'd rather not touch it.
- Guidance on typing the first argument of `Connection.recv_into` would be welcome!
- Do we want to ship a `py.typed` marker so that our typings take precedence over `types-OpenSSL`?

As a stretch goal I was wondering whether we can get a clean `mypy` run on `SSL.py`.  A couple of aspects of the code make this moderately hard:

- `_ffi.string` returns `bytes` but `mypy` does know this. Do we want to add: `_ffi_string: Callable[[Any], bytes] = _ffi.string` ?
- `_raise_ssl_error` is sometimes used in context where we *know* it will raise .. and sometimes not. This results in mypy complaining there is no `return` statement on some codepaths.

src/OpenSSL/SSL.py
jlaine: OK, so `filetype: int = FILETYPE_PEM` it is!

src/OpenSSL/py.typed
```diff
@@ -0,0 +1 @@
+Marker
```
alex, line 1: This file should be empty

jlaine: Huh. For some reason I've been putting this same content all over my projects for years, not sure where I got it from?! I'll clear it.

alex (approved):

jlaine:
I can't remember what your preferred policy is, do you want me to squash the changes down or would you rather do it?

reaperhulk:
We’re happy to squash ourselves.

alex:
We're fine to squash ourselves. Thanks.

If you want a side-quest, the next milestone would be to get mypy running over our tests directory (see `tox.ini`) :-)

PR #1310, opened 2024-06-21
Allow loading EC, ED25519, ED448 public keys from cryptography

jlaine:
Our code is quite capable of supporting these public key types, so allow users to do so.

We also beef up the test suite to test all these key types, along with DSA keys which were not explicitly tested.

### DIFF
--- a/src/OpenSSL/crypto.py
+++ b/src/OpenSSL/crypto.py
@@ -89,7 +89,16 @@ __all__ = [
 
 
 _Key = Union[
-    dsa.DSAPrivateKey, dsa.DSAPublicKey, rsa.RSAPrivateKey, rsa.RSAPublicKey
+    dsa.DSAPrivateKey,
+    dsa.DSAPublicKey,
+    ec.EllipticCurvePrivateKey,
+    ec.EllipticCurvePublicKey,
+    ed25519.Ed25519PrivateKey,
+    ed25519.Ed25519PublicKey,
+    ed448.Ed448PrivateKey,
+    ed448.Ed448PublicKey,
+    rsa.RSAPrivateKey,
+    rsa.RSAPublicKey,
 ]
 StrOrBytesPath = Union[str, bytes, PathLike]
 PassphraseCallableT = Union[bytes, Callable[..., bytes]]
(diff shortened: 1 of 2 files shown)


jlaine:
This also fixes our type annotations, which did not mention we support `from_cryptography` on elliptic curve private keys.

jlaine:
I now notice this seems to overlap with #636

src/OpenSSL/crypto.py
```diff
@@ -328,7 +340,7 @@ def from_cryptography_key(cls, crypto_key: _Key) -> "PKey":
             PublicFormat,
         )
 
-        if isinstance(crypto_key, (rsa.RSAPublicKey, dsa.DSAPublicKey)):
+        if hasattr(crypto_key, "public_bytes"):
```
alex, line 343: I'd prefer we list our all of the public key types here. There's no guarantee that we'll never add a public bytes method to a private key.

alex (commented):
@reaperhulk do you remember if there's a reason we did rsa/dsa only originally?

src/OpenSSL/crypto.py
jlaine: Fine by me, I'll make the requested change.

alex (approved):
LGTM, pending @reaperhulk confirming that there wasn't a good reason we limited this.

jlaine:
For context I was trying to eliminate `mypy` errors in `aiortc`, as it uses `PKey.from_cryptography` on elliptic curve. The code worked fine, but the definition of `_Key` did not include `ec.EllipticCurvePrivateKey`.

I then noticed how asymmetrical (heh) the behaviour was: for DSA and RSA we supported both private keys and public keys, for `ec` / `ed25519`  / `ed448`  only private keys. That's why this PR ended up bringing everything into line.

reaperhulk:
The `PKey` object has a `type()` method that returns the key type. This will work, but we only expose RSA/DSA constants to check the return value. We probably need to expose the additional constants from cryptography (bleh) or hard code them in pyOpenSSL.

alex:
I think it's true that you can't use `type()` effectively without that, but strictly speaking that problem already existed -- if you had a cert and got a key out of it for example. Therefore I don't think it's a blocker for this PR.

jlaine:
Thanks @alex ! Should I submit a follow up PR that mentions both my changes?

alex:
Ooops, yes, changelog would be good.

PR #1314, opened 2024-06-28
Fix `ruff` invocation

jlaine:
Ruff must now either be invoked as `ruff check` or `ruff format`.


mhils (approved):
Thanks!

jlaine:
Not sure why CI failed here, can someone give it a ~~kick in the head~~ gentle nudge?

PR #1317, opened 2024-07-18
Adapt to future ReadTheDocs changes

hynek:
Since I'm still an RTD maintainer I got an email from warning about upcoming changes. This blog post sums it up too:

https://about.readthedocs.com/blog/2024/07/addons-by-default/

I think this should do.

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -15,6 +15,8 @@ import os
 import re
 import sys
 
+import sphinx_rtd_theme
+
 HERE = os.path.abspath(os.path.dirname(__file__))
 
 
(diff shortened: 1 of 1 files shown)


doc/conf.py
```diff
@@ -122,13 +124,15 @@ def find_version(*file_paths):
 
 # The theme to use for HTML and HTML Help pages.  See the documentation for
 # a list of builtin themes.
-on_rtd = os.environ.get("READTHEDOCS", None) == "True"
+if os.environ.get("READTHEDOCS", None) == "True":
+    html_baseurl = os.environ.get("READTHEDOCS_CANONICAL_URL", "")
```
reaperhulk, line 128: I think this has to actually define the URL? https://docs.readthedocs.io/en/stable/reference/environment-variables.html#envvar-READTHEDOCS_CANONICAL_URL

hynek: (I didn’t get a notification for this :()

What do you mean exactly? My understanding is that this is set by RTD according to the external domain that is configured inside of RTD?

ie this one
image

reaperhulk: If that's the case then great 😄

reaperhulk (approved):

PR #1320, opened 2024-07-20
24.2.1

reaperhulk:


alex (dismissed):

alex (approved):

alex:
Because pyOpenSSL relies on bindings from cryptography that are difficult
to keep stable. The upper bound ensures we do not accidentally break old
versions of pyOpenSSL with new cryptography releases. We issue a new
pyOpenSSL release with every cryptography release to raise the bound.

Issue #1321, opened 2024-07-20
Deprecate/remove all non-SSL APIs in pyOpenSSL


alex:
# Motivation

pyOpenSSL has many APIs with poor designs that are entirely reliant on OpenSSL implementation details. For nearly all of these, pyca/cryptography has much better implementations with more thought out APIs.

Therefore, we would like to deprecate/remove anything besides SSL in pyOpenSSL (SSL has no equivalent in pyca/cryptography, nor is there is a path to adding one).

We've been doing this for a while already, but now we're filing an issue to actually track this in a systematic way.

# Strategy

Most APIs can simply be deprecated and removed. In some cases, where SSL relies on these (e.g., `PKey` objects for private keys or `X509` for certificates), we'll need to expand the SSL API to accept the pyca/cryptography versions of these objects (there are mostly already `from_cryptography` APIs we can keep around).

# Status

- [ ] `rand`
  - [ ] `add`, `status`
    - [ ] Deprecated
    - [ ] Removed
- [ ] `crypto`
  - [ ] `PKey`, `dump_publickey`, `dump_privatekey`, `load_publickey`, `load_privatekey`
    - [ ] Deprecated
    - [ ] Removed
  - [ ] `get_elliptic_curves`, `get_elliptic_curve`
    - [ ] Deprecated
    - [ ] Removed
  - [ ] `X509Name`
    - [ ] Deprecated
    - [ ] Removed
  - [ ] `X509Extension`
    - [x] Deprecated
    - [ ] Removed
  - [ ] `X509Req`, `dump_certificate_request`, `load_certificate_request`
    - [x] Deprecated
    - [ ] Removed
  - [ ] `X509`, `load_certificate`, `dump_certificate`
    - [ ] Deprecated
    - [ ] Removed
  - [ ] `X509Store`, `X509StoreContext`
    - [ ] Deprecated
    - [ ] Removed
  - [ ] `CRL`, `Revoked`, `dump_crl`, `load_crl`
    - [x] Deprecated
    - [ ] Removed
  - [ ] `sign`, `verify`
    - [x] Deprecated
    - [ ] Removed

phlax:
removing `crypto` has just broke gsutil - i suspect it will break a lot of other downstreams too

surely removing things like this would be better suited to a major (or at least minor) release rather than a patch release

alex:
[...]

PR #1325, opened 2024-07-27
Deprecate the rand module

alex:
refs #1321

### DIFF
--- a/src/OpenSSL/rand.py
+++ b/src/OpenSSL/rand.py
@@ -2,7 +2,15 @@
 PRNG management routines, thin wrappers.
 """
 
+import warnings
+
 from OpenSSL._util import lib as _lib
+
+warnings.warn(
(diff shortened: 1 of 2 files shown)


src/OpenSSL/rand.py
```diff
@@ -2,8 +2,16 @@
...
 """
 
+import warnings
+
 from OpenSSL._util import lib as _lib
 
+warnings.warn(
+    "OpenSSL.rand is deprecated - you should use os.urandom instead",
+    DeprecationWarning,
+    stacklevel=3,
```
mhils, line 12: Stacklevel 3 because of how CPython does imports?

alex: Because that's what I used the last time I tried to deprecate this 😭

mhils: 🥲

mhils (approved):

PR #1327, opened 2024-07-28
Allow `set_tmp_ecdh` to take cryptography elliptic curves

alex:
Deprecate `get_elliptic_curves` and `get_elliptic_curve`

Refs #1321

### DIFF
--- a/src/OpenSSL/SSL.py
+++ b/src/OpenSSL/SSL.py
@@ -3,12 +3,15 @@ from __future__ import annotations
 import os
 import socket
 import typing
+import warnings
 from errno import errorcode
 from functools import partial, wraps
 from itertools import chain, count
 from sys import platform
 from typing import Any, Callable, List, Optional, Sequence, TypeVar
 from weakref import WeakValueDictionary
+
+from cryptography.hazmat.primitives.asymmetric import ec
 
 from OpenSSL._util import (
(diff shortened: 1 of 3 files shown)


alex:
Will rebase once https://github.com/pyca/pyopenssl/pull/1328 merges

src/OpenSSL/SSL.py
```diff
@@ -1358,17 +1361,44 @@ def load_tmp_dh(self, dhfile: _StrOrBytesPath) -> None:
...
+                name = "prime192v1"
+            elif name == "secp256r1":
+                name = "prime256v1"
+            nid = _lib.OBJ_txt2nid(name.encode())
+            if nid == _lib.NID_undef:
+                _raise_current_error()
+
+            _lib.SSL_CTX_set_tmp_ecdh(
+                self._context,
+                _lib.EC_KEY_new_by_curve_name(nid),
```
reaperhulk, line 1400: I think we need to register this for gc because it doesn't look like SSL_CTX_set_tmp_ecdh takes ownership. It instead takes the EC_KEY, extracts the group, converts the group to a NID, then uses that NID from then on.

alex: fixed

reaperhulk (approved):

PR #1348, opened 2024-08-06
Deprecate passing `PKey` instances to `use_privatekey`

alex:
Allow passing cryptography keys instead.

Refs #1321

### DIFF
--- a/src/OpenSSL/crypto.py
+++ b/src/OpenSSL/crypto.py
@@ -84,18 +84,21 @@ __all__ = [
 ]
 
 
-_Key = Union[
+_PrivateKey = Union[
     dsa.DSAPrivateKey,
-    dsa.DSAPublicKey,
     ec.EllipticCurvePrivateKey,
-    ec.EllipticCurvePublicKey,
     ed25519.Ed25519PrivateKey,
-    ed25519.Ed25519PublicKey,
     ed448.Ed448PrivateKey,
-    ed448.Ed448PublicKey,
     rsa.RSAPrivateKey,
+]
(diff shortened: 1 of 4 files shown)


src/OpenSSL/crypto.py
```diff
@@ -84,6 +84,14 @@
...
 
 
+_PrivateKey = Union[
+    dsa.DSAPrivateKey,
+    ec.EllipticCurvePrivateKey,
+    ed25519.Ed25519PrivateKey,
+    ed448.Ed448PrivateKey,
+    rsa.RSAPrivateKey,
+]
+# TODO: Turn this into `PublicKey | PrivateKey` once we use `|`.
```
mhils, line 94: Can't we do that right away?
```python3
>>> a = Union[1,2]
>>> b = Union[3,4]
>>> Union[a,b]
typing.Union[1, 2, 3, 4]
```

alex:
Not on older pythons

mhils:
```python
> docker run --rm -it python:3.7.0
Python 3.7.0 (default, Oct 16 2018, 07:10:55)
[GCC 6.3.0 20170516] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from typing import Union
>>> a = Union[str, int]
>>> b = Union[bool, float]
>>> Union[a,b]
typing.Union[str, int, bool, float]
```

This looks ok to me? (I've also tested 3.8-3.10)

alex:
Ohhh, yeah, i guess I didn't try that. Will fix this evening

mhils (approved):

PR #1363, opened 2024-09-08
Add support for Python 3.13

hugovk:
The final Python 3.13 release candidate is out now! :rocket:

The Release Manager has issued a [call to action](https://discuss.python.org/t/python-3-13-0rc2-3-12-6-3-11-10-3-10-15-3-9-20-and-3-8-20-are-now-available/63161?u=hugovk]):

> We strongly encourage maintainers of third-party Python projects to prepare their projects for 3.13 compatibilities during this phase, and where necessary publish Python 3.13 wheels on PyPI to be ready for the final release of 3.13.0. Any binary wheels built against Python 3.13.0rc2 will work with future versions of Python 3.13. As always, report any issues to [the Python bug tracker](https://github.com/python/cpython/issues).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,6 +16,7 @@ jobs:
           - {VERSION: "3.10", TOXENV: "py310"}
           - {VERSION: "3.11", TOXENV: "py311"}
           - {VERSION: "3.12", TOXENV: "py312"}
+          - {VERSION: "3.13-dev", TOXENV: "py313"}
           - {VERSION: "pypy-3.9", TOXENV: "pypy3"}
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -55,6 +58,7 @@ jobs:
         uses: actions/setup-python@v5
         with:
           python-version: ${{ matrix.PYTHON.VERSION }}
+          allow-prereleases: true
```
alex, line 61: We prefer explicitly specifying `3.13-dev`, not doing `allow-prereleases`, and then switching on release day.

hugovk: Updated!

alex (approved):
Thanks!

PR #1382, opened 2024-11-28
Added a workflow to automatically upload artifacts on tag

alex:


reaperhulk (approved):

.github/workflows/release.yml
```diff
@@ -0,0 +1,56 @@
...
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
+
+      - name: Set up Python
+        uses: actions/setup-python@f677139bbe7f9c59b41e40162b753c062f5d49a3 # v5.2.0
+
+      - name: Install build dependencies
+        run: python -m pip install uv
```
mhils, line 24: Pinning all actions but then running `pip install uv` is a bit inconsistent. How about we use https://github.com/astral-sh/setup-uv here with a fixed uv version?

PR #1389, opened 2025-01-07
Use type-level deprecations

alex:


mhils (approved):
Much nicer, thanks! 🍰 

LGTM % CI errors.

alex:
Last challenge here: not compatible with python 3.7.

3.7 is currently about 10% of our downloads. OTOH, it's been unsupported by upstream for 18 months...

mhils:
How about something like this?

```python
if sys.version_info >= (3, 13):
    from warnings import deprecated
elif sys.version_info < (3, 8):
    def deprecated(msg, **_):
        return lambda f: f
else:
    from typing_extensions import deprecated
```

3.7 users won't get deprecation warnings then, but that probably is the least of their concerns...

alex:
Hmm, it works, but it also makes it a regression on 3.7, since we currently warn... annoying...

mhils (approved):

PR #1391, opened 2025-01-07
Fixes for recent mypy

alex:
Including ALPN typing which was totally broken


tox.ini
```diff
@@ -43,7 +43,7 @@ commands =
 
 [testenv:py311-mypy]
 deps =
-    mypy==1.1.1
+    mypy
```
mhils, line 46: This will break CI for unrelated PRs in the future if mypy ships an update with stricter rules. Is that really what we want?

alex: The ideal thing is probably what we do for pyca/cryptography: a constraints file where we pin + dependabot to update stuff.

However, that's way more maintenance than we typically give pyOpenSSL.

Given the choice between "pinned to some old thing, and always green" and "we find out immediately when stuff breaks", I prefer the latter.

I discovered this because a) mypy was failing locally for me (because I wasn't using the tox env), and b) old mypy doesn't support newer stuff.

The fact that old mypy was missing a legitimate typing error further pushes me to think this was a bad trade-off.

PR #1416, opened 2025-02-02
Attempt to prohibit mutating a Context after its in use

alex:


mhils (approved):
This looks really great, thanks!

PR #1429, opened 2025-05-20
make it an error to mutate after using a context

alex:


alex:
about a zillion of our own tests need to be fixed + twisted (which I believe @glyph is looking at)

reaperhulk (dismissed):

reaperhulk:
We need a changelog entry here

reaperhulk (approved):

PR #1432, opened 2025-07-15
Add ability to TLS 1.3 cipher suites on SSL Context

schwabe:


schwabe:
Needs test + changelog entry.

alex:
should be possible to rebase on main to resolve the CI issues

src/OpenSSL/SSL.py
```diff
@@ -1501,6 +1504,31 @@ def set_cipher_list(self, cipher_list: bytes) -> None:
...
+        See the OpenSSL manual for more information (e.g.
+        :manpage:`ciphers(1)`).
+
+        :param bytes ciphersuites: An OpenSSL cipher string containing
+            TLS 1.3+ ciphersuites.
+        :return: None
+
+        .. versionadded:: 25.2.0
+        """
+        ciphersuites = _text_to_bytes_and_warn("ciphersuites", ciphersuites)
```
alex, line 1523: No warning for new APIs, we should strictly enforce types.

tests/test_ssl.py
```diff
@@ -497,20 +497,23 @@ class TestContext:
 
     @pytest.mark.parametrize(
         "cipher_string",
-        [b"hello world:AES128-SHA", "hello world:AES128-SHA"],
+        [b"TLS_AES_128_GCM_SHA256", "TLS_AES_128_GCM_SHA256"],
     )
-    def test_set_cipher_list(
+    def test_set_ciphersuites(
```
alex, line 502: We need tests for both, you can't get rid of the old test.

src/OpenSSL/SSL.py
```diff
@@ -1501,6 +1504,29 @@ def set_cipher_list(self, cipher_list: bytes) -> None:
                 ],
             )
 
+    @_require_not_used
+    def set_ciphersuites(self, ciphersuites: bytes) -> None:
```
alex, line 1508: @reaperhulk should this be `set_tls13_ciphersuites`, or just match the OpenSSL name?

schwabe: I changed the PR to add the tls13 in there. I hope we don't have to rename it again when TLS 1.4/2.0 comes out

PR #1436, opened 2025-08-09
Switch from tox to nox

alex:


noxfile.py
```diff
@@ -0,0 +1,86 @@
...
+    session.env.update(
+        {
+            "PIP_NO_BINARY": "" if use_wheel else "cryptography",
+        }
+    )
+
+    deps = ["coverage>=4.2"]
+
+    if cryptography_version == "minimum":
+        deps.append("cryptography==41.0.5")
```
reaperhulk, line 38: Could we set `41.0.5` as a constant at the top of this script to make it easier for future updates? Now that this is just python we can organize it better than the tox ini 😄

PR #1442, opened 2025-09-10
Add get_group_name to expose SSL_get0_group_name

toddabrams:
- See corresponding PR in cryptography https://github.com/pyca/cryptography/pull/13384

### DIFF
--- a/src/OpenSSL/SSL.py
+++ b/src/OpenSSL/SSL.py
@@ -3222,6 +3222,18 @@ class Connection:
 
         return _ffi.string(profile.name)
 
+    def get_group_name(self) -> str | None:
+        """
+        Get the name of the negotiated group for the key exchange.
+
+        :return: A string giving the group name or :data:`None`.
+        """
+        group_name = _lib.SSL_get0_group_name(self._ssl)
+        if group_name == _ffi.NULL:
+            return None
+
+        return _ffi.string(group_name).decode("utf-8")
+
     def request_ocsp(self) -> None:
         """
         Called to request that the server sends stapled OCSP data, if
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -15,6 +15,7 @@ Deprecations:
 Changes:
 ^^^^^^^^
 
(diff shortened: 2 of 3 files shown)


reaperhulk:
This will be difficult to test until a cryptography release. That said, the cryptography-main job is currently failing because pyOpenSSL unconditionally assumes the binding will be there even when cryptography is linked against an OpenSSL that does not support it (in this case, that job links against system OpenSSL 3.0.2).

toddabrams:
@reaperhulk  This is 98% ready, I just need to figure out how to require cryptography 46.0.0 beyond what I've done in [9a06dc8](https://github.com/pyca/pyopenssl/pull/1442/commits/9a06dc81141a8116a94702460f6ed1061bd878ff)

toddabrams:
> That issue has an easy fix: we dropped python 3.7 in cryptography 46 so we'll drop python 3.7 here when we bump the bound 😄

@reaperhulk Should we drop python 3.7 here or in a different PR? Is there something else still pending?

I add commits here just in case: [cb720a2](https://github.com/pyca/pyopenssl/pull/1442/commits/cb720a2e64fdf4a25512fc172437ed0e1c8e8c5d) [5ef8f20](https://github.com/pyca/pyopenssl/pull/1442/commits/5ef8f20ec0ea264c9540607857c0caaa343cf334)

src/OpenSSL/SSL.py
```diff
@@ -3202,6 +3207,22 @@ def get_selected_srtp_profile(self) -> bytes:
...
+    @_requires_ssl_get0_group_name
+    def get_group_name(self) -> str | None:
+        """
+        Get the name of the negotiated group for the key exchange.
+
+        :return: A string giving the group name or :data:`None`.
+        """
+        session = _lib.SSL_get_session(self._ssl)
+        if session == _ffi.NULL:
+            return None
```
alex, line 3219: Why is the session needed, it doesn't look used?

toddabrams: Feel free to suggest another wording of course

alex (approved):
@reaperhulk Didn't think really about the API design, but the impl looks good. If the API works for you, plz merge.

PR #1451, opened 2025-09-20
test_crypto: Remove unused test data

davidben:
These are no longer referenced anywhere. They also turn out to be wrong. Per RFC 5280 section 5.1, crlEntryExtensions can only be present if the version is v2, but these are v1 with extensions. But it's moot because they aren't used anymore.


alex (approved):
love to delete software

PR #1513, opened 2026-06-12
Require cryptography 49, drop Python 3.8

alex:
- Bump the `cryptography` pin to `>=49.0.0,<50` now that 49 has been released, and update `MINIMUM_CRYPTOGRAPHY_VERSION` in the noxfile to match.
- Drop Python 3.8: `python_requires`, trove classifiers, the two 3.8 CI matrix rows, and the now-dead version gates (the `< (3, 8)` `deprecated` shim in `crypto.py` and the 3.9 `os.PathLike` gate in `_util.py`). The typing-extensions environment marker simplifies to just `python_version < '3.13'`.
- Changelog and docs updated accordingly.

https://claude.ai/code/session_01NiSGpbM6ady8SuJPmiZ9SP

---
_Generated by [Claude Code](https://claude.ai/code/session_01NiSGpbM6ady8SuJPmiZ9SP)_

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,7 +14,6 @@ jobs:
       matrix:
         PYTHON:
           # Base builds
-          - {VERSION: "3.8", NOXSESSION: "tests"}
           - {VERSION: "3.9", NOXSESSION: "tests"}
           - {VERSION: "3.10", NOXSESSION: "tests"}
           - {VERSION: "3.11", NOXSESSION: "tests"}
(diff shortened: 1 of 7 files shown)


reaperhulk (approved):

ahesford:
As a maintainer of a Linux distribution, I'm very curious what critical features in the latest cryptography mandate such a strict lower bound. Needlessly strict version bounds are a warrantless burden on users in general, and an even greater burden on distribution packagers.

reaperhulk:
We bump the lower bound when we depend on something in that version. In this case it's `x509.Name.from_bytes`, which is used to provide paths for migration on deprecated functions.

